PR #8, opened 2022-05-25
Add api, client

toddbaert:
First pass at API and client functionality, not including hooks and evaluation context details.

Added applicable tests based on spec requirements: 

```
 PASS  test/open-feature.spec.ts
  OpenFeature
    Requirement 1.1
      ✓ should be global singleton (2 ms)
    Requirement 1.2
      ✓ should be set provider
    Requirement 1.3
      ✓ should allow addition of hooks
    Requirement 1.4
      ✓ should implement a hook accessor
    Requirement 1.5
      ✓ should implement a client factory (1 ms)

 PASS  test/client.spec.ts
  OpenFeatureClient
    Requirement 1.6
      ✓ should allow addition of hooks
    Requirement 1.7, 1.8
      flag evaluation
         getBooleanValue
          ✓ should return boolean, and call boolean resolver (2 ms)
        getStringValue
          ✓ should return string, and call string resolver
        getNumberValue
          ✓ should return number, and call number resolver (1 ms)
        getObjectValue
          ✓ should return object, and call object resolver
    Requirement 1.9, 1.10
      detailed flag evaluation
         getBooleanDetails
          ✓ should return boolean details, and call boolean resolver (1 ms)
        getStringDetails
          ✓ should return string details, and call string resolver
        getNumberDetails
          ✓ should return number details, and call number resolver
        getObjectDetails
          ✓ should return object details, and call object resolver
    Requirement 1.11
      generic support
        ✓ should support generic
    Evaluation details structure
      Normal execution
        Requirement 1.10, 1.11
          ✓ should contain flag value (1 ms)
        Requirement 1.12
          ✓ should contain flag key
        Requirement 1.13
          ✓ should contain flag variant
        Requirement 1.14
          ✓ should contain reason
      Abnormal execution
        Requirement 1.18
          ✓ must not throw, must return default (1 ms)
        Requirement 1.15
          ✓ should contain error
          ✓ should contain "error" reason
    Requirement 1.21
      Transforming provider
        ✓ should run context transformer, and pass transformed context to resolver (1 ms)
      Non-transforming provider
        ✓ should pass context to resolver
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,7 +14,8 @@
 		 "sourceType":"module"
 	},
 	"plugins":[
-		 "@typescript-eslint"
+		 "@typescript-eslint",
+		 "check-file"
 	],
 	"rules":{
 		 "linebreak-style":[
@@ -28,6 +29,14 @@
 		 "semi":[
 				"error",
 				"always"
-		 ]
+		 ],
+		 "check-file/filename-naming-convention":[
+			"error",
+			{
+				"*.spec.{js,ts}":"*",
+				"**/jest.config.ts":"*",
+				 "*.{js,ts}":"KEBAB_CASE"
+			}
+	 ]
 	}
 }
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,0 +1,157 @@
+import { ERROR_REASON, GENERAL_ERROR } from './constants';
+import { OpenFeature } from './open-feature';
+import {
+  Client,
+  EvaluationContext,
+  EvaluationDetails,
+  FlagEvaluationOptions,
+  FlagValue,
+  Hook,
+  ResolutionDetails,
+  TransformingProvider,
+} from './types';
+
+type OpenFeatureClientOptions = {
+  name?: string;
+  version?: string;
+};
+
+export class OpenFeatureClient implements Client {
+  name?: string | undefined;
+  version?: string | undefined;
+  readonly context: EvaluationContext;
+
+  constructor(private readonly api: OpenFeature, options: OpenFeatureClientOptions, context: EvaluationContext = {}) {
+    this.name = options.name;
+    this.version = options.version;
+    this.context = context;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  addHooks(...hooks: Hook<FlagValue>[]): void {
+    throw new Error('Method not implemented.');
+  }
+
+  get hooks(): Hook<FlagValue>[] {
+    throw new Error('Method not implemented.');
+  }
+
+  async getBooleanValue(
+    flagKey: string,
+    defaultValue: boolean,
+    context?: EvaluationContext,
+    options?: FlagEvaluationOptions
+  ): Promise<boolean> {
+    return (await this.getBooleanDetails(flagKey, defaultValue, context, options)).value;
+  }
+
+  getBooleanDetails(
+    flagKey: string,
+    defaultValue: boolean,
+    context?: EvaluationContext,
+    options?: FlagEvaluationOptions
+  ): Promise<EvaluationDetails<boolean>> {
+    return this.evaluate<boolean>(flagKey, this.provider.resolveBooleanEvaluation, defaultValue, context, options);
+  }
+
+  async getStringValue(
+    flagKey: string,
+    defaultValue: string,
+    context?: EvaluationContext,
+    options?: FlagEvaluationOptions
+  ): Promise<string> {
+    return (await this.getStringDetails(flagKey, defaultValue, context, options)).value;
+  }
+
+  getStringDetails(
+    flagKey: string,
+    defaultValue: string,
+    context?: EvaluationContext,
+    options?: FlagEvaluationOptions
+  ): Promise<EvaluationDetails<string>> {
+    return this.evaluate<string>(flagKey, this.provider.resolveStringEvaluation, defaultValue, context, options);
+  }
+
+  async getNumberValue(
+    flagKey: string,
+    defaultValue: number,
+    context?: EvaluationContext,
+    options?: FlagEvaluationOptions
+  ): Promise<number> {
+    return (await this.getNumberDetails(flagKey, defaultValue, context, options)).value;
+  }
+
+  getNumberDetails(
+    flagKey: string,
+    defaultValue: number,
+    context?: EvaluationContext,
+    options?: FlagEvaluationOptions
+  ): Promise<EvaluationDetails<number>> {
+    return this.evaluate<number>(flagKey, this.provider.resolveNumberEvaluation, defaultValue, context, options);
+  }
+
+  async getObjectValue<T extends object>(
+    flagKey: string,
+    defaultValue: T,
+    context?: EvaluationContext,
+    options?: FlagEvaluationOptions
+  ): Promise<T> {
+    return (await this.getObjectDetails(flagKey, defaultValue, context, options)).value;
+  }
+
+  getObjectDetails<T extends object>(
+    flagKey: string,
+    defaultValue: T,
+    context?: EvaluationContext,
+    options?: FlagEvaluationOptions
+  ): Promise<EvaluationDetails<T>> {
+    return this.evaluate<T>(flagKey, this.provider.resolveObjectEvaluation, defaultValue, context, options);
+  }
+
+  private async evaluate<T extends FlagValue>(
+    flagKey: string,
+    resolver: (
+      flagKey: string,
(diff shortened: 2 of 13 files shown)


.eslintrc.json
```diff
@@ -28,6 +29,13 @@
...
 				"always"
-		 ]
+		 ],
+		 "check-file/filename-naming-convention":[
+			"error",
+			{
+				"*.spec.{js,ts}":"*",
+				 "*.{js,ts}":"KEBAB_CASE"
+			}
+	 ]
```
toddbaert, line 39: Force kabab case in file names.

src/client.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  addHooks(...hooks: Hook<FlagValue>[]): void {
+    throw new Error('Method not implemented.');
+  }
+
+  get hooks(): Hook<FlagValue>[] {
+    throw new Error('Method not implemented.');
+  }
```
toddbaert, line 37: Hooks are stubbed for now.

src/client.ts
```diff
@@ -0,0 +1,157 @@
...
+    return (await this.getBooleanDetails(flagKey, defaultValue, context, options)).value;
+  }
+
+  getBooleanDetails(
+    flagKey: string,
+    defaultValue: boolean,
+    context?: Context,
+    options?: FlagEvaluationOptions
+  ): Promise<EvaluationDetails<boolean>> {
+    return this.evaluate<boolean>(flagKey, this.provider.resolveBooleanEvaluation, defaultValue, context, options);
```
toddbaert, line 54: Instead of an ugly switch case, we can pass a reference to the provider function we want to run: `this.provider.resolveBooleanEvaluation` and then bind the provider to it when we call it: https://github.com/open-feature/node-sdk/pull/8/files#diff-25d66d74617fe2e23d7946bd6e3ba95640ab1b9bc8947445d604fc271c7c1f12R138

src/open-feature.ts
```diff
@@ -0,0 +1,59 @@
...
+import { NOOP_PROVIDER } from './no-op-provider';
+import { Client, Context, EvaluationLifeCycle, FlagValue, Hook, Provider } from './types';
+
+// use a symbol as a key for the global singleton
+const GLOBAL_OPENFEATURE_API_KEY = Symbol.for('@openfeature/js.api');
+
+type OpenFeatureGlobal = {
+  [GLOBAL_OPENFEATURE_API_KEY]?: OpenFeature;
+};
+const _global = global as OpenFeatureGlobal;
```
toddbaert, line 11: I could have done this in a separate file, but then I'd have to export it and it would be exported in the dist. Keeping it in this file makes it effectively private without introducing a bundler, which we may want to do later.

.eslintrc.json
```diff
@@ -28,6 +29,14 @@
...
-		 ]
+		 ],
+		 "check-file/filename-naming-convention":[
+			"error",
+			{
+				"*.spec.{js,ts}":"*",
+				"**/jest.config.ts":"*",
+				 "*.{js,ts}":"KEBAB_CASE"
+			}
+	 ]
```
toddbaert, line 40: Force kabab case in file names.

src/open-feature.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  set provider(provider: Provider) {
+    this._provider = provider;
+  }
+
+  get provider(): Provider {
+    return this._provider;
+  }
+
+  set context(context: Context) {
```
justinabrahms, line 52: Is your thinking that you can set an evaluation context on the api as well? I seemed to remember that we had a cascade of these where you could set the context on the API and client too, but I couldn't find it in the spec.

src/no-op-provider.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const REASON_NO_OP = 'No-op';
+
+/**
+ * The No-op provider is set by default, and simply always returns the default value.
+ */
+class NoopFeatureProvider implements Provider {
+  readonly name = 'No-op Provider';
+
+  isEnabled(_: string, defaultValue: boolean): Promise<boolean> {
```
beeme1mr, line 11: I thought we got rid of `isEnabled`

src/open-feature.ts
```diff
@@ -0,0 +1,59 @@
...
+// use a symbol as a key for the global singleton
+const GLOBAL_OPENFEATURE_API_KEY = Symbol.for('@openfeature/js.api');
+
+type OpenFeatureGlobal = {
+  [GLOBAL_OPENFEATURE_API_KEY]?: OpenFeature;
+};
+const _global = global as OpenFeatureGlobal;
+
+export class OpenFeature implements EvaluationLifeCycle {
+  private _provider: Provider = NOOP_PROVIDER;
```
beeme1mr, line 14: Does the linter enforce an underscore prefix for private properties? If not, it might be nice to add it so names stay consistent.

src/types.ts
```diff
@@ -0,0 +1,191 @@
+export type EvaluationContext = { targetingKey?: string } & Record<string, string | number | boolean | Date>;
```
beeme1mr, line 1: Are you waiting on the context PR to be merged before adding the rest? Also, can you add JSDocs above each value to explain how they're used?

src/client.ts
```diff
@@ -0,0 +1,157 @@
...
+  EvaluationContext,
+  EvaluationDetails,
+  FlagEvaluationOptions,
+  FlagValue,
+  Hook,
+  ResolutionDetails,
+  TransformingProvider,
+} from './types';
+
+type OpenFeatureClientOptions = {
```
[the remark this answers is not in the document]
toddbaert: > ```ts
> const client = openfeature.getClient({
>   name: "myClient", 
>   version: "1.0.1",
>   evaluationContext: { myContext: "test" }
> });
> ```

`evaluationContext` isn't an "option" to me, so I'm not crazy about having it in the options object. 

Thinking about it more, shouldn't options come last? Then you could be passing in an undefined or empty context (which makes more sense to me) if you want to set the name/version.

I don't really feel strongly either way.

src/no-op-provider.ts
toddbaert: We did, I will remove this. I auto implemented the interface, and at the time I did that is had this method.

src/open-feature.ts
toddbaert: I will look into adding that, I like it too.

src/types.ts
toddbaert: >Are you waiting on the context PR to be merged before adding the rest? 

Yes.

>  Also, can you add JSDocs above each value to explain how they're used?

What do you mean when you say "value"? Each interface?

src/client.ts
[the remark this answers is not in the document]
toddbaert: > Or we could do something like this:

```const client = openfeature.getClient({ myContext: "test"}, { name: "myClient", version: "1.0.1" })
```

Ya, that's what I was meaning, when I said "shouldn't options come last". As in, first the context, then the options.

src/open-feature.ts
justinabrahms: I was looking for it yesterday. :) We can accomplish the same with hooks though it will be maybe a little jankier. 

If/when we add it, there are some things that will need to be updated around hook context merges. No rush on my part for now.

src/types.ts
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * Interface that providers must implement to resolve flag values for their particular
+ * backend or vendor.
+ *
+ * Implementation for resolving all the required flag types must be defined.
+ *
+ * Additionally, a ContextTransformer function that transforms the OpenFeature context to the requisite user/context/attribute representation (typeof T)
+ * may also be implemented. This function will run immediately before the flag value resolver functions, appropriately transforming the context.
+ */
+export type Provider<T extends EvaluationContext | unknown = EvaluationContext> = T extends EvaluationContext
```
InTheCloudDan, line 161: While I'm in favor of this approach for a context transformer, is it in the spec at this time? If not do we want to move it out until it is supported?

justinabrahms:
It's in the spec as a should.

src/types.ts
toddbaert: https://github.com/open-feature/node-sdk/pull/8#issuecomment-1139637255

As Justin says, it's optional. I think with the recent conversations about the things like certain chars not being valid for particular vendor attribute keys, it might actually be more important to eventually make required.

beeme1mr (approved):

PR #10, opened 2022-05-30
Implement hooks

toddbaert:
Adds hooks, as per spec, along with tests.

Closes: https://github.com/open-feature/node-sdk/issues/7

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -6,7 +6,9 @@ import {
   EvaluationDetails,
   FlagEvaluationOptions,
   FlagValue,
+  FlagType,
   Hook,
+  HookContext,
   ResolutionDetails,
   TransformingProvider,
 } from './types.js';
@@ -20,6 +22,7 @@ export class OpenFeatureClient implements Client {
   name?: string | undefined;
   version?: string | undefined;
   readonly context: EvaluationContext;
+  private _hooks: Hook[] = [];
 
   constructor(private readonly api: OpenFeature, options: OpenFeatureClientOptions, context: EvaluationContext = {}) {
     this.name = options.name;
@@ -27,13 +30,16 @@ export class OpenFeatureClient implements Client {
     this.context = context;
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   addHooks(...hooks: Hook<FlagValue>[]): void {
-    throw new Error('Method not implemented.');
+    this._hooks = [...this._hooks, ...hooks];
   }
 
   get hooks(): Hook<FlagValue>[] {
-    throw new Error('Method not implemented.');
+    return this._hooks;
+  }
+
+  clearHooks(): void {
+    this._hooks = [];
   }
 
   async getBooleanValue(
@@ -51,7 +57,14 @@ export class OpenFeatureClient implements Client {
     context?: EvaluationContext,
     options?: FlagEvaluationOptions
   ): Promise<EvaluationDetails<boolean>> {
-    return this.evaluate<boolean>(flagKey, this.provider.resolveBooleanEvaluation, defaultValue, context, options);
+    return this.evaluate<boolean>(
+      flagKey,
+      this.provider.resolveBooleanEvaluation,
+      defaultValue,
+      'boolean',
+      context,
+      options
+    );
   }
 
   async getStringValue(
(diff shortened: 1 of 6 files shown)


src/client.ts
```diff
@@ -117,41 +144,117 @@ export class OpenFeatureClient implements Client {
...
 
     try {
-      // if a transformer is defined, run it to prepare the context.
+      const mergedHookContext = await this.beforeHooks(allHooks, hookContext, options);
+
+      // merge context in order: global, client, hook, invocation
+      const mergedContext = {
+        ...mergedHookContext,
+        ...invocationContext,
+      };
```
toddbaert, line 180: Since the spec dictates invocation context "wins" over any hooks, I merge invocation context attributes after merging that from before hooks. I think doing it any other way would be inconsistent and confusing.

@justinabrahms  @beeme1mr

src/client.ts
```diff
@@ -117,41 +144,117 @@ export class OpenFeatureClient implements Client {
...
+    for (const hook of hooks) {
+      // freeze the hookContext
+      Object.freeze(hookContext);
+
+      // use Object.assign to avoid modification of frozen hookContext
+      Object.assign(hookContext.context, {
+        ...hookContext.context,
+        ...(await hook?.before?.(hookContext, Object.freeze(options.hookHints))),
+      });
+    }
```
toddbaert, line 225: Normally I don't like doing await in loops, however, I think if we want to support async hooks, we can't just do a `Promise.all` - that might result in unexpected order of evaluation.

@beeme1mr

src/types.ts
```diff
@@ -187,11 +191,59 @@ export interface Client extends EvaluationLifeCycle, Features {
...
+  readonly defaultValue: T;
+  readonly flagType: FlagType;
+  readonly context: Readonly<EvaluationContext>;
+  readonly client: Client;
+  readonly provider: Provider;
+}
+
+export interface BeforeHookContext extends HookContext {
+  context: EvaluationContext;
+}
```
toddbaert, line 207: The `before` hook now has a slightly different type to denote the fact that the context is mutable only in this case (other hooks can't mutate context, per spec).

src/client.ts
justinabrahms: That seems correct to me. Here's my implementation: 

https://github.com/open-feature/java-sdk/blob/main/lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java#L61

And here is where we update the before hook context: https://github.com/open-feature/java-sdk/blob/main/lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java#L121

test/open-feature.spec.ts
```diff
@@ -3,39 +3,43 @@ import { OpenFeature } from '../src/open-feature';
...
 
   describe('Requirement 1.3', () => {
     it('should allow addition of hooks', () => {
-      // TODO: implement with hooks
+      expect(OpenFeature.addHooks).toBeDefined();
     });
   });
 
   describe('Requirement 1.4', () => {
     it('should implement a hook accessor', () => {
```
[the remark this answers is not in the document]
toddbaert: Agreed, fixed!

src/open-feature.ts
```diff
@@ -28,32 +30,35 @@ export class OpenFeature implements EvaluationLifeCycle {
...
 
-  set provider(provider: Provider) {
-    this._provider = provider;
+  static set provider(provider: Provider) {
+    this.instance._provider = provider;
   }
 
-  get provider(): Provider {
-    return this._provider;
+  static get provider(): Provider {
```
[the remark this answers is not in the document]
toddbaert: Noted.

beeme1mr (approved):
LGTM

src/open-feature.ts
```diff
@@ -10,14 +10,16 @@ type OpenFeatureGlobal = {
 };
 const _global = global as OpenFeatureGlobal;
 
-export class OpenFeature implements EvaluationLifeCycle {
+// TODO: make implement EvaluationLifeCycle, but statically...
```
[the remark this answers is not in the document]
toddbaert: Created https://github.com/open-feature/node-sdk/issues/13

PR #18, opened 2022-06-03
Change FlagValue to FlagValueType

toddbaert:
see title.


beeme1mr (approved):
Thanks for making this change. It's now more clear what this type represents.

PR #20, opened 2022-06-09
Remove provider/client refs, add metadata

toddbaert:
Removes refs to provider, client in `hookContext` (replacing with metadata), and provider accessor from API.


beeme1mr (approved):
Looks good, thanks!

PR #25, opened 2022-06-14
Update error handling, update spec refs

toddbaert:
- updates error handling as per: https://github.com/open-feature/spec/blob/main/specification/flag-evaluation/hooks.md#requirement-443, https://github.com/open-feature/spec/blob/main/specification/flag-evaluation/hooks.md#requirement-444
- updates spec references


src/client.ts
```diff
@@ -251,14 +251,26 @@ export class OpenFeatureClient implements Client {
   private async errorHooks(hooks: Hook[], hookContext: HookContext, err: unknown, options: FlagEvaluationOptions) {
     // run "error" hooks sequentially
     for (const hook of hooks) {
-      await hook?.error?.(hookContext, err, options.hookHints);
+      try {
+        await hook?.error?.(hookContext, err, options.hookHints);
+      } catch (err) {
+        // TODO: replace with injected logger
+        console.error(`Unhandled error during 'error' hook: ${err}`);
```
weyert, line 258: Unrelated to this PR but I think we could borrow the idea from opentelemetry for this:
https://github.com/open-telemetry/opentelemetry-js-api/blob/main/src/diag/types.ts

Then people could make a little adapter for their dream logger (winston, pino etc)

toddbaert: :100: . We have an issue for this as well... if you're interested you can feel free to take it on - either the spec, or just do an "out-of-spec" implementation in the node-sdk which we can use as a reference.


https://github.com/open-feature/spec/issues/39

PR #26, opened 2022-06-14
feat: add common erors to sdk

weyert:
Adds the common errors to the Node SDK: `FlagNotFoundError`, `ParseError`
and `TypeMismatchError`


src/errors/error.abstract.ts
```diff
@@ -0,0 +1,5 @@
+import type { ErrorCode } from './codes'
+
+export abstract class OpenFeatureError extends Error {
```
toddbaert, line 3: Take a look at this: https://github.com/Microsoft/TypeScript-wiki/blob/main/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work

weyert: It's part of the Error class that inherit from this abstract class. I think that's sufficient

toddbaert: In older targets, it can cause `instanceof` to break, as well as other things, as the article says:

Here's a playground example:

https://www.typescriptlang.org/play?target=1#code/C4TwDgpgBAQghgJwCrmgXigJQgYwPYIAmAPAM7AICWAdgOYA0UArtQNbV4Du1AfFAGRZcBEgHIARolGMW7LrwDcAKCVxx5BHBzAoOADZxSpKAHlI1AGIQ4wJgggBRBAgJQIAD2ARqhY05cIUADeSlBQahpaOviEEABcUBo0tEoAvir6hsZIEOT+rh5ePsZm3lY2do7OriFhdXX41BpM2gQAFAC2CUl0AJTBofV1pEyQCJ29yvXp9THoUABEpHgdEAC0cwtpGXhNeHoQAHR6eLRt1BCcUDl51eOiq0ZwtBCi-TTkcNQ4EHgAZqZzOVbPZ8ghekA

beeme1mr (approved):
Looks good to me! Thanks for adding this.

toddbaert (approved):

Issue #21, opened 2022-06-13
Standard reasons not included


weyert:
The specification discusses some default reasons such as `TARGETING_KEY_MISSING` and similar. I think it would be beneficial to define these as a n enum so they can be referred to by provider implementations

beeme1mr:
Hey @weyert, can this issue be closed?

PR #27, opened 2022-06-15
feat: add standard flag resolution codes

weyert:
Introduces a default set of flag resolution codes as defined in the spec

refs #21

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -4,7 +4,6 @@
  */
 export type JSONValue = null | string | number | boolean | Date | { [x: string]: JSONValue } | Array<JSONValue>;
 
-
 export type EvaluationContext = {
   /**
    * A string uniquely identifying the subject (end-user, or client service) of a flag evaluation.
@@ -182,10 +181,43 @@ export interface ProviderOptions<T = unknown> {
   contextTransformer?: ContextTransformer<T>;
 }
 
+export enum StandardResolutionReasons {
+  /**
+   * Indicates that the feature flag is targeting
+   * 100% of the targeting audience,
+   * e.g. 100% rollout percentage
+   */
+  TARGETING_MATCH = 'TARGETING_MATCH',
+  /**
+   * Indicates that the feature flag is targeting
+   * a subset of the targeting audience,
+   * e.g. less than 100% rollout percentage
+   */
+  SPLIT = 'SPLIT',
+  /**
+   * Indicates that the feature flag is disabled
+   */
+  DISABLED = 'DISABLED',
+  /**
+   * Indicates that the default value for
+   * the feature flag is returned
+   */
+  DEFAULT = 'DEFAULT',
+  /**
+   * Indicates an unknown issue occurred during evaluation
+   */
+  UNKNOWN = 'UNKNON',
+  /**
+   * Indicates that an error occurred during evaluation
+   */
+  ERROR = 'ERROR',
+}
+export type ResolutionReason = StandardResolutionReasons | string;
+
 export type ResolutionDetails<U> = {
   value: U;
   variant?: string;
-  reason?: string;
+  reason?: ResolutionReason;
   errorCode?: string;
 };
 


justinabrahms (approved):

src/types.ts
```diff
@@ -182,10 +181,43 @@ export interface ProviderOptions<T = unknown> {
...
+  DISABLED = 'DISABLED',
+  /**
+   * Indicates that the default value for
+   * the feature flag is returned
+   */
+  DEFAULT = 'DEFAULT',
+  /**
+   * Indicates an unknown issue occurred during evaluation
+   */
+  UNKNOWN = 'UNKNON',
```
beeme1mr, line 209:
```suggestion
  UNKNOWN = 'UNKNOWN ',
```

src/types.ts
```diff
@@ -182,10 +181,43 @@ export interface ProviderOptions<T = unknown> {
...
+  /**
+   * Indicates that an error occurred during evaluation
+   */
+  ERROR = 'ERROR',
+}
+export type ResolutionReason = StandardResolutionReasons | string;
```
beeme1mr, line 215: TypeScript will treat this as a string. That means you won't get auto completion in the IDE. That's not a dealbreaker but perhaps something like this would work better.

```typescript
export type ResolutionReason = keyof typeof StandardResolutionReasons | (string & Record<never, never>)
```

This isn't prefect either since it's using the key and not the value. It also may be a bit more difficult to understand what it's doing. It's up to you.

weyert: Cool, didn't even think about that. I have updated it

beeme1mr (approved):
Looks good to me!

src/types.ts
```diff
@@ -182,10 +181,43 @@ export interface ProviderOptions<T = unknown> {
...
+   * Indicates that the default value for
+   * the feature flag is returned
+   */
+  DEFAULT = 'DEFAULT',
+  /**
+   * Indicates an unknown issue occurred during evaluation
+   */
+  UNKNOWN = 'UNKNOWN',
+  /**
+   * Indicates that an error occurred during evaluation
```
toddbaert, line 211: It _may_ be worth adding that the `errorCode` field contains the details of this error in this comment.

toddbaert (approved):
I only have minor thoughts/concerns, consider them, but LGTM

src/types.ts
```diff
@@ -182,10 +181,43 @@ export interface ProviderOptions<T = unknown> {
...
+  SPLIT = 'SPLIT',
+  /**
+   * Indicates that the feature flag is disabled
+   */
+  DISABLED = 'DISABLED',
+  /**
+   * Indicates that the default value for
+   * the feature flag is returned
+   */
+  DEFAULT = 'DEFAULT',
```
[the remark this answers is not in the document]
beeme1mr: Adding fall-through may make sense to avoid confusion.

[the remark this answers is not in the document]
toddbaert: Ya, unless we can come up with another term... for when we return the "set" value of a flag. We could potentially do `PRESET` or `STATIC` or something like that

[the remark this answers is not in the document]
weyert: I am happy with both. Any preference?

weyert:
I have updated the wording. Do you want me to add the new constant/enum now?

toddbaert:
> I have updated the wording. Do you want me to add the new constant/enum now?

I think so, if it makes sense to you, otherwise we can do it later. As this stands, it's a good improvement I think. Up to you to decide if you like the additional enum idea.

weyert:
@toddbaert @beeme1mr I have updated the wording and added the `STATIC` enum :)

src/types.ts
```diff
@@ -182,10 +181,53 @@ export interface ProviderOptions<T = unknown> {
...
+  /**
+   * Indicates that the feature flag evaluated to the
+   * default value as passed in getBooleanValue/getBooleanValueDetails and
+   * similar functions in the Client   */
+  DEFAULT = 'DEFAULT',
+  /**
+   * Indicates that the feature flag evaluated to a 
+   * static value, for example, the default value for the flag
+   * 
+   * Note: Typically means  that no dynamic evaluation has been
```
beeme1mr, line 210:
```suggestion
   * Note: Typically means that no dynamic evaluation has been
```

toddbaert:
@weyert thanks again!

Issue #13, opened 2022-05-31
Make OpenFeature class implement EvaluationLifeCycle interface


toddbaert:
The OpenFeature API object (class) implements the methods defined on `EvaluationLifeCycle`, but doesn't implement the interface, because static members cannot implement an interface (at least trivially). This means there could be drift between the hook methods on the client, and the API.

We should find a solution for this, either with some typing magic, or by perhaps exporting an instance directly instead of using static methods.

anudeepreddy:
Hi,

Could we do something like below:
```js
export const OpenFeature: EvaluationLifeCycle = _OpenFeature;
                //↑____ Error is thrown here when _OpenFeature doesn't implement the methods and properties from EvaluationLifeCycle
```
This approach does the job of enforcing the contract but I am not sure if this is the right way to achieve it.

toddbaert:
I actually think the right approach might be to NOT export a class with static methods, but export an instance itself. That instance can be a class that we don't allow to be instantiated outside the API code.

anudeepreddy:
I have worked on this. I have made the following changes:
- Made the OpenFeature class implement EvaluationLifeCycle.
- made the getInstance static method public.
- changes the remaining static methods to non-static.

Should I go ahead and raise a PR?

toddbaert:
@anudeepreddy go ahead and open a PR. The only thing that sounds like it could be a problem is: `made the getInstance static method public`.

I think instead I would just create an instance and make that the export.

toddbaert:
Closed with https://github.com/open-feature/node-sdk/pull/44

PR #44, opened 2022-07-01
made OpenFeature implement EvaluationLifeCycle

anudeepreddy:
Signed-off-by: Anudeep Reddy <anudeepc85@gmail.com>

Hello👋,

I have worked on #13, the following are the changes that are included in the PR:
- Made the OpenFeature class implement EvaluationLifeCycle.
- made the getInstance static method public.
- changes the remaining static methods to non-static.

I would appreciate any feedback.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -23,7 +23,7 @@ export class OpenFeatureClient implements Client {
   readonly metadata: ClientMetadata;
   readonly context: EvaluationContext;
   private _hooks: Hook[] = [];
-
+  private openFeature: OpenFeature;
   constructor(
     // we always want the client to use the current provider,
     // so pass a function to always access the currently registered one.
@@ -36,6 +36,7 @@ export class OpenFeatureClient implements Client {
       version: options.version,
     } as const;
     this.context = context;
+    this.openFeature = OpenFeature.getInstance();
   }
 
   addHooks(...hooks: Hook<FlagValue>[]): void {
@@ -158,12 +159,12 @@ export class OpenFeatureClient implements Client {
   ): Promise<EvaluationDetails<T>> {
     // merge global, client, and evaluation context
 
-    const allHooks = [...OpenFeature.hooks, ...this.hooks, ...(options.hooks || [])];
+    const allHooks = [...this.openFeature.hooks, ...this.hooks, ...(options.hooks || [])];
     const allHooksReversed = [...allHooks].reverse();
 
     // merge global and client contexts
     const globalAndClientContext = {
-      ...OpenFeature.context,
+      ...this.openFeature.context,
       ...this.context,
     };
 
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -23,7 +23,7 @@ export class OpenFeatureClient implements Client {
   readonly metadata: ClientMetadata;
   readonly context: EvaluationContext;
   private _hooks: Hook[] = [];
-
+  private openFeature: OpenFeature;
```
toddbaert, line 26:
```suggestion

  private openFeature: OpenFeature;
```

src/open-feature.ts
```diff
@@ -11,15 +11,15 @@ type OpenFeatureGlobal = {
 const _global = global as OpenFeatureGlobal;
 
 // TODO: make implement EvaluationLifeCycle, but statically...
-export class OpenFeature {
+export class OpenFeature implements EvaluationLifeCycle {
```
toddbaert, line 14: Don't export this class at all, instead, just add something like this at the bottom of the file:

```
export const openFeature = OpenFeature.getInstance();
```

This way, we are EXPORTING an instance, and nobody should (nobody will be able) to call `OpenFeature.getInstance`.

Then, consuming code will just do:

`import { openFeature } from './open-feature.js'`, which will already be a singleton!

toddbaert (changes requested):
Great work so far! Please just export a singleton instead of the class though.

toddbaert:
@weyert This change will be breaking only in that it exports a lower case `openFeature` (an instance) instead of `OpenFeature` (a class). This is mostly a style choice, but I think it's the correct one and now is the time to do it. You should only have to change the case, nothing else should break.

weyert:
Why not rename the class to something different so that it doesn’t conflict with `OpenFeature`? To me, capitalised name indicates most often it’s a global/singleton instance

toddbaert:
> Why not rename the class to something different so that it doesn’t conflict with `OpenFeature`? To me, capitalised name indicates most often it’s a global/singleton instance

We no longer export the `OpenFeature` class, we just export this instance, which is better I think. It was supposed to be used as a singleton, so just exporting one is better.

I don't really mind if we export `OpenFeature` vs `openFeature`. I've seen both in singletons, TBH.

anudeepreddy:
> I don't really mind if we export OpenFeature vs openFeature. I've seen both in singletons, TBH.

What should be the name of the class if we export the singleton as OpenFeature? In case if Java and python it is named OpenFeatureAPI, should the same be done here aswell?

toddbaert:
> What should be the name of the class if we export the singleton as OpenFeature? In case if Java and python it is named OpenFeatureAPI, should the same be done here as well?

Yep, I think that's fine.

toddbaert (approved):
LGTM. @weyert I think this new version will have zero breaking changes.

anudeepreddy:
Looking forward to making more contributions 😄

beeme1mr:
Thanks @anudeepreddy!

PR #64, opened 2022-07-08
Add Codecov upload step

beeme1mr:
Signed-off-by: Michael Beemer <michael.beemer@dynatrace.com>

### DIFF
--- a/.github/workflows/publish-dev.yaml
+++ b/.github/workflows/publish-dev.yaml
@@ -32,6 +32,12 @@ jobs:
       - name: Build
         run: npm run build
 
+      - name: Test
+        run: npm run test
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
+
       - name: Configure git
         run: |
           git config --global user.name 'openfeature-ci'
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-dev.yaml
```diff
@@ -32,6 +32,12 @@ jobs:
       - name: Build
         run: npm run build
 
+      - name: Test
+        run: npm run test
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
+
```
toddbaert, line 40: Do we need this here? We didn't have lint or test here because presumably that stuff was already done in the pr-checks.

beeme1mr: We need to run code coverage on the main branch so Codecov can generate a coverage diff.

toddbaert: ah, OK. If it's not a requirement to the actual publish, it could be in it's own job then, so we don't slow the publish job down, but not a huge deal.

toddbaert (approved):

PR #70, opened 2022-07-12
Add cucumber integration tests

toddbaert:
This PR adds a PoC for a cross-language integration test suite. The suite works by running a [flagd image](https://github.com/open-feature/test-harness/blob/main/Dockerfile) with an extra layer of config.

Specifically, the PR:

- Adds sample gerkin tests. These are committed to this branch, but they should reside somewhere else, so all SDKs can use them. We certainly need more tests, but they are quite easy to write. These are just to prove the concept.
- Adds a jest test suite that parses and runs the gerkin.
- Adds CI which runs a custom flagd container with the flags in question.

My main question is where this gerkin should reside... It could be in the [test harness repo](https://github.com/open-feature/test-harness), or it could even be embedded in the spec and parsed out by our spec parser.

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -10,6 +10,12 @@ jobs:
   build-test-lint:
     runs-on: ubuntu-latest
 
+    services:
+      flagd:
+        image: ghcr.io/open-feature/flagd-testbed:latest
+        ports:
+          - 8080:8080
+
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
@@ -17,11 +23,14 @@ jobs:
       - name: Install
         run: npm ci
 
+      - name: Build
+        run: npm run build
+
       - name: Lint
         run: npm run lint
 
-      - name: Build
-        run: npm run build
+      - name: Integration
+        run: npm run integration
 
       - name: Test
         run: npm run test
--- a/integration/features/evaluation.feature
+++ b/integration/features/evaluation.feature
@@ -1,0 +1,16 @@
+Feature: Flag evaluation
+
+  Scenario: Resolves boolean value
+    Given A boolean flag called boolean-flag with value true exists
+    When Flag is evaluated with default value false
+    Then The resolved value should match the flag value
+
+  Scenario: Resolves string value
+    Given A string flag called string-flag with value #CC0000 exists
+    When Flag is evaluated with default value #0000CC
+    Then The resolved value should match the flag value
+
+  Scenario: Resolves number value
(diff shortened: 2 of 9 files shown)


tsconfig.json
```diff
@@ -28,7 +28,15 @@
     // "rootDir": "./",                                  /* Specify the root folder within your source files. */
     // "moduleResolution": "node",                       /* Specify how TypeScript looks up a file from a given module specifier. */
     // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
-    // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
+
```
beeme1mr, line 31: Revert

integration/features/evaluation.feature
```diff
@@ -0,0 +1,16 @@
...
+
+  Scenario: Resolves string value
+    Given A string flag called string-flag with value #CC0000 exists
+    When Flag is evaluated with default value #0000CC
+    Then The resolved value should match the flag value
+
+  Scenario: Resolves number value
+    Given A number flag called number-flag with value 1 exists
+    When Flag is evaluated with default value 2
+    Then The resolved value should match the flag value
```
toddbaert, line 16: This cannot reside here long term, it has to be in some common repo, and probably included as a submodule. It could be parsed out of the spec, or committed [here](https://github.com/open-feature/test-harness)

integration/step-definitions/evaluation.spec.ts
```diff
@@ -0,0 +1,68 @@
+import { defineFeature, loadFeature } from 'jest-cucumber';
+import { OpenFeature } from '../../src/open-feature.js';
+
+// load the feature file.
+const feature = loadFeature('integration/features/evaluation.feature');
+
+// get a client (flagd provider registered in setup)
+const client = OpenFeature.getClient();
+
+defineFeature(feature, (test) => {
```
toddbaert, line 10: These tests are associated with the gerkin. If any test descriptions don't line up with the gerkin clauses, the test fails.

.github/workflows/pr-checks.yaml
```diff
@@ -7,21 +7,30 @@ on:
...
-  build-test-lint:
+  build-lint-integration-test:
     runs-on: ubuntu-latest
 
+    services:
+      flagd:
+        image: ghcr.io/open-feature/flagd-testbed:latest
+        ports:
+          - 8080:8080
+
```
toddbaert, line 18: runs the custom flagd image in CI

toddbaert:
This was implemented in another PR

PR #148, opened 2022-08-09
feat!: remove ctx transformer, add provider hooks

toddbaert:
- remove context transformer
- adds provider hooks
- updates to new hook ordering
- updates tests accordingly, included referenced spec numbers

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -10,8 +10,8 @@ import {
   FlagValueType,
   Hook,
   HookContext,
+  Provider,
   ResolutionDetails,
-  TransformingProvider,
 } from './types';
 
 type OpenFeatureClientOptions = {
@@ -27,7 +27,7 @@ export class OpenFeatureClient implements Client {
   constructor(
     // we always want the client to use the current provider,
     // so pass a function to always access the currently registered one.
-    private readonly providerAccessor: () => TransformingProvider<unknown>,
+    private readonly providerAccessor: () => Provider,
     options: OpenFeatureClientOptions,
     context: EvaluationContext = {}
   ) {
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -175,26 +176,14 @@ export class OpenFeatureClient implements Client {
...
-        ...mergedHookContext,
-        ...invocationContext,
-      };
-
-      // if a transformer is defined, run it to prepare the context
-      const transformedContext =
-        typeof this.provider.contextTransformer === 'function'
-          ? await this.provider.contextTransformer(mergedContext)
-          : mergedContext;
+      const frozenContext = await this.beforeHooks(allHooks, hookContext, options);
```
toddbaert, line 183: chose this name because the context object is actually frozen (immutable here) after the before hooks have run...

src/types.ts
```diff
@@ -104,11 +104,12 @@ export interface Features {
...
- * Function which transforms the EvaluationContext to a type useful for the provider.
+ * Interface that providers must implement to resolve flag values for their particular
+ * backend or vendor.
+ *
+ * Implementation for resolving all the required flag types must be defined.
  */
-export type ContextTransformer<T = unknown> = (context: EvaluationContext) => T;
-
-interface GenericProvider<T> {
+export interface Provider extends Pick<Partial<EvaluationLifeCycle>, 'hooks'> {
```
toddbaert, line 112: @beeme1mr this is the way I've represented the hooks on the provider. I chose this for a few reasons...

We don't want any one to be able to ADD hooks, so I've just added a hook property getter... then the hook author can add hooks at the point they define their provider. We don't want anyone else attaching hooks with an `addHooks()` method, or clearning them with `clearHooks()`.

I also don't want to use a base class, interfaces are always more flexible.

beeme1mr: I approve

benjiro (approved):
LGTM

beeme1mr (approved):

toddbaert:
Tested this in the playground, looks good.

PR #187, opened 2022-08-25
fix: update the client to conform to the provider interface

beeme1mr:
The resolve argument in the evaluation method was still using evaluation options.


toddbaert (approved):
LGTM - sorry for missing on the last one.

PR #212, opened 2022-09-12
feat: add transaction propagation

beeme1mr:

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -194,6 +194,7 @@ export class OpenFeatureClient implements Client {
     // merge global and client contexts
     const mergedContext = {
       ...OpenFeature.getContext(),
+      ...OpenFeature.getTransactionContext(),
       ...this._context,
       ...invocationContext,
     };
--- a/src/open-feature.ts
+++ b/src/open-feature.ts
@@ -94,9 +94,9 @@ class OpenFeatureAPI implements GlobalApi {
 
   setTransactionContextPropagator(transactionContextPropagator: TransactionContextPropagator): OpenFeatureAPI {
     const baseMessage = 'Invalid TransactionContextPropagator, will not be set: ';
-    if (typeof transactionContextPropagator.getTransactionContext !== 'function') {
+    if (typeof transactionContextPropagator?.getTransactionContext !== 'function') {
       this._logger.error(`${baseMessage}: getTransactionContext is not a function.`);
-    } else if (typeof transactionContextPropagator.setTransactionContext !== 'function') {
+    } else if (typeof transactionContextPropagator?.setTransactionContext !== 'function') {
       this._logger.error(`${baseMessage}: setTransactionContext is not a function.`);
     } else {
       this._transactionContextPropagator = transactionContextPropagator;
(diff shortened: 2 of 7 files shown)


src/types.ts
```diff
@@ -284,3 +284,22 @@ export interface Hook<T extends FlagValue = FlagValue> {
    */
   finally?(hookContext: Readonly<HookContext<T>>, hookHints?: HookHints): Promise<void> | void;
 }
+
+/**
+ * Transaction context is a mechanism for adding transaction specific context that
+ * is merged with evaluation context prior to flag evaluation. Examples of potential
+ * transaction specific context include: a user id, user agent, or IP address.
+ */
+export type TransactionContext = EvaluationContext;
```
beeme1mr, line 293: @toddbaert what do you think about this? We could use evaluation context but having a transaction concept type seemed nice.

src/types.ts
```diff
@@ -520,3 +527,56 @@ export interface Hook<T extends FlagValue = FlagValue> {
...
+
+export interface TransactionContextPropagator {
+  /**
+   * EXPERIMENTAL: Transaction context propagation is experimental and subject to change.
+   * The OpenFeature Enhancement Proposal regarding transaction context can be found [here](https://github.com/open-feature/ofep/pull/32).
+   *
+   * Returns the currently defined transaction context using the registered transaction
+   * context propagator.
+   *
+   * @returns {TransactionContext} The currently defined transaction context
```
toddbaert, line 562:
```suggestion
   * @returns {TransactionContext} The current transaction context
```

src/open-feature.ts
```diff
@@ -78,6 +91,23 @@ class OpenFeatureAPI implements GlobalApi {
...
+  setTransactionContext<R>(
+    transactionContext: TransactionContext,
+    callback: (...args: unknown[]) => R,
+    ...args: unknown[]
+  ): void {
+    this._transactionContextPropagator.setTransactionContext(transactionContext, callback, ...args);
+  }
+
+  getTransactionContext(): TransactionContext {
+    return this._transactionContextPropagator.getTransactionContext();
```
toddbaert, line 109: We should think about whether it's better to throw (and therefore default) if something goes wrong here, or just ignore the transaction context.

For example, if `getTransactionContext` is not a function, that will be an exception, which will cause a default.... or we could check that is a function, and return undefined. I think throwing/defaulting is PROBABLY better, but I'm not sure.

We could also do some validation at registration time like we do for the logger.

src/client.ts
```diff
@@ -194,6 +194,7 @@ export class OpenFeatureClient implements Client {
     // merge global and client contexts
     const mergedContext = {
       ...OpenFeature.getContext(),
+      ...OpenFeature.getTransactionContext(),
```
toddbaert, line 197: This needs to be in the try/catch now I think. `getTransactionContext()` runs user code.

toddbaert (changes requested):
https://github.com/open-feature/js-sdk/pull/212/files#r984831394

toddbaert (approved):

src/open-feature.ts
toddbaert: fixed

src/client.ts
toddbaert: fixed

PR #219, opened 2022-09-14
feat: add logger

beeme1mr:
Features include:

- A default logger that writes warning and errors to the console.
- A safe logger that will wrap user providers loggers to ensure stability
- Global logger can be updated with a custom logger
- Clients can be updated with a customer logger
- Custom client loggers override global loggers

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,5 +1,6 @@
 import { ERROR_REASON, GENERAL_ERROR } from './constants';
 import { OpenFeature } from './open-feature';
+import { SafeLogger } from './logger';
 import {
   Client,
   ClientMetadata,
@@ -10,6 +11,7 @@ import {
   FlagValueType,
   Hook,
   HookContext,
+  Logger,
   Provider,
   ResolutionDetails,
 } from './types';
@@ -23,11 +25,13 @@ export class OpenFeatureClient implements Client {
   readonly metadata: ClientMetadata;
   private _context: EvaluationContext;
   private _hooks: Hook[] = [];
+  private _clientLogger?: Logger;
 
   constructor(
     // we always want the client to use the current provider,
     // so pass a function to always access the currently registered one.
     private readonly providerAccessor: () => Provider,
+    private readonly globalLogger: () => Logger,
     options: OpenFeatureClientOptions,
     context: EvaluationContext = {}
   ) {
@@ -38,11 +42,19 @@ export class OpenFeatureClient implements Client {
     this._context = context;
   }
 
-  set context (context: EvaluationContext) {
+  set logger(logger: Logger) {
+    this._clientLogger = new SafeLogger(logger);
+  }
+
+  get logger(): Logger {
+    return this._clientLogger || this.globalLogger();
+  }
+
+  set context(context: EvaluationContext) {
     this._context = context;
   }
 
-  get context (): EvaluationContext {
+  get context(): EvaluationContext {
     return this._context;
   }
 
@@ -153,7 +165,12 @@ export class OpenFeatureClient implements Client {
 
   private async evaluate<T extends FlagValue>(
     flagKey: string,
-    resolver: (flagKey: string, defaultValue: T, context: EvaluationContext) => Promise<ResolutionDetails<T>>,
+    resolver: (
+      flagKey: string,
+      defaultValue: T,
+      context: EvaluationContext,
+      logger: Logger
+    ) => Promise<ResolutionDetails<T>>,
     defaultValue: T,
     flagType: FlagValueType,
     invocationContext: EvaluationContext = {},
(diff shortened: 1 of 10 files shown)


src/client.ts
```diff
@@ -246,9 +264,8 @@ export class OpenFeatureClient implements Client {
       try {
         await hook?.error?.(hookContext, err, options.hookHints);
       } catch (err) {
-        // TODO: replace with injected logger
-        console.error(`Unhandled error during 'error' hook: ${err}`);
-        console.error((err as Error).stack);
+        this.logger.error(`Unhandled error during 'error' hook: ${err}`);
+        this.logger.error((err as Error)?.stack);
```
toddbaert, line 268: Good find here.

src/logger.ts
```diff
@@ -0,0 +1,55 @@
...
+    try {
+      for (const level of LOG_LEVELS) {
+        if (!logger[level] || typeof logger[level] !== 'function') {
+          throw new Error(`The provided logger is missing the ${level} method.`);
+        }
+      }
+      this.logger = logger;
+    } catch (err) {
+      console.error(err);
+      console.warn('Falling back to the default logger.');
```
toddbaert, line 30: I fee like these should either both be `warn` or `error`.

src/logger.ts
```diff
@@ -0,0 +1,55 @@
...
+  info(...args: unknown[]): void {
+    this.log('info', ...args);
+  }
+  debug(...args: unknown[]): void {
+    this.log('debug', ...args);
+  }
+
+  private log(level: keyof Logger, ...args: unknown[]) {
+    try {
+      this.logger[level](...args);
```
toddbaert, line 50: I actually think you  should call this with `.apply` or `.call` and pass the logger as the `thisArg`. If the logger method called any members, they would be undefined.

test/client.spec.ts
```diff
@@ -138,51 +138,51 @@ describe(OpenFeatureClient.name, () => {
...
           const booleanDetails = await client.getBooleanDetails(booleanFlag, defaultBooleanValue);
 
           expect(booleanDetails.value).toEqual(BOOLEAN_VALUE);
           expect(booleanDetails.variant).toEqual(BOOLEAN_VARIANT);
-          expect(MOCK_PROVIDER.resolveBooleanEvaluation).toHaveBeenCalledWith(booleanFlag, defaultBooleanValue, {});
+          expect(MOCK_PROVIDER.resolveBooleanEvaluation).toHaveBeenCalledWith(booleanFlag, defaultBooleanValue, {}, {});
         });
       });
 
-      describe(OpenFeatureClient.prototype.getStringDetails.name, () => {
```
toddbaert, line 153: This makes me sad. These tests are all brittle now.

I'm not really a fan of changing these just for particular tooling.

src/logger.ts
toddbaert: Nevermind, I think this is fine since you're still calling this as a method, not a function ref.

toddbaert (approved):

toddbaert:
Looks good to me, good testing.

test/logger.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      expect(mockedLogger.info.mock.calls).toEqual([
+        ['in before hook'],
+        ['resolving boolean value'],
+        ['in after hook'],
+        ['in finally hook'],
+      ]);
+    });
+
+    it('should use the default logger in the error hook', async () => {
```
toddbaert, line 158: Is this test name correct? It seems like were not using the default logger.

test/logger.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      expect(client.logger).toBeInstanceOf(DefaultLogger);
+    });
+
+    it('should provide a logger to the provider', async () => {
+      const client = OpenFeature.getClient();
+      const mockedLogger = new MockedLogger();
+      client.logger = mockedLogger;
+      await client.getBooleanValue('test', false);
+      expect(mockedLogger.info).toHaveBeenCalledWith('resolving boolean value');
```
toddbaert, line 140: I think a lot of duplicated strings in assertions could be constants.

test/logger.spec.ts
beeme1mr: Good catch.

test/logger.spec.ts
beeme1mr: Done

src/client.ts
dyladan, line 268: Technically any type can be caught. A `typeof` check would avoid this casting

src/logger.ts
```diff
@@ -0,0 +1,55 @@
...
+  warn(...args: unknown[]): void {
+    this.log('warn', ...args);
+  }
+  info(...args: unknown[]): void {
+    this.log('info', ...args);
+  }
+  debug(...args: unknown[]): void {
+    this.log('debug', ...args);
+  }
+
```
dyladan, line 47: These can be defined in a loop by your constructor and would make your output smaller. Not sure if minification is a priority for you.

beeme1mr: That's an interesting idea. I'll keep that in mind for future optimizations.

PR #221, opened 2022-09-15
feat!: convert to chainable methods

beeme1mr:
Updated various touchpoints in the global API and client to be chainable methods. This is a breaking change that is intended to provide a more consistent developer experience.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -42,34 +42,80 @@ export class OpenFeatureClient implements Client {
     this._context = context;
   }
 
-  set logger(logger: Logger) {
+  /**
+   * Sets a logger on the client. This logger supersedes to the global logger
+   * and is passed to various components in the SDK.
+   *
+   * @param {Logger} logger The logger to to be used
+   * @returns {OpenFeatureClient} OpenFeature Client
+   */
+  setLogger(logger: Logger): OpenFeatureClient {
     this._clientLogger = new SafeLogger(logger);
+    return this;
   }
 
-  get logger(): Logger {
-    return this._clientLogger || this.globalLogger();
-  }
-
-  set context(context: EvaluationContext) {
+  /**
+   * Sets client evaluation context that will be used during flag evaluations
+   * on this client.
+   *
+   * @param {EvaluationContext} context Client evaluation context
+   * @returns {OpenFeatureClient} OpenFeature Client
+   */
+  setContext(context: EvaluationContext): OpenFeatureClient {
     this._context = context;
+    return this;
   }
 
-  get context(): EvaluationContext {
+  /**
+   * Access the client evaluation context.
+   *
+   * @returns {EvaluationContext} Client evaluation context
+   */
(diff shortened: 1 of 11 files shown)


src/types.ts
```diff
@@ -116,8 +115,15 @@ export interface Features {
...
-export interface Provider extends Pick<Partial<EvaluationLifeCycle>, 'hooks'> {
+export interface Provider {
   readonly metadata: ProviderMetadata;
+  /**
+   * A provider hook exposes a mechanism for provider authors to register hooks
+   * to tap into various stages of the flag evaluation lifecycle. These hooks can
+   * be used to perform side effects and mutate the context for purposes of the
+   * provider. Provider hooks are not configured or controlled by the application author.
+   */
+  readonly hooks?: Hook[];
```
tcarrio, line 126: Either work, but is there any reason why this isn't following the same pattern of `getHooks`?

tcarrio (approved):
Just the one question on idioms for hooks across provider and client.

src/client.ts
```diff
@@ -42,34 +42,80 @@ export class OpenFeatureClient implements Client {
...
+    return this;
   }
 
-  get logger(): Logger {
-    return this._clientLogger || this.globalLogger();
-  }
-
-  set context(context: EvaluationContext) {
+  /**
+   * Sets client evaluation context that will be used during flag evaluations
```
toddbaert, line 58:
```suggestion
   * Sets the evaluation context that will be used during flag evaluations
```

src/client.ts
```diff
@@ -42,34 +42,80 @@ export class OpenFeatureClient implements Client {
...
+   *
+   * @returns {EvaluationContext} Client evaluation context
+   */
+  getContext(): EvaluationContext {
     return this._context;
   }
 
-  addHooks(...hooks: Hook<FlagValue>[]): void {
+  /**
+   * Adds client hooks that will run during flag evaluations on this client. Client
```
toddbaert, line 79:
```suggestion
   * Adds hooks that will run during flag evaluations on this client. Client
```

toddbaert (approved):
This looks good. I wasn't thrilled with an chain-able API at first, but now that I see it I definitely like it.

Added minor/optional commends on js-doc, but I like them as well. Nice work.

src/client.ts
beeme1mr: I think it' important to highlight that this **IS** scoped to the client. Perhaps it could be rephased to something like:
`Settings the evaluation context on the client so it can be used during flag evaluation.`

src/client.ts
beeme1mr: same as above. I think it's important to specify that these hooks are scoped to the client. Have to rephrase but I would like to include that information.

src/client.ts
beeme1mr: Sets the evaluation context that will be used during flag evaluations on this client.

PR #224, opened 2022-09-19
feat!: improve generic type accuracy

toddbaert:
- restricts generic args on object resolver (`T` is now `T extends JsonValue`)
- adds optional generic args for string/number resolver for unions of string/number literals

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -11,6 +11,7 @@ import {
   FlagValueType,
   Hook,
   HookContext,
+  JsonValue,
   Logger,
   Provider,
   ResolutionDetails,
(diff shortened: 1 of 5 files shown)


toddbaert:
rebased on main b/c of conflicts.

beeme1mr (approved):

kinyoklion (approved):
LGTM. Works with my provider (with a tiny update to match the new resolveObjectEvaluation signature.)

src/client.ts
```diff
@@ -154,38 +155,41 @@ export class OpenFeatureClient implements Client {
...
-   * @param {string} defaultValue The value returned if an error occurs
+   * @template {string} T A optional generic argument constraining the string
+   * @param {T} defaultValue The value returned if an error occurs
    * @param {EvaluationContext} context The evaluation context used on an individual flag evaluation
    * @param {FlagEvaluationOptions} options Additional flag evaluation options
-   * @returns {Promise<string>} Flag evaluation response
+   * @returns {Promise<T>} Flag evaluation response
    */
-  async getStringValue(
+  async getStringValue<T extends string = string>(
```
beeme1mr, line 164: I think this won't work as expected. The default value will be returned as a string literal.

```typescript
const value = await client.getStringValue("flag", "test");
```

Value type will be "test" not string.

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABFApgZygHgCqJQD1TABM1EBlKAJxjAHNEBeC62ugPgApaAHEKAFyJsASiG4A3gChEiKiiggqSXvykBfKVIgIMyAJ48U2dFCbJTnAEQALGFZEBuKUA

beeme1mr: It looks like it can be fixed with an overload.

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABFApgZygChmADiKALkQwCccBzASmLMoG4AoUSWBZdKAHgBVEUAHqjAATNCSjkwFAHzY8BYjxqIezcNHhJUGXvyEpR4utLk58RVSr4BvRokSkUUEKSTmCTAL6NGEBBjIAJ64KDyciAC8HBiYAEQAFjBxVEyMMMCImFAhYRGRBYhxEACGSABGKIhlQVBJ0imIdg7+YGhwADYoAHQdcBTxCShOKYw+fgFQwaHhGABMUTHcicmIAD5FQXBxcispaRlZOTOcCwXRcVuNzYit7V29-YPDKKNeQA

tcarrio: I think the goal here is trying to make it *possible* to provide types for the generic return type but default to simply "string"?

For the same reason @beeme1mr pointed out, there are issues with this approach. The return value should essentially be `string | typeof defaultValue`, which boils down to just `string` really when not providing a generic type explicitly.

The benefit here would be cases where perhaps we do know the possible types like maybe `"on"|"off"`.

toddbaert: > I think this won't work as expected. The default value will be returned as a string literal.

Ya, that's pretty interesting. The TS compiler is using what it knows about the default arg to make inferences about T, even though we are defaulting T to string.  It actually works as expected if `defaultValue` isn't a string literal, but some unknown string.

> I think the goal here is trying to make it possible to provide types for the generic return type but default to simply "string"?

Yep, that was the intent. Unfortunately due to the nature of the testing (`expect(...).toEqual(...);`), the problem you describe wasn't caught by the compiler. The new behavior works as expected but the functionality when not passing a generic explicitly is clearly broken.

toddbaert: > The return value should essentially be string | typeof defaultValue where the default value is really T, which boils down to just string really when not providing a generic type explicitly.

I couldn't get that to work, which makes sense to me since the union of `string | typeof defaultValue` these two would be basically a string, no matter how restricted `defaultValue` is, but it's possible I misunderstood your proposal.

For now I've used overloads something like @beeme1mr pointed out.

beeme1mr (approved):

Issue #242, opened 2022-09-27
Update SDK to be compliant with spec v0.5.0


beeme1mr:
Update the SDK to be compliant with the v0.5.0 spec.

### ⚠ BREAKING CHANGES

 * https://github.com/open-feature/spec/pull/142 Error codes must be a constrained set of values (e.g. enum).

### Additional changes

 * https://github.com/open-feature/spec/pull/140 Clarify that `reason` must be a string in the `flag resolution details` and `evaluation details`.
 * https://github.com/open-feature/spec/pull/142 Add optional `error message` to `flag resolution details`.
 
## General

 - [ ] Add/update a spec version badge on the readme.

[![Specification](https://img.shields.io/static/v1?label=Specification&message=v0.5.0&color=yellow)](https://github.com/open-feature/spec/tree/v0.5.0)

**Full Changelog**: https://github.com/open-feature/spec/compare/v0.4.0...v0.5.0

PR #244, opened 2022-09-28
feat!: errorCode as enum, reason as string

toddbaert:
- makes errorCode an enum
- makes reason a string
- adds errorMessage to resolution/evaluation details

This PR updates the SDK to be conformant to spec v0.5.0

Closes: https://github.com/open-feature/js-sdk/issues/242


src/client.ts
```diff
@@ -223,14 +224,17 @@ export class OpenFeatureClient implements Client {
 
       return evaluationDetails;
     } catch (err: unknown) {
-      const errorCode = (!!err && (err as { code: string }).code) || GENERAL_ERROR;
+
+      const errorMessage: string = (err as Error)?.message;
+      const errorCode: ErrorCode = (err as OpenFeatureError)?.code || ErrorCode.GENERAL;
```
toddbaert, line 229: Only behavioral change. Get code from `.code` or default to GENERAL, get message from `.message`.

src/errors/index.ts
```diff
@@ -1,5 +1,7 @@
-export * from './codes';
+export * from './general-error';
```
toddbaert, line 1: Extracted codes to type to remove circular import. `./types` should import nothing.

src/types.ts
```diff
@@ -272,55 +272,85 @@ export interface Provider {
   ): Promise<ResolutionDetails<T>>;
 }
 
-export enum StandardResolutionReasons {
+export const StandardResolutionReasons = {
```
toddbaert, line 275: I ended up using an object here because it's "less breaking" this way.

These are strings so we can still set them without referencing this structure.

weyert (approved):
Makes sense to me

test/client.spec.ts
```diff
@@ -1,7 +1,7 @@
-import { OpenFeatureClient } from '../src/client.js';
-import { ERROR_REASON, GENERAL_ERROR } from '../src/constants.js';
-import { OpenFeature } from '../src/open-feature.js';
-import { Client, EvaluationContext, EvaluationDetails, JsonArray, JsonObject, JsonValue, Provider, ResolutionDetails } from '../src/types.js';
+import { OpenFeatureClient } from '../src/client';
+import { ErrorCode, FlagNotFoundError } from '../src/index';
```
beeme1mr, line 2: Remove `/index`

src/types.ts
```diff
@@ -272,55 +272,85 @@ export interface Provider {
...
 }
+
 export type ResolutionReason = keyof typeof StandardResolutionReasons | (string & Record<never, never>);
 
 export type ResolutionDetails<U> = {
   value: U;
   variant?: string;
   reason?: ResolutionReason;
-  errorCode?: string;
+  errorCode?: ErrorCode;
```
beeme1mr, line 352: Add more advanced conditional types.

toddbaert: Will make a new PR for this. It's tricky because we accept any string as a reason, and that's matching `ERROR` when I don't want it to.

toddbaert: https://github.com/open-feature/js-sdk/issues/246

beeme1mr (approved):

PR #316, opened 2022-11-04
feat: eventing poc

toddbaert:
A PoC for eventing, meant to be used with the experimental flagd-web provider.

Specifically:

- adds an `addHandler()` method to the client which can be passed an event type (`PROVIDER_READY`, `PROVIDER_ERROR`,`PROVIDER_CONFIGURATION_CHANGED`, `PROVIDER_SHUTDOWN` and a function to run when that event is fired)
- adds a new `CACHED` default reason
- adds a new `EventProvider` interface, which providers can OPTIONALLY implement to fire events.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,4 +1,35 @@
+import EventEmitter from 'events';
+
 export type PrimitiveValue = null | boolean | string | number;
+
+export interface EventProvider {
+  readonly events: EventEmitter;
+  readonly ready: boolean;
+}
+
+export enum ProviderEvents {
+  Ready = 'PROVIDER_READY',
+  Error = 'PROVIDER_ERROR',
+  ConfigurationChanged = 'PROVIDER_CONFIGURATION_CHANGED',
+  Shutdown = 'PROVIDER_SHUTDOWN',
+};
+
+export enum ApiEvents {
+  ProviderChanged = 'providerChanged',
+}
+
+export interface Eventing {
+  addHandler(notificationType: string, handler: Handler): void
+}
+
+export type EventContext = {
+  notificationType: string;
+  [key: string]: unknown;
+}
+
+export type Handler = (eventContext?: EventContext) => void
+
+export type EventCallbackMessage = (eventContext: EventContext) => void
 
 export type JsonObject = { [key: string]: JsonValue };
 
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -1,5 +1,50 @@
+import EventEmitter from 'events';
+
 export type PrimitiveValue = null | boolean | string | number;
 
+export interface ProviderEventing {
+  events?: EventEmitter;
+  ready?: boolean;
+}
```
toddbaert, line 8: I'll take the time to document all these types if this looks good to everyone.

src/types.ts
```diff
@@ -1,5 +1,48 @@
+import EventEmitter from 'events';
+
 export type PrimitiveValue = null | boolean | string | number;
 
+export interface EventProvider {
+  readonly events: EventEmitter;
+  readonly ready: boolean;
+}
```
toddbaert, line 8: Providers can optionally implement this (backwards compatible) to support events.

All they must do is mark themselves as ready with the boolean, and call `events.emit(xxx)` to support events.

kinyoklion:
Using EventEmitter can be somewhat annoying if you are working on non-node platforms. I am not sure if [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) is any better in that regard. You can of course, but the more polyfills people need the more annoying things are. Sometimes it is better to just mirror some of that API.

toddbaert:
> Using EventEmitter can be somewhat annoying if you are working on non-node platforms. I am not sure if [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) is any better in that regard. You can of course, but the more polyfills people need the more annoying things are. Sometimes it is better to just mirror some of that API.

Yes, for web, we will probably have a completely separate SDK. To minimize dependencies, in my web SDK PoC, I've used `window.dispatchEvent` (see [here](https://github.com/open-feature/js-sdk/pull/353/files#diff-ed2b939cef6b7455068ab86c443272afeb42f47a0264b82ec98fbd19d774e4d1R53)). I've never really loved the ergonomics of that API, bit it works well-enough and I think reducing the need for polyfills is worth it.

PR #318, opened 2022-11-04
chore: fix publish experimental hash

toddbaert:
@beeme1mr  I'm not sure if this was always the case, or something changed, but there's a minor bug in the experimental publish that caused it to be labeled with the wrong SHA.

The contents are correct, but the SHA would actually be the SHA of the HEAD of main (which is a bit confusing). This fixes that to be the SHA of the PR branch HEAD.

### DIFF
--- a/.github/workflows/publish-experimental.yml
+++ b/.github/workflows/publish-experimental.yml
@@ -32,7 +32,7 @@ jobs:
 
       # create an experimental version at this SHA (vX.X.X-experimental-4742ef14cc0ea07e3569eee56899937452a55a9c)
       - name: Version
-        run: npm version "$(npm run current-version -s)-experimental-$GITHUB_SHA" --no-git-tag-version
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-experimental.yml
```diff
@@ -32,9 +32,9 @@ jobs:
 
       # create an experimental version at this SHA (vX.X.X-experimental-4742ef14cc0ea07e3569eee56899937452a55a9c)
       - name: Version
-        run: npm version "$(npm run current-version -s)-experimental-$GITHUB_SHA" --no-git-tag-version
+        run: npm version "$(npm run current-version -s)-experimental-$(git rev-parse HEAD)" --no-git-tag-version
```
toddbaert, line 35: because this isn't triggered by a PR (but by a comment) GITHUB_SHA actually just defaults to the HEAD ref of main.

beeme1mr (approved):

PR #349, opened 2022-12-05
chore: remove kodiak, use automerge in renovate

toddbaert:
See title.


renovate.json
```diff
@@ -3,5 +3,12 @@
   "extends": [
     "config:base"
   ],
-  "semanticCommits": "enabled"
+  "semanticCommits": "enabled",
+  "packageRules": [
+    {
+      "matchUpdateTypes": ["minor", "patch"],
+      "matchCurrentVersion": "!/^0/",
```
toddbaert, line 10: The matchCurrentVersion setting above is a rule to exclude any dependencies which are pre-1.0.0 because those can make breaking changes at any time according to the SemVer spec.

beeme1mr (approved):

PR #353, opened 2022-12-20
chore: monorepo conversion + experimental web sdk

toddbaert:
This PR transfroms this repo into a monorepo, with server (exsiting), client (new) and shared (new) packages.

There are no tests yet for the client. Publishing _should_ work but is yet untested.

The experimental web SDK will be published as `0.0.1-experimental` and features: 

- native browser events
- "close" functionality (implemented in server as well)
- async context setting
- sync evaluation

Though we want `close` and `events` in the server SDK, they are not implemented in this PR (though it would be very easy to do so and I've left TODO's about it). **There are no functional changes in the server SDK with this PR.**

Here is a [direct link](https://github.com/open-feature/js-sdk/blob/experimental-web-sdk/README.md) to the README.md changes.

**_I'm not expecting reviewers to go through this with a fine-toothed comb. Most of the changes are just establishing the monorepo, and the functional stuff in the new web-sdk is just experimental._**.

Closes: https://github.com/open-feature/js-sdk/issues/367


package.json
```diff
@@ -1,20 +1,23 @@
 {
-  "name": "@openfeature/js-sdk",
-  "version": "1.0.0",
+  "name": "@openfeature/web-sdk",
+  "version": "1.0.99",
   "description": "OpenFeature SDK for JavaScript",
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm/index.js",
-  "types": "./dist/types/index.d.ts",
+  "types": "./dist/types/client/index.d.ts",
```
toddbaert, line 7: if we wanted to actually build this in this repo, we'd need to maintain 2 package.json files.

.github/workflows/release-please.yml
```diff
@@ -42,14 +42,14 @@ jobs:
...
-        with:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          BRANCH: gh-pages
-          FOLDER: typedoc
+      # - name: Deploy Documentation 🚀
+      #   uses: JamesIves/github-pages-deploy-action@releases/v3
+      #   with:
+      #     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      #     BRANCH: gh-pages
+      #     FOLDER: typedoc
```
toddbaert, line 55: @beeme1mr this is the only thing I haven't toyed with at all.

.github/workflows/release-please.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - name: Publish to NPM
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
-        run: npm publish --access public
+        run: npm run publish-all
```
toddbaert, line 45: This script attempts to publish the client and server modules by calling a script in each of them. If the artifact already exists, it's a no-op, similar to what we do in many other contribs.

.release-please-manifest.json
```diff
@@ -1,3 +1,5 @@
 {
-  ".": "1.1.0"
+  "packages/server": "1.1.0",
+  "packages/client": "0.0.1-experimental",
+  "packages/shared": "0.0.1"
```
toddbaert, line 4: The `shared` module is never published, but it's listed here so release-please can do it's magic and bump dependent packages when there's changes here.

package.json
```diff
@@ -57,14 +46,24 @@
...
     "ts-node": "^10.8.2",
     "typedoc": "^0.23.16",
     "typescript": "^4.7.4"
-  }
+  },
+  "workspaces": [
+    "packages/server",
+    "packages/client",
+    "packages/shared"
+  ]
```
toddbaert, line 68: new npm workspaces field.

packages/client/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "browser",
+    "web"
+  ],
+  "author": "",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/open-feature/js-sdk/issues"
+  },
+  "homepage": "https://github.com/open-feature/js-sdk#readme",
+  "devDependencies": {
```
toddbaert, line 48: Both server and client packages have this dev dep, this helps release please know that the modules are impacted by changes in this `shared` module.

packages/server/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "node"
+  ],
+  "author": "",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/open-feature/js-sdk/issues"
+  },
+  "homepage": "https://github.com/open-feature/js-sdk#readme",
+  "devDependencies": {
+    "@openfeature/shared": "file:../shared"
```
toddbaert, line 47: https://github.com/open-feature/js-sdk/pull/353/files#r1128152215

README.md
```diff
@@ -35,79 +39,36 @@ or
...
-// create a client
-const client = OpenFeature.getClient('my-app');
-
-// get a bool value
-const boolValue = await client.getBooleanValue('boolFlag', false);
+See [README.md](./packages/server/README.md)
 
-// get a string value
-const stringValue = await client.getStringValue('stringFlag', 'default');
+## Client
```
beeme1mr, line 44: What do you think about removing this section until we've verified that everything was published propertly?

beeme1mr: I think we should make it clear that this section is experimental and not ready for production use.

package.json
```diff
@@ -1,31 +1,27 @@
 {
-  "name": "@openfeature/js-sdk",
-  "version": "1.1.0",
+  "name": "@openfeature/js",
+  "version": "0.0.0",
```
beeme1mr, line 3: Is the version used?

toddbaert: I don't think this version is used for anything, no. The root project is not published.

README.md
toddbaert: I'm fine with that.

package.json
beeme1mr: It may be worth explicitly setting this package.json to private.
https://docs.npmjs.com/cli/v9/configuring-npm/package-json#private

packages/client/src/types.ts
```diff
@@ -0,0 +1,373 @@
...
+   * to tap into various stages of the flag evaluation lifecycle. These hooks can
+   * be used to perform side effects and mutate the context for purposes of the
+   * provider. Provider hooks are not configured or controlled by the application author.
+   */
+  readonly hooks?: Hook[];
+
+  // client vs global context?
+  onContextChange?(oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void>
+
+  // TODO: move to common Provider type when we want close in server
```
beeme1mr, line 39: Are you going to address this TODO in this PR?

packages/server/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+or
+
+```shell
+yarn add @openfeature/js-sdk
+```
+
+## Usage
+
+To configure the SDK you'll need to add a provider to the `OpenFeature` global signleton. From there, you can generate a `client` which is usable by your code. While you'll likely want a provider for your specific backend, we've provided a `NoopProvider`, which simply returns the default value.
```
beeme1mr, line 19: > 
```suggestion
To configure the SDK you'll need to add a provider to the `OpenFeature` global singleton. From there, you can generate a `client` which is usable by your code. While you'll likely want a provider for your specific backend, we've provided a `NoopProvider`, which simply returns the default value.
```

beeme1mr (approved):
The current server works as expected. This will allow to rapidly iterate on the web sdk.

package.json
toddbaert: You might be right about that, I think it makes sense.

toddbaert: Done.

README.md
toddbaert: I've removed this for now.

packages/client/src/types.ts
```diff
@@ -0,0 +1,373 @@
...
+  onClose?(): Promise<void>;
+
+  /**
+   * Resolve a boolean flag and its evaluation details.
+   */
+  resolveBooleanEvaluation(
+    flagKey: string,
+    defaultValue: boolean,
+    context: EvaluationContext,
```
moredip, line 48: I'm suprised that we're passing context here. I thought that the static context paradigm would mean you can't pass context at the point of evaluation, in exchange for it being a synchronous call.

PR #384, opened 2023-03-23
feat: add caching for pipelines

### LINKED ISSUE #310: GitHub Action Cache
Configure appropriate GitHub actions to use a cache.

RealAnna:
## This PR

- adds npm caching in all pipelines, with cache setup node takes 5s against 12s of before
image

### Related Issues

Fixes #310

### Notes

since 14 is now too old, we decided that 16 was the best version to avoid breaking changes

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -44,6 +44,9 @@
     "url": "https://github.com/open-feature/js-sdk/issues"
   },
   "homepage": "https://github.com/open-feature/js-sdk#readme",
+  "engines": {
+    "node": ">=16"
+  },
   "devDependencies": {
     "@openfeature/shared": "0.0.2"
   }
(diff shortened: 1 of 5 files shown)


packages/server/package.json
```diff
@@ -44,6 +44,9 @@
     "url": "https://github.com/open-feature/js-sdk/issues"
   },
   "homepage": "https://github.com/open-feature/js-sdk#readme",
+  "engines": {
+    "node": ">=16"
+  },
```
toddbaert, line 49: Just a note for others. Previously the server SDK was requiring node 14, but that's very dead. We are upgrading to 16 here but not considering it a breaking change since the version is so old (even 16 is partially EOL).

https://endoflife.date/nodejs

PR #389, opened 2023-03-30
feat!: use bundled event emitter type

toddbaert:
:warning: This change only impacts the experimental web-sdk.

This includes an aliased, bundled version of `EventEmitter` so we have full control over that API/implementation, and consumers don't need the `events` dep for browsers.

We currently bundle [events](https://www.npmjs.com/package/events) (a browser polyfill for the node events API) into this package, but a provider author who wants to implement events in their provider would need to install that package as well. This removes the need for that. Interested in others' thoughts.

### DIFF
--- a/packages/client/src/client.ts
+++ b/packages/client/src/client.ts
@@ -15,7 +15,7 @@ import {
   ResolutionDetails,
   StandardResolutionReasons
 } from '@openfeature/shared';
-import { EventEmitter } from 'events';
+import { OpenFeatureEventEmitter } from '.';
 import { OpenFeature } from './open-feature';
 import { Client, FlagEvaluationOptions, Hook, Provider } from './types';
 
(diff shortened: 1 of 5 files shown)


kinyoklion:
I think this is a step in the right direction. I know you are only intending to change the client at the moment, but I think this would generally be important for server as well, though it may also become less so over time. Edge environments have their own non-node version of a number of libraries, so it would be convenient to be able to swap. I think Cloudflare added EventEmitter around a week ago, for instance.

beeme1mr:
I'm wondering if we should define our own interface that the `events` library just happens to conform with. I'm not sure if it's worth the effort, but it would be a way to support custom event emitters and ensure that `events` doesn't change in a breaking way.

toddbaert:
> I'm wondering if we should define our own interface that the `events` library just happens to conform with. I'm not sure if it's worth the effort, but it would be a way to support custom event emitters and ensure that `events` doesn't change in a breaking way.

I think that exposing (and aliasing) our own export allow us to defer this decision. Of course, the underlying events lib could change underneath us, but between reasonable unit tests (yet to be written) and a locked version of that dep, I think we can call this good enough for now.

I dont think it would be a huge effort though, so we could do what you say if you feel like it's worth it.

beeme1mr (approved):

PR #394, opened 2023-04-03
fix: fix docs

toddbaert:
Fix doc publishing

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -41,5 +41,8 @@ jobs:
       - name: Test
         run: npm run test
 
+      - name: Docs
+        run: npm run docs
+
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -47,5 +47,9 @@
   "homepage": "https://github.com/open-feature/js-sdk#readme",
   "devDependencies": {
     "@openfeature/shared": "0.0.2"
+  },
+  "typedoc": {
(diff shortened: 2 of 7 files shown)


packages/client/package.json
```diff
@@ -47,5 +47,9 @@
   "homepage": "https://github.com/open-feature/js-sdk#readme",
   "devDependencies": {
     "@openfeature/shared": "0.0.2"
+  },
+  "typedoc": {
+    "displayName": "OpenFeature Web SDK",
+    "entryPoint": "./src/index.ts"
```
toddbaert, line 53: This overrides the default entry point, which comes automatically from the package.json `main` field, and causes issues due to requiring source maps to get back to the TS..

packages/client/src/types.ts
```diff
@@ -16,7 +16,8 @@ import {
   ProviderMetadata,
   ResolutionDetails,
 } from '@openfeature/shared';
-import { OpenFeatureEventEmitter } from '.';
+import { EventEmitter as OpenFeatureEventEmitter } from 'events';
+export { OpenFeatureEventEmitter };
```
toddbaert, line 20: This fixes a dependency cycle.

.github/workflows/pr-checks.yaml
```diff
@@ -41,5 +41,8 @@ jobs:
       - name: Test
         run: npm run test
 
+      - name: Docs
+        run: npm run docs
+
```
toddbaert, line 46: Added docs to the CI, which will prevent regressions here again.

beeme1mr (approved):

PR #395, opened 2023-04-03
fix: import cycle

toddbaert:
Fix import cycle caused by previous export/import from barrel file.


packages/client/src/open-feature.ts
```diff
@@ -1,8 +1,16 @@
-import { ApiEvents, EvaluationContext, FlagValue, Logger, OpenFeatureCommonAPI, ProviderEvents, ProviderMetadata, SafeLogger } from '@openfeature/shared';
-import { OpenFeatureEventEmitter } from '.';
```
toddbaert, line 2: These files were both importing the barrel file, and exporting from it, which obviously is a cycle.

beeme1mr (approved):

PR #400, opened 2023-04-05
test: add client SDK integration tests

### LINKED ISSUE #376: [testing] create integration tests for web-sdk
We should create a cucumber suite for the web-sdk (similar to that in the js-sdk) that runs in a simulated browser environment. We may want to wait until the spec PR is merged or more stable, but we should have some tests sooner than later.

mowies:
### This PR
- implements all integration tests from the test harness for the client package

Fixes #376


jest.config.ts
```diff
@@ -124,11 +124,15 @@ export default {
     },
     {
       displayName: 'client-e2e',
-      testEnvironment: 'jsdom',
+      testEnvironment: 'node',
```
toddbaert, line 127: A bit annoying/surprising, but these tests actually work in node with the latest changes, and DON'T work in jsdom. This is because the `jsdom` test platform is actually missing some of the newest web-standards which recent node versions actually have (`fetch`, `TextEncoder`, etc).

jest.config.ts
```diff
@@ -124,11 +124,15 @@ export default {
...
       preset: 'ts-jest',
       testMatch: ['<rootDir>/packages/client/integration/**/*.spec.ts'],
       modulePathIgnorePatterns: ['.*/node-modules/'],
       setupFiles: ['<rootDir>/packages/client/integration/step-definitions/setup.ts'],
+      moduleNameMapper: {
+        '^uuid$': require.resolve('uuid'),
```
toddbaert, line 133: `uuid` doesn't ship a browser CJS module, unfortunately, so this is a workaround for that broken inport - it rewrites it to the node import.

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "test": "jest --selectProjects=server --selectProjects=shared --verbose",
     "integration-server": "git submodule update --init --recursive && shx cp test-harness/features/evaluation.feature packages/server/integration/features && jest --selectProjects=server-e2e --verbose",
-    "integration-client": "git submodule update --init --recursive && shx cp test-harness/features/evaluation.feature packages/client/integration/features && jest --selectProjects=client-e2e --verbose",
+    "integration-client": "jest --selectProjects=client-e2e --verbose",
```
toddbaert, line 9: I've temporarily removed the module pull and copy so you can add tests 1 by 1. Currently only the first test is implemented.

packages/client/integration/step-definitions/evaluation.spec.ts
```diff
@@ -32,320 +43,12 @@ defineFeature(feature, (test) => {
       /^a boolean flag with key "(.*)" is evaluated with default value "(.*)"$/,
       async (key: string, defaultValue: string) => {
         flagKey = key;
-        value = await client.getBooleanValue(flagKey, defaultValue === 'true');
+        value = await client!.getBooleanValue(flagKey, defaultValue === 'true');
```
toddbaert: This test passes! :D

packages/client/integration/step-definitions/evaluation.spec.ts
```diff
@@ -22,6 +23,16 @@ const givenAnOpenfeatureClientIsRegisteredWithCacheDisabled = (
...
 
 defineFeature(feature, (test) => {
+  beforeAll((done) => {
+    client.addHandler(ProviderEvents.Ready, () => {
+      done();
+    })
+  });
+
+  afterAll(() => {
+    OpenFeature.close();
```
toddbaert, line 33: You may see an error like: 

```
Jest did not exit one second after the test run has completed.

'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

This is actually a bug in the flagd web provider which I will fix (and add you to). We aren't properly shutting down the event stream. Once this fix is published that will stop.

.github/workflows/pr-checks.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: 18
```
toddbaert, line 26: see [here](https://github.com/open-feature/js-sdk/pull/400/commits/14b5d0036f13c666228851e5cfd06ef3cef761d3#r1163207162)

toddbaert:
> Build is failing because of this circular dependency issue with the flagd web provider and NPM, other than that, everything should be ready @toddbaert

@mowies I pushed a couple changes: 

- updated flagd-web to 0.3.4, which adds a proper shutdown which stops jest from hanging see [here](https://github.com/open-feature/js-sdk-contrib/pull/325)).
- installed the flagd and flagd-web provider with `npm install --save-dev --legacy-peer-deps`. `--legacy-peer-deps` is an NPM option that ignores peer deps, which helps us workaround this npm issue (I think it's a bug, TBH). This only needs to be used to generate a sane `package-lock.json` when either of these packages is updated and not when `npm install` or `npm ci` is run, so the build works just fine in CI. This is also only a dev thing, so there's no impact on consumers.

Everything seems to be working now! Thanks for all your efforts here!

.github/workflows/pr-checks.yaml
```diff
@@ -13,17 +13,11 @@ jobs:
...
-        image: ghcr.io/open-feature/flagd-testbed:latest
-        ports:
-          - 8013:8013
-
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 16
```
toddbaert: revert back to 16

toddbaert (approved):

PR #410, opened 2023-04-18
test: add events tests for web-sdk

### LINKED ISSUE #407: [web-sdk] Add tests for events
The web-sdk implements events as per this [spec PR]([specification/sections/05-events.md](https://github.com/open-feature/spec/pull/171) (see specification/sections/05-events.md). 
We've implemented the tests equivalent to the server SDK [here](https://github.com/open-feature/js-sdk/commit/39f64742cf308bb83a4299b069d61f6ebfac9a46), but because the events features don't exist in the server yet, we need to implement them in the web SDK fresh.

Primarily, we want to cover the event logic.

DoD:

- test(s) that assert when the provider is initialized (the initialize promise resolves) the READY event fires on a registered client
- test(s) that assert when the provider initialization fails (the initialize promise rejects) the ERROR event fires on a registered client
- test(s) that assert when a provider fires the CONFIGURATION_CHANGED event, the event fires on a registered client
- other reasonable assertions based on the draft spec PR.

RealAnna:
## This PR

- added tests based on [spec PR](https://github.com/open-feature/spec/pull/171)

### Related Issues

Fixes #407

### DIFF
--- a/packages/client/test/events.spec.ts
+++ b/packages/client/test/events.spec.ts
@@ -1,54 +1,10 @@
-import {
-  OpenFeature,
-  ApiEvents,
-  ProviderEvents,
-  Provider,
-} from '../src';
-import EventEmitter from 'events';
+import { addListener, nextTick } from 'process';
+import { ApiEvents, OpenFeature, Provider, ProviderEvents, OpenFeatureEventEmitter } from '../src';
+
 
 const ERROR_REASON = 'error';
 const ERROR_CODE = 'MOCKED_ERROR';
 
-const MOCK_EMITTER: EventEmitter = {
-  emit: jest.fn( () => {
-    return true;
-  }),
-  initialize: jest.fn( () => {
-    return MOCK_EMITTER;
-  })
-}as unknown as EventEmitter;
-
-const MOCK_ERROR_EMITTER: EventEmitter = {
-  emit: jest.fn( (event) => {
-    if (event == ProviderEvents.Ready) {
-      throw Error('Event error!');
-    }
-    return true;
-  }),
-  initialize: jest.fn( () => {
-    return MOCK_EMITTER;
-  })
-}as unknown as EventEmitter;
-
-const MOCK_PROVIDER: Provider = {
-  metadata: {
-    name: 'mock-events-success',
-  },
-  initialize: jest.fn(() =>  {
-    return Promise.resolve(undefined);
-  }),
-  events: MOCK_EMITTER,
-} as unknown as Provider;
-
-const ERROR_PROVIDER: Provider = {
-  metadata: {
-    name: 'mock-events-failure',
-  },
-  initialize: jest.fn(() =>  {
-    return Promise.resolve(undefined);
-  }),
-  events: MOCK_ERROR_EMITTER,
-} as unknown as Provider;
 
 describe('Events', () => {
   // set timeouts short for this suite.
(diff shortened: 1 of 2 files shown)


packages/client/test/events.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('Requirement 5.1.1', () => {
+    it('The provider **MAY** define a mechanism for signalling the occurrence of an event`PROVIDER_READY`',
+       async () => {
+
+         await OpenFeature.setProvider(MOCK_PROVIDER);
+         expect(OpenFeature['_provider']).toBe(MOCK_PROVIDER);
```
toddbaert, line 66: I don't think we need to test this private prop. If you want, you can test the metadata: https://github.com/open-feature/js-sdk/blob/0c2befdbdbfbf46dede010e00f0efcd8e0f0f160/packages/client/src/open-feature.ts#L57

packages/client/test/events.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+describe('Events', () => {
+  // set timeouts short for this suite.
+  jest.setTimeout(1000);
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('Requirement 5.1.1', () => {
+    it('The provider **MAY** define a mechanism for signalling the occurrence of an event`PROVIDER_READY`',
```
toddbaert, line 62: Nitpick: `it('The provider **MAY**...` reads weird. I recommend something like ``it('defines a mechanism...` so the test output is more readable, we don't need to put the spec text in each `it()`, IMHO. Doesn't really matter though.

packages/client/test/events.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+              return this;
+            })
+          } as unknown as EventEmitter;
+
+
+          await OpenFeature.setProvider(errProvider);
+          expect(OpenFeature['_provider']).toBe(errProvider);
+
+          expect(errProvider.initialize).toHaveBeenCalled();
+          //expect(apiMockEmitter.emit).toHaveBeenCalledWith(ProviderEvents.Error); TODO figure out how to check client error event, apiEvents uses only CONFIGURATION_CHANGED
```
toddbaert, line 113: You can create a client, and instead of making this function async, you can use a `done` param. You can put that in the `addHandler` callback (after any needed assertions) and that will end the test.

Here is an example of a test that uses `done()`: https://github.com/open-feature/js-sdk/blob/39f64742cf308bb83a4299b069d61f6ebfac9a46/packages/client/test/hooks.spec.ts#L53-L73. Be sure to add a try/catch with a `done(err)`.

packages/client/test/events.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+
+
+describe('Events', () => {
+  // set timeouts short for this suite.
+  jest.setTimeout(1000);
+  let API = new (OpenFeatureAPI as any)(); 
+
+  afterEach(() => {
+    jest.clearAllMocks();
+    API = new (OpenFeatureAPI as any)(); //TODO this hacky thing should be removed when https://github.com/open-feature/spec/pull/183/files is implemented
```
toddbaert, line 44:
```suggestion
    API = new (OpenFeatureAPI as any)(); //TODO this hacky thing should be removed with provider mapping: https://github.com/open-feature/js-sdk/issues/421
```

toddbaert: I made an issue for this.

packages/client/test/events.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  });
+
+  describe('Requirement 5.2.1, 5.2.3, 5.2.4, 5.2.5 ', () => {
+    it('The `client` MUST provide an `addHandler` function for attaching callbacks to `provider events`, which accepts event type(s) and a `event handler function`.', (done) => {
+      const client = API.getClient();
+      client.addHandler(ProviderEvents.Ready, () => {
+        done();
+      });
+      API.setProvider(MOCK_PROVIDER);
```
toddbaert, line 116: if this is a dupe (seems to be) you can remove it. It's implicitly tested below.

toddbaert (approved):
Thanks! LGTM!

The current implementation of events is a but rough, and the spec is still in progress, but this is a great start for so we can have some confidence as things are implemented.

toddbaert (changes requested):
I think I found one issue [here](https://github.com/RealAnna/js-sdk/pull/2), which is a bit of a misunderstanding.

toddbaert (approved):

PR #411, opened 2023-04-18
test: add tests to provider init and shutdown

### LINKED ISSUE #406: [web-sdk] Add tests for initialization/shutdown
The web-sdk implements initialization and shutdown as per this [spec PR](https://github.com/open-feature/spec/pull/179) and this [OFEP](https://github.com/open-feature/ofep/pull/46).

We've implemented the tests equivalent to the server SDK [here](https://github.com/open-feature/js-sdk/commit/39f64742cf308bb83a4299b069d61f6ebfac9a46), but because the init/shutdown features don't exist in the server yet, we need to implement them in the web SDK.

Primarily, we want to cover the init/shutdown logic in the [`setProvider` method](https://github.com/open-feature/js-sdk/blob/main/packages/client/src/open-feature.ts#L86-L112), and in the [`close` method](https://github.com/open-feature/js-sdk/blob/main/packages/client/src/open-feature.ts#L114-L116).

DoD:

- test(s) that assert shutdown is called on the old provider when a new one is set
- test(s) that assert initialize is called on the new provider
- test(s) that assert shutdown is called on the provider when the `close()` method on the API is called
- tests are numbered/named in accordance with [spec PR](https://github.com/open-feature/spec/pull/179)

RealAnna:
## This PR

- adds checks for calls to the init/shutdown methods of a Provider

### Related Issues

Fixes #406

### DIFF
--- a/packages/client/test/open-feature.spec.ts
+++ b/packages/client/test/open-feature.spec.ts
@@ -1,6 +1,18 @@
-import { OpenFeatureClient } from '../src/client';
-import { OpenFeature, OpenFeatureAPI } from '../src/open-feature';
-import { Provider,  } from '../src/types';
+import {OpenFeatureClient} from '../src/client';
+import {OpenFeature, OpenFeatureAPI} from '../src/open-feature';
+import {Provider,} from '../src/types';
+
+const MOCK_PROVIDER: Provider = {
+  metadata: {
+    name: 'mock-events-success',
+  },
+  initialize: jest.fn(() => {
+    return Promise.resolve('started');
+  }),
+  onClose: jest.fn(() => {
+    return Promise.resolve('closed');
+  }),
+} as unknown as Provider;
 
 describe('OpenFeature', () => {
   afterEach(() => {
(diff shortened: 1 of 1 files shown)


packages/client/test/open-feature.spec.ts
```diff
@@ -15,10 +27,28 @@ describe('OpenFeature', () => {
...
-      const fakeProvider = { metadata: 'test' } as unknown as Provider;
+      const fakeProvider = {metadata: 'test'} as unknown as Provider;
       OpenFeature.setProvider(fakeProvider);
       expect(OpenFeature.providerMetadata === fakeProvider.metadata).toBeTruthy();
     });
+
+    describe('Requirement 1.1.2.2', () => {
+      it('MUST invoke the `initialize` function on the newly registered provider before using it to resolve flag values', () => {
+        OpenFeature.setProvider(MOCK_PROVIDER);
+        expect(OpenFeature['_provider']).toBe(MOCK_PROVIDER);
```
toddbaert, line 38: Make an assertion about the public metadata field instead, I think.

packages/client/test/open-feature.spec.ts
```diff
@@ -51,12 +81,22 @@ describe('OpenFeature', () => {
...
+
   });
 
+  describe('Requirement 1.6.1', () => {
+    it('MUST define a `shutdown` function, which, when called, must call the respective `shutdown` function on the active provider', () => {
+      OpenFeature.setProvider(MOCK_PROVIDER);
+      expect(OpenFeature.providerMetadata.name).toBe('mock-events-success');
+      OpenFeature.close().then(
+        () => {
+          expect(MOCK_PROVIDER.onClose).toHaveBeenCalled();
```
toddbaert, line 98: Since this test isn't async, and also there's no `done()` being used, this test will end before the expectation runs.

You can either make it async and await the promise, or put a called to `done()` inside the `then`.

packages/client/test/open-feature.spec.ts
```diff
@@ -15,10 +27,28 @@ describe('OpenFeature', () => {
...
+      });
+    });
+    describe('Requirement 1.1.2.3', () => {
+      it('MUST invoke the `shutdown` function on the previously registered provider once it\'s no longer being used to resolve flag values.', () => {
+        const fakeProvider = {metadata: 'test'} as unknown as Provider;
+        OpenFeature.setProvider(MOCK_PROVIDER);
+        OpenFeature.setProvider(fakeProvider);
+        expect(MOCK_PROVIDER.onClose).toHaveBeenCalled();
+      });
+    });
```
toddbaert, line 49: These look good!

toddbaert (approved):

toddbaert:
Thanks @RealAnna

PR #413, opened 2023-05-01
fix: remove events exports from server

toddbaert:
Moves some events-related exports (unused by server API) from `/shared` to `/client`. We will eventually move some of these back, but for now I don't want these types to be seen by the server SDK (even if they are unused) since they are experimental.


beeme1mr (approved):
Good catch!

PR #422, opened 2023-05-05
chore: refactor events

toddbaert:
No functional changes. Refactor to how events are done, specifically with how they are maintained even across provider changes. This code is easier to understand IMHO, and also requires a few less structures/types.

Thanks to @RealAnna 's recent tests, we can do these changes with confidence :partying_face:

### DIFF
--- a/packages/client/src/open-feature.ts
+++ b/packages/client/src/open-feature.ts
@@ -8,7 +8,7 @@ import {
 } from '@openfeature/shared';
 import { OpenFeatureClient } from './client';
 import { NOOP_PROVIDER } from './no-op-provider';
-import { ApiEvents, Client, Hook, OpenFeatureEventEmitter, Provider, ProviderEvents } from './types';
+import { Client, Hook, OpenFeatureEventEmitter, Provider, ProviderEvents } from './types';
 
 // use a symbol as a key for the global singleton
 const GLOBAL_OPENFEATURE_API_KEY = Symbol.for('@openfeature/js.api');
@@ -19,10 +19,10 @@ type OpenFeatureGlobal = {
 const _globalThis = globalThis as OpenFeatureGlobal;
 
 export class OpenFeatureAPI extends OpenFeatureCommonAPI {
-  private _apiEvents = new OpenFeatureEventEmitter();
   private _providerReady = false;
   protected _hooks: Hook[] = [];
   protected _provider: Provider = NOOP_PROVIDER;
+  private readonly _events = new OpenFeatureEventEmitter();
 
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   private constructor() {
@@ -87,9 +87,8 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI {
       this._provider = provider;
       this._providerReady = false;
 
-      if (!this._provider.events) {
-        this._provider.events = new OpenFeatureEventEmitter();
-      }
+      this.removeListeners(oldProvider);
+      this.attachListeners(this._provider);
 
       if (typeof this._provider?.initialize === 'function') {
         this._provider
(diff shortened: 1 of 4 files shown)


packages/client/src/types.ts
```diff
@@ -45,10 +45,6 @@ export interface EventData {
   changeMetadata?: { [key: string]: boolean | string } // similar to flag metadata
 }
 
-export enum ApiEvents {
-  ProviderChanged = 'providerChanged',
-}
-
```
toddbaert, line 51: These events were an internal implementation detail.

packages/client/src/open-feature.ts
```diff
@@ -87,9 +87,8 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI {
       this._provider = provider;
       this._providerReady = false;
```
beeme1mr, line 88: Should we also emit an event here?

toddbaert: The latest actually specifies that the provider shouldn't be switched until it's initialized, but because of the rapid changes I haven't implemented any actual changes to be in sync with the latest events proposal. I'd rather keep this as a pure refactor and not change any behavior.

toddbaert:
@RealAnna please take a look since you've spent time in this area recently, if you have a moment.

packages/client/src/open-feature.ts
```diff
@@ -87,9 +87,8 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI {
...
 
-      if (!this._provider.events) {
-        this._provider.events = new OpenFeatureEventEmitter();
-      }
+      this.removeListeners(oldProvider);
+      this.attachListeners(this._provider);
```
RealAnna, line 91: thanks to these now the event tests pass also without reinitializing the API, the only exception is ```
Requirement 5.1.1', () => {
    it('The provider defines a mechanism for signaling the occurrence of an event ''PROVIDER_READY`'
```
The ready is not handled if we register it after... 🦆 do you think it is worth investigating why?

Comment (toddbaert) [reply packages/client/src/open-feature.ts:91]: @RealAnna OK so I found a bug with my implementation. I was still emitting the events directly from the provider events object. That's a problem, because some providers wont implement events, but as the recent spec change dictates, READY/ERROR events on the client must STILL must fire in the case on the initialization. For more details on that, see [here](https://github.com/open-feature/spec/pull/182/files#diff-daf30d840f7e7b1752899ac2c4b268af214e6c8e7058549c8d4218abd2e86ba3R104-R129). I added tests for this case.

I think what you were pointing out was an issue because we were adding handlers directly to the provider events. I know we chose to do this initially because it scoped the tests more (unit style instead of integration style) but I think this isn't ideal. The spec doesn't actually say that it needs to be possible to subscribe to provider events - the provider just needs to be able to emit them. The important thing to test is that handlers run on the `client` when those events are emitted. I have updated the tests in question to work that way (some already did work that way).

See my changes [here](https://github.com/open-feature/js-sdk/pull/422/commits/1b447a853a4233455b7776eea2fe19cf1f121cf2).

File: packages/client/test/events.spec.ts [hunk 4]
```diff
@@ -44,34 +65,67 @@ describe('Events', () => {
...
+        client.addHandler(ProviderEvents.Error, () => {
+          try {
+            expect(API.providerMetadata.name).toBe(MOCK_ERROR_NO_EVENT_PROVIDER.metadata.name);
+            expect(MOCK_ERROR_NO_EVENT_PROVIDER.initialize).toHaveBeenCalled();
+            done();
+          } catch (err) {
+            done(err);
+          }
+        });
+        API.setProvider(MOCK_ERROR_NO_EVENT_PROVIDER);
```
toddbaert, line 126: @RealAnna These are the new tests I added. Similar to what was here before, except the mock provider has no `events` object. They failed before my last change.

RealAnna (approved):

packages/client/src/open-feature.ts
RealAnna: @toddbaert I see your point, this makes a lot more sense to me now!

RealAnna (approved):

PR #426, opened 2023-05-20
feat: add support for flag metadata

### LINKED ISSUE #425: implement flag metadata: open-feature/spec@74c373e
https://github.com/open-feature/spec/pull/169

beeme1mr:
## This PR

- add support for flag metadata

### Related Issues

Fixes #425

### Notes

This feature is part of the recently released v0.6.0 of the OpenFeature spec.

### How to test

### DIFF
--- a/packages/server/src/client.ts
+++ b/packages/server/src/client.ts
@@ -212,6 +212,7 @@ export class OpenFeatureClient implements Client {
 
       const evaluationDetails = {
         ...resolution,
+        flagMetadata: Object.freeze(resolution.flagMetadata ?? {}),
         flagKey,
       };
 
@@ -229,6 +230,7 @@ export class OpenFeatureClient implements Client {
         errorMessage,
         value: defaultValue,
         reason: StandardResolutionReasons.ERROR,
+        flagMetadata: Object.freeze({}),
         flagKey,
       };
     } finally {
--- a/packages/server/test/client.spec.ts
+++ b/packages/server/test/client.spec.ts
@@ -402,6 +402,39 @@ describe('OpenFeatureClient', () => {
     });
   });
 
+  describe('Requirement 1.4.13, Requirement 1.4.14', () => {
+    it('should return immutable `flag metadata` as defined by the provider', async () => {
+      const flagMetadata = {
+        url: 'https://test.dev',
+        version: '1',
+      };
+
+      const flagMetadataProvider = {
+        name: 'flag-metadata',
+        resolveBooleanEvaluation: jest.fn((): Promise<ResolutionDetails<boolean>> => {
+          return Promise.resolve({
+            value: true,
+            flagMetadata,
+          });
+        }),
+      } as unknown as Provider;
+
+      OpenFeature.setProvider(flagMetadataProvider);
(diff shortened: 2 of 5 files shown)


packages/shared/src/types.ts
```diff
@@ -125,16 +125,25 @@ export enum ErrorCode {
 
 export type ResolutionReason = keyof typeof StandardResolutionReasons | (string & Record<never, never>);
 
+/**
+ * A structure which supports definition of arbitrary properties, with keys of type string, and values of type boolean, string, or number.
+ * 
+ * This structure is populated by a provider for use by an Application Author (via the Evaluation API) or an Application Integrator (via hooks).
+ */
+export type FlagMetadata = Record<string, string | number | boolean>;
```
beeme1mr, line 133: This complies with [requirement 2.2.9](https://openfeature.dev/specification/sections/providers#requirement-229) but doesn't have a test. Do you think one is needed? I couldn't find the other 2.x tests.

toddbaert: Requirements that are structural (types) or prescriptions to providers don't really require tests here IMO.

packages/server/src/client.ts
```diff
@@ -229,6 +230,7 @@ export class OpenFeatureClient implements Client {
         errorMessage,
         value: defaultValue,
         reason: StandardResolutionReasons.ERROR,
+        flagMetadata: Object.freeze({}),
```
toddbaert, line 233: I'm OK with this... I think it's better than keeping it undefined. It's a bit weird that this field is here and "always empty" in error cases. I would be nice if we could mark it undefined in the type if the `reason` is error.

On the other hand, that typing wouldn't help a plain JS user, and they could get into trouble.

packages/server/test/client.spec.ts
```diff
@@ -402,6 +402,39 @@ describe('OpenFeatureClient', () => {
     });
   });
 
+  describe('Requirement 1.4.13, Requirement 1.4.14', () => {
```
toddbaert, line 405: These seem to cover it!

toddbaert (approved):

packages/server/src/client.ts
lukas-reining: Yeah thats what I thought.
Question to me was if read-only is good enough or if we need to have the promise by freeze.

beeme1mr: I don't have a strong opinion either way, honesty. I'll see if I can create a discriminated union type that doesn't include flag metadata when the reason is error.

toddbaert: > I don't have a strong opinion either way, honesty. I'll see if I can create a discriminated union type that doesn't include flag metadata when the reason is error.

If you can, great. I dont think there's much risk here either way TBH since this is just empty in the error case. I think it's fine to keep it as is.

lukas-reining (approved):

packages/server/src/client.ts
beeme1mr: I couldn't find a way to do this without massively complicating the type. For the sake of maintainability, I'll leave it as is.

PR #427, opened 2023-05-20
refactor: remove unnecessary promises from client tests

beeme1mr:
## This PR

- remove unnecessary promises from client tests

### Notes

The client APIs are synchronous, but the tests were awaiting responses.


toddbaert (approved):
Thanks for noticing!

PR #429, opened 2023-05-23
feat: add named client support

### LINKED ISSUE #428: implement named client support: open-feature/spec@4cf8229
Add named client support to the JS SDK based on the recent update to the spec: https://github.com/open-feature/spec/commit/4cf8229d23a88feeaadf8b066e5d3662b94ff262

This feature has already been implemented in Java. It can be used as a reference for the JS implementation.
https://github.com/open-feature/java-sdk/pull/388

lukas-reining:
## This PR
Implements named client support. #428 

Had to also add the provider status that is defined in the provider lifecycle and used for events.

### Related Issues

Fixes #428 

### Notes
This feature is part of the recently released v0.6.0 of the OpenFeature spec.

### Follow-up Tasks
I would after this continue with the events and init/shutdown to round everything up.

### How to test

### DIFF
--- a/packages/client/e2e/step-definitions/evaluation.spec.ts
+++ b/packages/client/e2e/step-definitions/evaluation.spec.ts
@@ -7,7 +7,7 @@ import {
   ResolutionDetails,
   StandardResolutionReasons,
 } from '@openfeature/shared';
-import { OpenFeature, ProviderEvents } from '../../';
+import { OpenFeature, ProviderEvents } from '../../src';
 // load the feature file.
 const feature = loadFeature('packages/client/e2e/features/evaluation.feature');
 
@@ -24,7 +24,9 @@ const givenAnOpenfeatureClientIsRegisteredWithCacheDisabled = (
 defineFeature(feature, (test) => {
   beforeAll((done) => {
     client.addHandler(ProviderEvents.Ready, () => {
-      done();
+      setTimeout(() => {
+        done(); // TODO Fix by setting readiness correctly in provider
+      }, 2000);
     });
   });
 
--- a/packages/client/e2e/step-definitions/evaluation.spec.ts
+++ b/packages/client/e2e/step-definitions/evaluation.spec.ts
@@ -25,7 +25,7 @@ defineFeature(feature, (test) => {
   beforeAll((done) => {
     client.addHandler(ProviderEvents.Ready, () => {
       setTimeout(() => {
-        done(); // TODO Fix by setting readiness correctly in provider
+        done(); // TODO remove this once flagd provider properly implements readiness (for now, we add a 2s wait).
       }, 2000);
     });
   });
--- a/packages/client/src/client.ts
+++ b/packages/client/src/client.ts
@@ -1,19 +1,29 @@
-import { EvaluationContext, OpenFeatureError } from '@openfeature/shared';
-import { SafeLogger } from '@openfeature/shared';
 import {
   ClientMetadata,
   ErrorCode,
+  EvaluationContext,
   EvaluationDetails,
   FlagValue,
   FlagValueType,
   HookContext,
   JsonValue,
   Logger,
+  OpenFeatureError,
+  ProviderStatus,
   ResolutionDetails,
-  StandardResolutionReasons
+  SafeLogger,
+  StandardResolutionReasons,
 } from '@openfeature/shared';
 import { OpenFeature } from './open-feature';
-import { Client, FlagEvaluationOptions, Handler, Hook, OpenFeatureEventEmitter, Provider, ProviderEvents } from './types';
+import {
+  Client,
+  FlagEvaluationOptions,
+  Handler,
+  Hook,
+  OpenFeatureEventEmitter,
+  Provider,
+  ProviderEvents,
+} from './types';
 
 type OpenFeatureClientOptions = {
   name?: string;
@@ -21,28 +31,31 @@ type OpenFeatureClientOptions = {
 };
 
 export class OpenFeatureClient implements Client {
-  readonly metadata: ClientMetadata;
   private _hooks: Hook[] = [];
   private _clientLogger?: Logger;
 
   constructor(
     // functions are passed here to make sure that these values are always up to date,
-    // and so we don't have to make these public properties on the API class. 
+    // and so we don't have to make these public properties on the API class.
     private readonly providerAccessor: () => Provider,
-    private readonly providerReady: () => boolean,
     private readonly events: () => OpenFeatureEventEmitter,
     private readonly globalLogger: () => Logger,
-    options: OpenFeatureClientOptions,
-  ) {
-    this.metadata = {
-      name: options.name,
-      version: options.version,
-    } as const;
+    private readonly options: OpenFeatureClientOptions
+  ) {}
+
+  get metadata(): ClientMetadata {
+    return {
+      name: this.options.name,
+      version: this.options.version,
+      provider: this.providerAccessor().metadata,
+    };
   }
 
   addHandler(eventType: ProviderEvents, handler: Handler): void {
     this.events().on(eventType, handler);
-    if (eventType === ProviderEvents.Ready && this.providerReady()) {
+    const providerReady = !this._provider.status || this._provider.status === ProviderStatus.READY;
+
+    if (eventType === ProviderEvents.Ready && providerReady) {
       // run immediately, we're ready.
       handler();
     }
@@ -67,11 +80,7 @@ export class OpenFeatureClient implements Client {
     return this;
   }
 
-  getBooleanValue(
-    flagKey: string,
-    defaultValue: boolean,
-    options?: FlagEvaluationOptions
-  ): boolean {
+  getBooleanValue(flagKey: string, defaultValue: boolean, options?: FlagEvaluationOptions): boolean {
     return this.getBooleanDetails(flagKey, defaultValue, options).value;
   }
 
@@ -80,20 +89,10 @@ export class OpenFeatureClient implements Client {
     defaultValue: boolean,
     options?: FlagEvaluationOptions
   ): EvaluationDetails<boolean> {
-    return this.evaluate<boolean>(
-      flagKey,
-      this._provider.resolveBooleanEvaluation,
-      defaultValue,
-      'boolean',
-      options
-    );
+    return this.evaluate<boolean>(flagKey, this._provider.resolveBooleanEvaluation, defaultValue, 'boolean', options);
   }
 
-  getStringValue<T extends string = string>(
-    flagKey: string,
-    defaultValue: T,
-    options?: FlagEvaluationOptions
-  ): T {
+  getStringValue<T extends string = string>(flagKey: string, defaultValue: T, options?: FlagEvaluationOptions): T {
     return this.getStringDetails<T>(flagKey, defaultValue, options).value;
   }
 
@@ -112,11 +111,7 @@ export class OpenFeatureClient implements Client {
     );
   }
 
-  getNumberValue<T extends number = number>(
-    flagKey: string,
-    defaultValue: T,
-    options?: FlagEvaluationOptions
-  ): T {
+  getNumberValue<T extends number = number>(flagKey: string, defaultValue: T, options?: FlagEvaluationOptions): T {
     return this.getNumberDetails(flagKey, defaultValue, options).value;
   }
 
@@ -153,12 +148,7 @@ export class OpenFeatureClient implements Client {
 
   private evaluate<T extends FlagValue>(
     flagKey: string,
-    resolver: (
-      flagKey: string,
-      defaultValue: T,
-      context: EvaluationContext,
-      logger: Logger
-    ) => ResolutionDetails<T>,
+    resolver: (flagKey: string, defaultValue: T, context: EvaluationContext, logger: Logger) => ResolutionDetails<T>,
     defaultValue: T,
     flagType: FlagValueType,
     options: FlagEvaluationOptions = {}
--- a/packages/client/src/client.ts
+++ b/packages/client/src/client.ts
@@ -47,7 +47,7 @@ export class OpenFeatureClient implements Client {
     return {
       name: this.options.name,
       version: this.options.version,
-      provider: this.providerAccessor().metadata,
+      providerMetadata: this.providerAccessor().metadata,
     };
   }
 
--- a/packages/client/src/no-op-provider.ts
+++ b/packages/client/src/no-op-provider.ts
@@ -12,6 +12,12 @@ class NoopFeatureProvider implements Provider {
   } as const;
 
   get status(): ProviderStatus {
+    /**
+     * This is due to the NoopProvider not being a real provider.
+     * We do not want it to trigger the Ready event handlers, so we never set this to ready.
+     * With the NoopProvider assigned, the client can be assumed to be uninitialized.
+     * https://github.com/open-feature/js-sdk/pull/429#discussion_r1202642654
+     */
     return ProviderStatus.NOT_READY;
   }
 
--- a/packages/client/src/no-op-provider.ts
+++ b/packages/client/src/no-op-provider.ts
@@ -1,4 +1,4 @@
-import { ResolutionDetails, JsonValue } from '@openfeature/shared';
+import { JsonValue, ProviderStatus, ResolutionDetails } from '@openfeature/shared';
 import { Provider } from './types';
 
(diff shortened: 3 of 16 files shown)


packages/client/src/client.ts
```diff
@@ -1,48 +1,61 @@
...
-      version: options.version,
-    } as const;
+    private readonly options: OpenFeatureClientOptions
+  ) {}
+
+  get metadata(): ClientMetadata {
+    return {
+      name: this.options.name,
+      version: this.options.version,
+      provider: this.providerAccessor().metadata,
```
beeme1mr, line 50: FYI @toddbaert, I talked to @lukas-reining about adding this. I don't think it's specced out but it makes sense to include provider metadata in the client now that we have multi-provider support.

packages/client/src/client.ts
```diff
@@ -1,48 +1,61 @@
...
+    };
   }
 
   addHandler(eventType: ProviderEvents, handler: Handler): void {
     this.events().on(eventType, handler);
-    if (eventType === ProviderEvents.Ready && this.providerReady()) {
+    const providerReady = !this._provider.status || this._provider.status === ProviderStatus.READY;
```
beeme1mr, line 56: I'm assuming `!this._provider.status` is there to support backward compatibility for providers that haven't set a status. If that's the case, could you add a comment stating that? This is something that may be confusing to future devs if they don't understand the history.

packages/client/src/no-op-provider.ts
```diff
@@ -11,6 +11,10 @@ class NoopFeatureProvider implements Provider {
     name: 'No-op Provider',
   } as const;
 
+  get status(): ProviderStatus {
+    return ProviderStatus.NOT_READY;
```
beeme1mr, line 15: Shouldn't the no op provider always be ready?

packages/client/src/open-feature.ts
```diff
@@ -80,58 +81,107 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI {
...
+  setProvider(provider: Provider): this;
+  /**
+   * Sets the provider that OpenFeature will use for flag evaluations of providers with the given name.
+   * Setting a provider supersedes the current provider used in new and existing clients with that name.
+   *
+   * @param {string} client The name to identify the client
+   * @param {Provider} provider The provider responsible for flag evaluations.
+   * @returns {GlobalApi} OpenFeature API
+   */
+  setProvider(client: string, provider: Provider): this;
```
beeme1mr, line 100:
```suggestion
  setProvider(clientName: string, provider: Provider): this;
```

Please update the jsdocs if you agree with this suggestion.

packages/client/src/open-feature.ts
```diff
@@ -80,58 +81,107 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI {
...
+  setProvider(clientOrProvider?: string | Provider, providerOrUndefined?: Provider): this {
+    const client = stringOrUndefined(clientOrProvider);
+    const provider = objectOrUndefined<Provider>(clientOrProvider) ?? objectOrUndefined<Provider>(providerOrUndefined);
+
+    if (!provider) {
+      return this;
+    }
+
+    const oldProvider = client ? this._providers.get(client) ?? this._provider : this._provider;
```
beeme1mr, line 109:
```suggestion
    const oldProvider = this.getProviderForClient(client);
```

packages/server/src/open-feature.ts
```diff
@@ -71,22 +73,107 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI {
...
+  setProvider(provider: Provider): this;
+  /**
+   * Sets the provider that OpenFeature will use for flag evaluations of clients with the given name.
+   * Setting a provider supersedes the current provider used in new and existing clients with that name.
+   *
+   * @param {string} client The name to identify the client
+   * @param {Provider} provider The provider responsible for flag evaluations.
+   * @returns {GlobalApi} OpenFeature API
+   */
+  setProvider(client: string, provider: Provider): this;
```
beeme1mr, line 92:
```suggestion
  setProvider(clientName: string, provider: Provider): this;
```

Same suggestion as in the client sdk.

packages/server/src/open-feature.ts
```diff
@@ -71,22 +73,107 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI {
...
+    nameOrContext?: string | EvaluationContext,
+    versionOrContext?: string | EvaluationContext,
+    contextOrUndefined?: EvaluationContext
+  ): Client {
+    const name = stringOrUndefined(nameOrContext);
+    const version = stringOrUndefined(versionOrContext);
+    const context =
+      objectOrUndefined<EvaluationContext>(nameOrContext) ??
+      objectOrUndefined<EvaluationContext>(versionOrContext) ??
+      objectOrUndefined<EvaluationContext>(contextOrUndefined);
```
beeme1mr, line 160: Nice 😄

packages/server/src/types.ts
```diff
@@ -339,24 +340,76 @@ export interface GlobalApi
...
+
+  /**
+   * Sets the provider that OpenFeature will use for flag evaluations of clients with the given name.
+   * Setting a provider supersedes the current provider used in new and existing clients with that name.
+   *
+   * @param {string} client The name to identify the client
+   * @param {Provider} provider The provider responsible for flag evaluations.
+   * @returns {GlobalApi} OpenFeature API
+   */
+  setProvider(client: string, provider: Provider): GlobalApi;
```
beeme1mr, line 412:
```suggestion
  setProvider(clientName: string, provider: Provider): GlobalApi;
```

beeme1mr (commented):
I've left some feedback, but it looks good to me.

packages/client/src/client.ts
lukas-reining: Yeah exactly.
This is from Requirement 2.4.2:  `Providers without this field can be assumed to be ready immediately.`
I surely can add a comment.

packages/client/src/open-feature.ts
lukas-reining: Makes sense, I will do.

packages/server/src/open-feature.ts
lukas-reining: What does that mean? 😅
Should I change that?

beeme1mr: No, all good. It's a nice way to handle relatively complex overloads.

packages/client/src/no-op-provider.ts
lukas-reining: Formerly, the default no-op provider never got ready, because only after calling setProvider the providerReady flag was set to true.
All the tests for the events depend on the ready flag only being called once.
But because `Requirement 2.4.2: Providers without this field can be assumed to be ready immediately ` also for the no-op provider we would have a provider-ready flag which means in the tests, the provider-ready handler is at least called twice.
I thought as the no-op is the "fallback" anyways I would just set it to NOT_READY but well I also see that this may be wrong.
Not sure what to do about this.

beeme1mr: Would the no op provider ever be set to `READY`? @toddbaert any thoughts on this?

beeme1mr: Yep, makes sense. A comment would clear up my concerns. Thanks for the update.

lukas-reining: Added it here @beeme1mr https://github.com/open-feature/js-sdk/pull/429/commits/59621b664d4edbc6a7692070799d9c82c0647fa6# 🙂

packages/client/src/client.ts
toddbaert: I see why we'd want this. Though I wonder if it might be better to call it `providerMetadata` to make it clearer.

packages/client/src/client.ts
toddbaert: I understand why you did it. I agree lots of comment doesn't hurt.

packages/client/e2e/step-definitions/evaluation.spec.ts
```diff
@@ -24,7 +24,9 @@ const givenAnOpenfeatureClientIsRegisteredWithCacheDisabled = (
 defineFeature(feature, (test) => {
   beforeAll((done) => {
     client.addHandler(ProviderEvents.Ready, () => {
-      done();
+      setTimeout(() => {
+        done(); // TODO Fix by setting readiness correctly in provider
+      }, 2000);
```
toddbaert, line 29: I don't quite understand this... can we avoid this timeout? Is this a TODO for this PR or to fix later?

packages/client/test/open-feature.spec.ts
```diff
@@ -76,11 +107,23 @@ describe('OpenFeature', () => {
       expect(namedClient.metadata.name).toEqual(name);
     });
 
+    it('should return a client with the default provider if no provider has been bound to the name', () => {
+      const namedClient = OpenFeature.getClient('unbound');
+      expect(namedClient.metadata.provider.name).toEqual(OpenFeature.providerMetadata.name);
+    });
+
+    it('should return a client with the provider bound to the name', () => {
+      const name = 'my-client';
```
toddbaert, line 116:
```suggestion
      const name = 'my-named-client';
```

packages/server/src/open-feature.ts
```diff
@@ -71,22 +73,107 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI {
     return this;
   }
 
-  setProvider(provider: Provider): this {
-    // ignore no-ops
-    if (this._provider !== provider) {
+  /**
+   * Sets the provider that OpenFeature will use for flag evaluations of clients without a name.
```
toddbaert, line 77:
```suggestion
   * Sets the default provider that OpenFeature will use for flag evaluations.
   * This provider will be used by unnamed clients and name clients to which no provider is bound.
```

packages/server/src/open-feature.ts
```diff
@@ -15,6 +16,7 @@ const _globalThis = globalThis as OpenFeatureGlobal;
 export class OpenFeatureAPI extends OpenFeatureCommonAPI {
   protected _hooks: Hook[] = [];
   protected _provider: Provider = NOOP_PROVIDER;
```
toddbaert, line 18: Nit: we might want to consider naming this `_defaultProvider`

packages/client/src/open-feature.ts
```diff
@@ -80,58 +81,107 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI {
...
+   * Sets the provider that OpenFeature will use for flag evaluations of providers with the given name.
+   * Setting a provider supersedes the current provider used in new and existing clients with that name.
+   *
+   * @param {string} clientName The name to identify the client
+   * @param {Provider} provider The provider responsible for flag evaluations.
+   * @returns {GlobalApi} OpenFeature API
+   */
+  setProvider(clientName: string, provider: Provider): this;
+  setProvider(clientOrProvider?: string | Provider, providerOrUndefined?: Provider): this {
+    const client = stringOrUndefined(clientOrProvider);
```
toddbaert, line 102:
```suggestion
    const clientName = stringOrUndefined(clientOrProvider);
```

packages/client/test/events.spec.ts
```diff
@@ -2,198 +2,235 @@ import {
...
 describe('Events', () => {
   // set timeouts short for this suite.
   jest.setTimeout(1000);
-  let API: OpenFeatureAPI = new (OpenFeatureAPI as any)();
+  let clientId = uuid();
 
   afterEach(() => {
-    jest.clearAllMocks();
-    API = new (OpenFeatureAPI as any)(); //TODO this hacky thing should be removed with provider mapping: https://github.com/open-feature/js-sdk/issues/421
+    // jest.clearAllMocks();
```
toddbaert, line 83: Should this be commented?

packages/server/src/open-feature.ts
```diff
@@ -71,22 +73,107 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI {
...
+   * Setting a provider supersedes the current provider used in new and existing clients with that name.
+   *
+   * @param {string} clientName The name to identify the client
+   * @param {Provider} provider The provider responsible for flag evaluations.
+   * @returns {GlobalApi} OpenFeature API
+   */
+  setProvider(clientName: string, provider: Provider): this;
+  setProvider(clientOrProvider?: string | Provider, providerOrUndefined?: Provider): this {
+    const client = stringOrUndefined(clientOrProvider);
+    const provider = objectOrUndefined<Provider>(clientOrProvider) ?? objectOrUndefined<Provider>(providerOrUndefined);
```
toddbaert, line 95: We might be able to make more of of the `setProvider` function common when we add `events` and `init/shutdown` features to the server, with some help from generics.

toddbaert:
I left a few minor comments, but this looks very good to me. Take a look at my comments but I think I can approve this soon.

packages/client/test/events.spec.ts
lukas-reining: @toddbaert I changed that now.
The problem was, that there was actually a problem with the implementation of the events.
Events that came up in one provider have been emitted to every client.
I changed that a bit so that we have global and client events.
I would continue on that in my PR for events 🙂

packages/server/src/open-feature.ts
lukas-reining: Makes sense, I also thought about that and would try to add it in the events PR.
Does that make sense to you?

packages/client/src/client.ts
lukas-reining: Changed it.

packages/client/e2e/step-definitions/evaluation.spec.ts
lukas-reining: The spec says, that a provider without the status flag should be considered ready.
As the FlagdProvider certainly does not provide that right now, the code is directly called before the provider is initialized. This leads to the tests failing.
We could try to find a better way, but @beeme1mr and me discussed it and said, that we would just accept this now and fix the FlagD-Provider asap.

packages/client/src/open-feature.ts
lukas-reining: Changed.

packages/client/src/client.ts
lukas-reining: So should we do something here?

packages/server/src/open-feature.ts
toddbaert: > Makes sense, I also thought about that and would try to add it in the events PR. Does that make sense to you?

Yes, or even after as a pure refactor. I don't mind either way.

packages/client/test/events.spec.ts
toddbaert: Yes, when we implement events in server we should also add tests for this specific use-case (check that provider x doesn't cause events for a client associated with provider y).

Since events are only implemented experimentally in web, I'm OK with doing that in a later PR.

packages/client/e2e/step-definitions/evaluation.spec.ts
toddbaert: https://github.com/open-feature/js-sdk/pull/429/files/59621b664d4edbc6a7692070799d9c82c0647fa6#r1204327488

packages/client/e2e/step-definitions/evaluation.spec.ts
```diff
@@ -24,7 +24,9 @@ const givenAnOpenfeatureClientIsRegisteredWithCacheDisabled = (
 defineFeature(feature, (test) => {
   beforeAll((done) => {
     client.addHandler(ProviderEvents.Ready, () => {
-      done();
+      setTimeout(() => {
+        done(); // TODO Fix by setting readiness correctly in provider
```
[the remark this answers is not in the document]
toddbaert: Thanks for your explanation about this. I've tried to make it a tiny bit clearer.

toddbaert (approved):

lukas-reining:
> Approved, thanks! There are a few minor lint warnings that may be worth addressing before you merge

Nice thanks @beeme1mr!
I fixed the linter warnings too now.
I am not allowed to merge the branch, so how do we do it @beeme1mr @toddbaert?

toddbaert:
> > Approved, thanks! There are a few minor lint warnings that may be worth addressing before you merge
> 
> Nice thanks @beeme1mr! I fixed the linter warnings too now. I am not allowed to merge the branch, so how do we do it @beeme1mr @toddbaert?

rebased and merged

Issue #424, opened 2023-05-16
Changes to support spec `v0.6.0` compliance


toddbaert:
## Implement specification v0.6.0 changes

See [changes since last release](https://github.com/open-feature/spec/compare/v0.5.2...v0.6.0).

Please note there's no need to complete all these tasks in a single PR (in fact, that's not recommended).

### Functional requirements:

- [x] #425
- [x] #428
- [x] implement initialization/shutdown: https://github.com/open-feature/spec/commit/a4ffec3d4416b8ad6d12d1ebfa2faf81ad153f65
- [x] implement events: https://github.com/open-feature/spec/commit/bf9e45b1d51c316ee2d6340bc2b807294995d0ad

### Non-functional requirements:

- [x] common types/code implemented in `packages/shared`
  - :mega:  **_much is already implemented in `packages/client` and will need to be moved to `packages/shared`_** :mega: 
- [x] add relevant tests
- [x] #440
- [x] update specification badge to `v0.6.0`: [![Specification](https://img.shields.io/static/v1?label=Specification&message=v0.6.0&color=yellow)](https://github.com/open-feature/spec/tree/v0.6.0)

### Keep in mind:

- backwards compatibility is _highly encouraged_
  - take every precaution to ensure existing code written by _application authors_ and existing providers implementations, do not break
- it's recommended these features are released together

lukas-reining:
As said I could work on this @toddbaert 🙂

toddbaert:
@lukas-reining @beeme1mr WRT events, I'm wondering whether or not we should expose events on the client using the OpenFeatureEventEmitter (which is basically a node EventEmitter) or not...

So basically... should we do a custom method:

client.addhandler(ProviderEvent, handlerFn)

or just leverage native events:

client.events.on(ProviderEvent, handlerFn)

lukas-reining:
@toddbaert I would say having  the `OpenFeatureEventEmitter` and also exposing it from the client makes sense.

But instead of just aliasing it, I would make this an interface that we can implement.
Having the class OpenFeatureEventEmitter and specifying exactly what we need e.g. `client.addHandler(ProviderEvent, handlerFn)` would decouple us from NodeEventEmitter.
This could make it easier to test and also enables us to define exactly what we need for the interface instead of having all the functionality of `NodeEventEmitter` leak through.

> Each have their own benefits. We have full control over a custom implementation, but on the other hand the EventEmitter route gives us a lot for free (once, removal, etc)

[...]

PR #436, opened 2023-05-25
feat: add init/shutdown and events

### LINKED ISSUE #434: implement initialization/shutdown: open-feature/spec@a4ffec3
### Requirements

Add initialization/shutdown support to the JS SDK based on the recent update to the spec: https://github.com/open-feature/spec/commit/a4ffec3d4416b8ad6d12d1ebfa2faf81ad153f65

### LINKED ISSUE #435: implement events: open-feature/spec@bf9e45b
### Requirements

Add initialization/shutdown support to the JS SDK based on the recent update to the spec: https://github.com/open-feature/spec/commit/bf9e45b1d51c316ee2d6340bc2b807294995d0ad

lukas-reining:
## This PR

Implements init/shutdown and events support from spec version 0.6.0.  #424  #434 #435 

### Related Issues

Fixes #434 
Fixes #435 

### Notes
Moved as much code as possible to the shared package.
Maybe we could even share some more code of the client.

### How to test

### DIFF
--- a/packages/client/e2e/step-definitions/evaluation.spec.ts
+++ b/packages/client/e2e/step-definitions/evaluation.spec.ts
@@ -23,15 +23,15 @@ const givenAnOpenfeatureClientIsRegisteredWithCacheDisabled = (
 
 defineFeature(feature, (test) => {
   beforeAll((done) => {
-    client.addHandler(ProviderEvents.Ready, () => {
+    client.addHandler(ProviderEvents.Ready, async () => {
       setTimeout(() => {
         done(); // TODO remove this once flagd provider properly implements readiness (for now, we add a 2s wait).
       }, 2000);
     });
   });
 
-  afterAll(() => {
-    OpenFeature.close();
+  afterAll(async () => {
+    await OpenFeature.close();
   });
 
   test('Resolves boolean value', ({ given, when, then }) => {
--- a/packages/server/e2e/step-definitions/evaluation.spec.ts
+++ b/packages/server/e2e/step-definitions/evaluation.spec.ts
@@ -1,5 +1,13 @@
 import { defineFeature, loadFeature } from 'jest-cucumber';
-import { JsonValue, JsonObject, EvaluationDetails, EvaluationContext, ResolutionDetails, StandardResolutionReasons } from '@openfeature/shared';
+import {
+  JsonValue,
+  JsonObject,
+  EvaluationDetails,
+  EvaluationContext,
+  ResolutionDetails,
+  StandardResolutionReasons,
+  ProviderEvents,
+} from '@openfeature/shared';
 import { OpenFeature } from './../../src';
 // load the feature file.
 const feature = loadFeature('packages/server/e2e/features/evaluation.feature');
@@ -7,12 +15,25 @@ const feature = loadFeature('packages/server/e2e/features/evaluation.feature');
 // get a client (flagd provider registered in setup)
 const client = OpenFeature.getClient();
 
-const givenAnOpenfeatureClientIsRegisteredWithCacheDisabled = (given: (stepMatcher: string, stepDefinitionCallback: () => void) => void) => {
+const givenAnOpenfeatureClientIsRegisteredWithCacheDisabled = (
+  given: (stepMatcher: string, stepDefinitionCallback: () => void) => void
+) => {
   // TODO: when the FlagdProvider is updated to support caching, we may need to disable it here for this test to work as expected.
   given('a provider is registered with cache disabled', () => undefined);
 };
 
 defineFeature(feature, (test) => {
+  beforeAll((done) => {
+    client.addHandler(ProviderEvents.Ready, async () => {
+      setTimeout(() => {
+        done(); // TODO remove this once flagd provider properly implements readiness (for now, we add a 2s wait).
+      }, 2000);
+    });
+  });
+
+  afterAll(async () => {
+    await OpenFeature.close();
+  });
 
   test('Resolves boolean value', ({ given, when, then }) => {
     let value: boolean;
@@ -265,10 +286,13 @@ defineFeature(feature, (test) => {
       }
     );
 
-    and(/^a flag with key "(.*)" is evaluated with default value "(.*)"$/, async (key: string, defaultValue: string) => {
-      flagKey = key;
-      value = await client.getStringValue(flagKey, defaultValue, context);
-    });
+    and(
+      /^a flag with key "(.*)" is evaluated with default value "(.*)"$/,
+      async (key: string, defaultValue: string) => {
+        flagKey = key;
+        value = await client.getStringValue(flagKey, defaultValue, context);
+      }
+    );
 
     then(/^the resolved string response should be "(.*)"$/, (expectedValue: string) => {
       expect(value).toEqual(expectedValue);
--- a/packages/server/src/no-op-provider.ts
+++ b/packages/server/src/no-op-provider.ts
@@ -1,4 +1,4 @@
-import { ResolutionDetails, JsonValue } from '@openfeature/shared';
+import { ResolutionDetails, JsonValue, ProviderStatus } from '@openfeature/shared';
 import { Provider } from './types';
 
 const REASON_NO_OP = 'No-op';
@@ -10,6 +10,16 @@ class NoopFeatureProvider implements Provider {
   readonly metadata = {
     name: 'No-op Provider',
   } as const;
+
+  get status(): ProviderStatus {
+    /**
+     * This is due to the NoopProvider not being a real provider.
+     * We do not want it to trigger the Ready event handlers, so we never set this to ready.
+     * With the NoopProvider assigned, the client can be assumed to be uninitialized.
+     * https://github.com/open-feature/js-sdk/pull/429#discussion_r1202642654
+     */
+    return ProviderStatus.NOT_READY;
+  }
 
   resolveBooleanEvaluation(_: string, defaultValue: boolean): Promise<ResolutionDetails<boolean>> {
     return this.noOp(defaultValue);
--- a/packages/shared/src/events.ts
+++ b/packages/shared/src/events.ts
@@ -1,0 +1,106 @@
+import { Logger, ManageLogger } from './types';
+import EventEmitter from 'events';
+
+export type EventMetadata = {
+  [key: string]: string | boolean | number;
+};
+
+export type EventDetails = {
+  clientName?: string;
+  message?: string;
+  flagsChanged?: string[];
+  metadata?: EventMetadata;
+};
+
+export type EventHandler = (eventDetails?: EventDetails) => Promise<unknown>;
+
+export interface Eventing {
+  /**
+   * Adds a handler for the given provider event type.
+   * The handlers are called in the order they have been added.
+   * @param {ProviderEvents} eventType The provider event type to listen to
+   * @param {EventHandler} handler The handler to run on occurrence of the event type
+   */
+  addHandler(eventType: ProviderEvents, handler: EventHandler): void;
+
+  /**
+   * Removes a handler for the given provider event type.
+   * @param {ProviderEvents} eventType The provider event type to remove the listener for
+   * @param {EventHandler} handler The handler to remove for the provider event type
+   */
+  removeHandler(eventType: ProviderEvents, handler: EventHandler): void;
+
+  /**
+   * Gets the current handlers for the given provider event type.
+   * @param {ProviderEvents} eventType The provider event type to get the current handlers for
+   * @returns {EventHandler[]} The handlers currently attached to the given provider event type
+   */
+  getHandlers(eventType: ProviderEvents): EventHandler[];
+}
+
+export enum ProviderEvents {
+  /**
+   * The provider is ready to evaluate flags.
+   */
+  Ready = 'PROVIDER_READY',
+
+  /**
+   * The provider is in an error state.
+   */
+  Error = 'PROVIDER_ERROR',
+
+  /**
+   * The flag configuration in the source-of-truth has changed.
+   */
+  ConfigurationChanged = 'PROVIDER_CONFIGURATION_CHANGED',
+
+  /**
+   * The provider's cached state is not longer valid and may not be up-to-date with the source of truth.
+   */
+  Stale = 'PROVIDER_STALE',
+}
+
+export class OpenFeatureEventEmitter implements ManageLogger<OpenFeatureEventEmitter> {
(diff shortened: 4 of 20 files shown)


packages/client/e2e/step-definitions/evaluation.spec.ts
```diff
@@ -23,15 +23,15 @@ const givenAnOpenfeatureClientIsRegisteredWithCacheDisabled = (
 
 defineFeature(feature, (test) => {
   beforeAll((done) => {
-    client.addHandler(ProviderEvents.Ready, () => {
+    client.addHandler(ProviderEvents.Ready, async () => {
```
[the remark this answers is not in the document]
toddbaert: I don't think this is needed.

packages/shared/src/events.ts
```diff
@@ -0,0 +1,106 @@
+import { Logger, ManageLogger } from './types';
```
toddbaert, line 1: To me it seems a little inconsistent that we have `types` which contains all our types except the event-related types which are here.

I think we should be consistent and either keep everything in a single types file, or create a `types/` dir with a few different files in it, one of which would be this one.

If we want to go the second route, I suggest we create a refactor ticket and we keep this the way it is in this PR. If you want to put everything in the one types file, please do that in this PR.

What do you think @lukas-reining ?

packages/server/e2e/step-definitions/evaluation.spec.ts
```diff
@@ -1,18 +1,39 @@
...
+const givenAnOpenfeatureClientIsRegisteredWithCacheDisabled = (
+  given: (stepMatcher: string, stepDefinitionCallback: () => void) => void
+) => {
   // TODO: when the FlagdProvider is updated to support caching, we may need to disable it here for this test to work as expected.
   given('a provider is registered with cache disabled', () => undefined);
 };
 
 defineFeature(feature, (test) => {
+  beforeAll((done) => {
+    client.addHandler(ProviderEvents.Ready, async () => {
```
toddbaert, line 27:
```suggestion
    client.addHandler(ProviderEvents.Ready, () => {
```

toddbaert: Same as [here](https://github.com/open-feature/js-sdk/pull/436/files#r1205813869). Let me know I'm I'm wrong for some reason.

packages/server/src/no-op-provider.ts
```diff
@@ -11,6 +11,16 @@ class NoopFeatureProvider implements Provider {
     name: 'No-op Provider',
   } as const;
 
+  get status(): ProviderStatus {
+    /**
+     * This is due to the NoopProvider not being a real provider.
+     * We do not want it to trigger the Ready event handlers, so we never set this to ready.
+     * With the NoopProvider assigned, the client can be assumed to be uninitialized.
+     * https://github.com/open-feature/js-sdk/pull/429#discussion_r1202642654
```
toddbaert, line 19: Very good explanation to me.

packages/shared/src/events.ts
lukas-reining: I would go for the types /dir and do the refactor ticket.

packages/shared/src/open-feature.ts
```diff
@@ -1,21 +1,208 @@
...
+  get providerMetadata(): ProviderMetadata {
+    return this._defaultProvider.metadata;
+  }
 
   getContext(): EvaluationContext {
     return this._context;
   }
 
-  setTransactionContextPropagator(transactionContextPropagator: TransactionContextPropagator): OpenFeatureCommonAPI {
+  /**
```
toddbaert, line 44: I'm very happy with how much of this was able to be implemented in `/shared`

packages/client/e2e/step-definitions/evaluation.spec.ts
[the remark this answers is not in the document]
lukas-reining: The problem is, that EventEmitter only catches errors in async functions. 
If you have a normal function that throws an error,the other handlers are not called anymore. That's the reason I decided to go only for async.

packages/shared/src/open-feature.ts
```diff
@@ -1,21 +1,208 @@
...
+    // ignore no-ops
+    if (oldProvider === provider) {
+      return this;
+    }
+
+    if (clientName) {
+      this._clientProviders.set(clientName, provider);
+    } else {
+      this._defaultProvider = provider;
+    }
```
toddbaert, line 110: I think there's one MINOR thing here that's from from the spec: https://github.com/open-feature/spec/blob/main/specification/sections/01-flag-evaluation.md#requirement-1122

> The provider mutator function MUST invoke the initialize function on the newly registered provider before using it to resolve flag values.

We should be initializing providers and THEN setting them, so that things "keep working" if the provider is switched.

packages/shared/src/open-feature.ts
```diff
@@ -1,21 +1,208 @@
...
+    );
+  }
+
+  private handleShutdownError(provider: P, err: unknown) {
+    this._logger.error(`Error during shutdown of provider ${provider.metadata.name}: ${err}`);
+
+    if (err instanceof Error) {
+      this._logger.error(err.stack);
+    }
+    this._logger.error((err as Error)?.stack);
```
toddbaert, line 202: This seems odd. We are logging the stack no matter what, so no need to check I think?

Apologies if I wrote this and you moved it, not sure. :sweat_smile:

lukas-reining: That is true😅
And yes it is true I moved it from the other place.

packages/shared/src/open-feature.ts
lukas-reining: Makes sense, I will change it.

packages/shared/src/events.ts
toddbaert: I will create the ticket!

packages/shared/src/open-feature.ts
toddbaert, line 110: I accidentally deleted this a comment I made, but I'll retype it:

I think as per [1.1.2.2](https://openfeature.dev/specification/sections/flag-evaluation#requirement-1122) we should only set the provider after initialization to prevent a period of unavailability when the provider has been switched and it still starting up.

packages/client/e2e/step-definitions/evaluation.spec.ts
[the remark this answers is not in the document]
toddbaert: > The problem is, that EventEmitter only catches errors in async functions. If you have a normal function that throws an error,the other handlers are not called anymore. That's the reason I decided to go only for async.

Oh, I think that's actually against spec. See [5.2.5](https://openfeature.dev/specification/sections/events#requirement-525): 


I think we can fix this by running the handler in a try/catch [here](https://github.com/open-feature/js-sdk/pull/436/files#diff-1d25041624e23bb1be334f0c7033cf4fadf2801dae21687f77f5511cd5935b92R55)?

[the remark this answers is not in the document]
toddbaert: If I'm right, please add a test for this.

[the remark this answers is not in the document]
lukas-reining: I actually have a test for it.

packages/shared/src/open-feature.ts
lukas-reining: I will change it.

packages/shared/src/events.ts
toddbaert: Created: https://github.com/open-feature/js-sdk/issues/437

packages/client/e2e/step-definitions/evaluation.spec.ts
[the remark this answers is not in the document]
toddbaert: Ah, ok, makes sense to me as long as the type system enforces it as you say.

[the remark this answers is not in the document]
lukas-reining: So should I change that?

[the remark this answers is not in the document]
toddbaert: No if I can only add async functions and thrown errors in async functions keep the other handlers running, I agree this is within spec and a good solution.

toddbaert:
@lukas-reining I'll manually test this tomorrow (just to make sure everything still bundles/packages nicely) but once the remaining issues are closed it think this is good to merge from my perspective.

packages/shared/src/open-feature.ts
lukas-reining: Changed!

packages/shared/src/open-feature.ts
lukas-reining: Changed!

packages/shared/src/events.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  setLogger(logger: Logger): this {
+    this._eventLogger = logger;
+    return this;
+  }
+
+  private get _logger() {
+    return this._eventLogger || this.globalLogger?.();
+  }
+}
```
lukas-reining, line 106: @toddbaert This would be the solution if we want sync and async handlers.
This would make it safe for javascript too.
I think I would go for this, what do you think?
```suggestion
export class OpenFeatureEventEmitter implements ManageLogger<OpenFeatureEventEmitter> {
  private readonly _handlers = new Map<EventHandler, EventHandler>();
  private readonly eventEmitter = new EventEmitter({ captureRejections: true });
  private _eventLogger?: Logger;

  constructor(private readonly globalLogger?: () => Logger) {
    this.eventEmitter.on('error', (err) => {
      this._logger?.error('Error running event handler:', err);
    });
  }

  emit(eventType: ProviderEvents, context?: EventDetails): void {
    this.eventEmitter.emit(eventType, context);
  }

  addHandler(eventType: ProviderEvents, handler: EventHandler): void {
    const asyncHandler = async (context?: EventDetails) => {
      await handler(context);
    };

    this._handlers.set(handler, asyncHandler);
    this.eventEmitter.on(eventType, asyncHandler);
  }

  removeHandler(eventType: ProviderEvents, handler: EventHandler): void {
    const asyncHandler = this._handlers.get(handler);

    if (!asyncHandler) {
      return;
    }

    this.eventEmitter.removeListener(eventType, asyncHandler);
  }

  removeAllHandlers(eventType?: ProviderEvents): void {
    // If giving undefined, the listeners are not removed, so we have to check explicitly
    if (eventType) {
      this.eventEmitter.removeAllListeners(eventType);
    } else {
      this.eventEmitter.removeAllListeners();
    }
  }

  getHandlers(eventType: ProviderEvents): EventHandler[] {
    return this.eventEmitter.listeners(eventType) as EventHandler[];
  }

  setLogger(logger: Logger): this {
    this._eventLogger = logger;
    return this;
  }

  private get _logger() {
    return this._eventLogger || this.globalLogger?.();
  }
}
```

toddbaert: I think this is an elegant solution to the problem. The use of the weakmap indexed by the wrapped handler seems like a good idea.

I'm not sure the `await` on line 80 is actually needed. I don't _think_ there's a need to await handlers since we don't have any ordering guarantees. We could add a `.catch()` and log errors in the execution of the wrapped handler as well. What do you think about that?

lukas-reining: Nice then let us go for that solution, I will quickly add this 🙂
The wait is actually needed because we have the error listener in the constructor. We could also remove that listener and catch the error in the handler but I would personally do it as it is.

toddbaert: > The wait is actually needed because we have the error listener in the constructor.

Ah yes, I see. That wasn't clear to me initially... but maybe it's better to have the `await` on line 80 anyway since we need to have the `async` keyword on the handler wrapper.

toddbaert: Can you add some comments explaining why the handlers are wrapped and we're doing all this? Basically in a line or two explain what we've been talking about, just for the future.

lukas-reining: > but maybe it's better to have the await on line 80 anyway since we need to have the async keyword on the handler wrapper.

Does that mean we leave it as it is?

lukas-reining: Sure I will add a comment and add the change.

lukas-reining: Done, please check if the comments are okay @toddbaert 🙂

toddbaert: Look great to me!

toddbaert (approved):
I've tested the packaging locally and consumed the web and server SDKs in a web app, and esm and commonjs node apps. Everything looks great and I didn't have any breaking changes.

Nothing else to say except excellent work. Approving.

toddbaert:
I encourage any @open-feature/sdk-javascript-approvers and @open-feature/sdk-javascript-maintainers to review.

packages/shared/src/events.ts
```diff
@@ -0,0 +1,121 @@
...
+   */
+  Error = 'PROVIDER_ERROR',
+
+  /**
+   * The flag configuration in the source-of-truth has changed.
+   */
+  ConfigurationChanged = 'PROVIDER_CONFIGURATION_CHANGED',
+
+  /**
+   * The provider's cached state is not longer valid and may not be up-to-date with the source of truth.
```
beeme1mr, line 58:
```suggestion
   * The provider's cached state is no longer valid and may not be up-to-date with the source of truth.
```

packages/shared/src/events.ts
```diff
@@ -0,0 +1,121 @@
...
+      this.eventEmitter.removeAllListeners();
+    }
+  }
+
+  getHandlers(eventType: ProviderEvents): EventHandler[] {
+    return this.eventEmitter.listeners(eventType) as EventHandler[];
+  }
+
+  setLogger(logger: Logger): this {
+    this._eventLogger = logger;
```
beeme1mr, line 114: Should this use `SafeLogger` like we're doing in the OpenFeatureCommonAPI?

beeme1mr (approved):
I noticed a few little things, but overall it looks great. Thank you very much for all your hard work.

packages/shared/src/events.ts
lukas-reining: Makes sense, I will change that tomorrow

weyert (approved):
Thank you, looks good to me! Excited about the shutdown change :)

packages/shared/src/events.ts
lukas-reining: Added the safe logger @beeme1mr 🙂

thomaspoignant (approved):

PR #441, opened 2023-05-30
chore: add 0.6.0 examples to README.md

### LINKED ISSUE #440: add examples and documentation, see template README

lukas-reining:
## This PR

Adds examples for named providers, events and the global shutdown of spec 0.6.0.
It does not provide an example for flag metadata, as I think an example is not needed here.

### Related Issues

Fixes #440

### DIFF
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -22,11 +22,14 @@ For more information, see this [issue](https://github.com/open-feature/spec/issu
 
 ### What is OpenFeature?
 
-[OpenFeature][openfeature-website] is an open standard that provides a vendor-agnostic, community-driven API for feature flagging that works with your favorite feature flag management tool.
+[OpenFeature][openfeature-website] is an open standard that provides a vendor-agnostic, community-driven API for feature
+flagging that works with your favorite feature flag management tool.
 
 ### Why standardize feature flags?
 
-Standardizing feature flags unifies tools and vendors behind a common interface which avoids vendor lock-in at the code level. Additionally, it offers a framework for building extensions and integrations and allows providers to focus on their unique value proposition.
+Standardizing feature flags unifies tools and vendors behind a common interface which avoids vendor lock-in at the code
+level. Additionally, it offers a framework for building extensions and integrations and allows providers to focus on
+their unique value proposition.
 
 ## 🔍 Requirements:
 
@@ -72,7 +75,8 @@ const boolValue = client.getBooleanValue('boolFlag', false);
 
 ### Context-aware evaluation:
 
-Sometimes the value of a flag must take into account some dynamic criteria about the application or user, such as the user location, IP, email address, or the location of the server.
+Sometimes the value of a flag must take into account some dynamic criteria about the application or user, such as the
+user location, IP, email address, or the location of the server.
 In OpenFeature, we refer to this as [`targeting`](https://openfeature.dev/specification/glossary#targeting).
 If the flag system you're using supports targeting, you can provide the input data using the `EvaluationContext`.
 
@@ -86,7 +90,10 @@ const boolValue = client.getBooleanValue('some-flag', false);
 
 ### Providers:
 
-To develop a provider, you need to create a new project and include the OpenFeature SDK as a dependency. This can be a new repository or included in an existing contrib repository available under the OpenFeature organization. Finally, you’ll then need to write the provider itself. In most languages, this can be accomplished by implementing the provider interface exported by the OpenFeature SDK.
+To develop a provider, you need to create a new project and include the OpenFeature SDK as a dependency. This can be a
+new repository or included in an existing contrib repository available under the OpenFeature organization. Finally,
+you’ll then need to write the provider itself. In most languages, this can be accomplished by implementing the provider
+interface exported by the OpenFeature SDK.
 
 ```typescript
 import { JsonValue, Provider, ResolutionDetails } from '@openfeature/web-sdk';
@@ -94,7 +101,7 @@ import { JsonValue, Provider, ResolutionDetails } from '@openfeature/web-sdk';
 // implement the provider interface
 class MyProvider implements Provider {
   readonly metadata = {
-    name: 'My Provider',
+    name: 'My Provider'
   } as const;
 
   resolveBooleanEvaluation(flagKey: string, defaultValue: boolean): ResolutionDetails<boolean> {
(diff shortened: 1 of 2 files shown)


packages/client/README.md
```diff
@@ -159,6 +172,58 @@ class MyLogger implements Logger {
...
+// Uses YourOtherProviderOfChoice as it is set explicitly
+const client1 = OpenFeature.getClient("client-1")
+
+// Uses YourProviderOfChoice as no provider is set
+const client2 = OpenFeature.getClient("client-2")
+```
+
+### Events:
+
+Events allow to react to state changes in the provider or underlying flag management system.
```
beeme1mr, line 199:
```suggestion
Events provide a way to react to state changes in the provider or underlying flag management system.
```

packages/client/README.md
```diff
@@ -159,6 +172,58 @@ class MyLogger implements Logger {
...
+  console.log(`Ready event from: ${eventDetails.clientName}:`, eventDetails);
+});
+
+// Specific client
+const client = OpenFeature.getClient();
+client.addHandler(ProviderEvents.Error, async (eventDetails) => {
+  console.log(`Error event from: ${eventDetails.clientName}:`, eventDetails);
+});
+```
+
```
beeme1mr, line 216: It may be worth adding a note stating that not all providers support all event types. What do you think @toddbaert?

packages/client/README.md
```diff
@@ -159,6 +172,58 @@ class MyLogger implements Logger {
...
+### Shutdown:
+
+The OpenFeature API provides a close function to perform a cleanup of all providers attached to any client.
+
+```typescript
+import { OpenFeature, ProviderEvents } from '@openfeature/web-sdk';
+
+await OpenFeature.close()
```
beeme1mr, line 224: Perhaps wrap this in a `onbeforeunload` event handler or something similar. It may be confusing to see the close method called directly like this. 🤷‍♂️

packages/server/README.md
```diff
@@ -161,6 +161,58 @@ class MyLogger implements Logger {
...
+```
+
+### Shutdown:
+
+The OpenFeature API provides a close function to perform a cleanup of all providers attached to any client.
+
+```typescript
+import { OpenFeature, ProviderEvents } from '@openfeature/web-sdk';
+
+await OpenFeature.close()
```
beeme1mr, line 213: Similar comment to what I said on the client. Perhaps the example should hook into the on exit process event.

packages/server/README.md
```diff
@@ -161,6 +161,58 @@ class MyLogger implements Logger {
...
+// Uses YourOtherProviderOfChoice as it is set explicitly
+const client1 = OpenFeature.getClient("client-1")
+
+// Uses YourProviderOfChoice as no provider is set
+const client2 = OpenFeature.getClient("client-2")
+```
+
+### Events:
+
+Events allow to react to state changes in the provider or underlying flag management system.
```
beeme1mr, line 188:
```suggestion
Events provide a way to react to state changes in the provider or underlying flag management system.
```

beeme1mr (commented):
Thanks! Overall, it looks great. I'm just not sure how we should document the shutdown section.

packages/server/README.md
lukas-reining: Makes sense, but I am not sure which handler to use.
The Node docs state that the exit handler may only be synchronous. https://nodejs.org/api/process.html#event-exit
We may add a SIGINT handler, that is called then. A bit too much detail but maybe okay? 

```suggestion
// Shutdown on CTRL+C
process.on('SIGINT', async () => {
  await OpenFeature.close();
  process.exit();
});
```

packages/client/README.md
lukas-reining: Here we have the same problem as with the server sdk.
The onbeforeunload may be used to trigger the close, but as it also is only sync, it will propably not finish.
Not sure what to use here.

packages/client/README.md
lukas-reining: I would also add this. People only looking at this example may expect these events to be dispatched by any provider.

Added a sentence to that, not sure if this is good enough.
https://github.com/open-feature/js-sdk/pull/441/commits/7f35b8e07aaad0bec0e4983b98828bcef624f392

packages/server/README.md
weyert: Think it's a clear enough example. I don't think it's necessary to also include `SIGTERM` and its friends for things like Kubernetes etc

packages/server/README.md
```diff
@@ -161,6 +161,62 @@ class MyLogger implements Logger {
...
+OpenFeature.setProvider("client-1", new YourOtherProviderOfChoice())
+
+// Uses YourProviderOfChoice (the default)
+const unnamedClient = OpenFeature.getClient()
+
+// Uses YourOtherProviderOfChoice as it is set explicitly
+const client1 = OpenFeature.getClient("client-1")
+
+// Uses YourProviderOfChoice as no provider is set
+const client2 = OpenFeature.getClient("client-2")
```
weyert, line 183: Can anyone think of a situation were you would like to `close`/`shutdown` a client e.g. only `client2`? If I follow everything correctly at the moment is everything or nothing. You can not only `dispose`/`shutdown` provider2.

lukas-reining: Good question, also shortly thought about it.
The spec at least only defines the shutdown for API and provider🤔

packages/client/README.md
beeme1mr: It looks like your change wasn't included in this branch. Was that intentional?

lukas-reining: Mh I can see it as part of the changes in the files and also see the changes listed🤔

lukas-reining: Or which change do you mean?

packages/server/README.md
```diff
@@ -161,6 +161,62 @@ class MyLogger implements Logger {
...
+```
+
+### Events:
+
+Events provide a way to react to state changes in the provider or underlying flag management system.
+You can listen to events of either the OpenFeature API or individual clients.
+
+Providers do not have to dispatch events, so they may not dispatch any or only some kinds of events.
+The events after initialization, `PROVIDER_READY` on success, `PROVIDER_ERROR` on failure during initialization,
+are dispatched for every provider.
```
beeme1mr, line 193:
```suggestion
The events after initialization, `PROVIDER_READY` on success, `PROVIDER_ERROR` on failure during initialization,
are dispatched for every provider.
However, other event types may not be supported by your provider.
Please refer to the documentation of the provider you're using to see what events are supported.
```

What do you think about this phrasing?

packages/server/README.md
beeme1mr: I can't think of a realistic scenario where someone would want to do this. I think it's best to leave that out until one is identified.

beeme1mr (commented):
I added a suggested rephrasing of the provider event disclaimer. If you agree with the suggestion, please update the server and client readme.

I think we're in a mergeable spot with this but I would recommend we wait to merge until right before we're ready to release the SDK.

packages/server/README.md
lukas-reining: Yeah thats good, I will add that.

PR #445, opened 2023-06-02
fix: use a different symbol for server/client

toddbaert:
Fixes an issue that can cause bunders in repos with both the web-sdk and js-sdk to incorrectly resolve and bundle module components because both modules use the same global symbols for the API.

### DIFF
--- a/packages/client/src/open-feature.ts
+++ b/packages/client/src/open-feature.ts
@@ -4,7 +4,7 @@ import { NOOP_PROVIDER } from './no-op-provider';
 import { Client, Hook, Provider } from './types';
 
 // use a symbol as a key for the global singleton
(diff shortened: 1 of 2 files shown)


toddbaert:
@moredip I'm confident enough that this is the root cause of your issue that I think we should just commit it. I've seen the same behavior myself, but it's very tough to reproduce consistently.

toddbaert:
> I think it make sense to use a different symbol for client and server. I'm just wondering if we should call it client and server instead of `web-sdk` and `js-sdk`. Client/server is what we went with on the main readme.

I had the same thought but went with the actual artifact name because I thought that would make it very clear in debugging. `Client` is a bit overloaded.

PR #447, opened 2023-06-07
fix: updated typedoc config to support monorepos

beeme1mr:
## This PR

- updated typedoc config to support monorepos

### Notes

A recent update the Typedoc changed how `packages` work with the `entryPointStrategy`. The release notes can be found [here](https://github.com/TypeStrong/typedoc/releases/tag/v0.24.0).

This is a temporary workaround because legacy support will be removed in 0.25.0, according to [the docs](https://typedoc.org/options/input/#entrypointstrategy).

### Follow-up Tasks

Update the configuration to use a non legacy strategy.


packages/client/README.md
```diff
@@ -221,7 +221,8 @@ client.addHandler(ProviderEvents.Error, async (eventDetails) => {
 
 ### Shutdown:
 
-The OpenFeature API provides a close function to perform a cleanup of all providers attached to any client.
+The OpenFeature API provides a close function to perform a cleanup of all registered providers.
+This should only be called when your application is in the process of shutting down.
```
beeme1mr, line 225: @toddbaert @lukas-reining I updated the text here so that we could release a patch version. It's a bit of a workaround but I think the additional context is valuable. What do you think?

toddbaert (approved):

toddbaert (approved):

PR #461, opened 2023-06-23
fix: events issues

toddbaert:
While doing the Java implementation, I found some issues I think we have in the JS implementation.

- we remove all handlers from the old provider when a new one is set - but other clients could still be listening. We need to remove only some.
- when we run handlers immediately in the situation a client is already READY, we don't wrap the handler in a try/catch and run it async.
- for each new client name we add an entry to a map (`_clientEvents`) even if no handlers are added. If an author dynamically generated clients with unique names (perhaps per request) this would constitute a memory leak.


toddbaert:
@lukas-reining feel free to do this if you want (self-assign) or tell me if I'm mistaken about anything. Otherwise I will tackle it sometime next week!

lukas-reining:
@toddbaert I will have a look tomorrow and would take this🙂

toddbaert:
> @toddbaert I will have a look tomorrow and would take thisslightly_smiling_face

OK let me know if you need any help!

lukas-reining (commented):
@toddbaert I started to fix the issues. Would be great if you have a look at changes to fix the first two issues.

- [x] - we remove all handlers from the old provider when a new one is set - but other clients could still be listening. We need to remove only some.
- [x] - when we run handlers immediately in the situation a client is already READY, we don't wrap the handler in a try/catch and run it async.
- [ ] - for each new client name we add an entry to a map (`_clientEvents`) even if no handlers are added. If an author dynamically generated clients with unique names (perhaps per request) this would constitute a memory leak.

For the last issue I am not too sure tbh @toddbaert. I totally see the symptom of the problem that you are describing. On the other hand it feels to me that as an application author you should be aware that getting a client could lead to resources being allocated and that there is no "dispose" function for a client.

We currently have the problem for these three collections in the OpenFeatureCommonAPI:
-  _clientEventHandlers
- _clientProviders
- _clientEvents

For the _clientEvents we may be able to use a week map or at least initialize it lazy.
But for _clientProviders and _clientEvents this is still the same memory leak. An application author can at any point ask for a client of a given name, even if the "last one with that name" has been garbage collected long ago. In this case we still expect the long time ago assigned provider to be used, so as long as there is no way of telling the OpenFeature API that a client name can be disposed, I do not see a solution to remove that memory leak.

To me this indicates that we might either have to have function to tell the API to "forget about" a client with a given name e.g.`OpenFeatureAPI.disposeClient(name?: string)`, or that we have to accept that memory leak and assume that application authors are aware of getting a clients means creating state.
What do you think? Any other idea?🙂

toddbaert:
The first issues seem to be fixed (though the code here, as in Go and Java is pretty tricky :sweat_smile: I don't think there's much more that can be done to help that)

> On the other hand it feels to me that as an application author you should be aware that getting a client could lead to resources being allocated and that there is no "dispose" function for a client.

I agree.

> For the _clientEvents we may be able to use a week map or at least initialize it lazy.
But for _clientProviders and _clientEvents this is still the same memory leak. An application author can at any point ask for a client of a given name, even if the "last one with that name" has been garbage collected long ago. In this case we still expect the long time ago assigned provider to be used, so as long as there is no way of telling the OpenFeature API that a client name can be disposed, I do not see a solution to remove that memory leak.

You're absolutely right about this. I misread the code. _There's no issue with creating a new map entry with a new provider_ - we wouldn't expect authors to dynamically register a multitude of those anyway. The real problem would be what I mentioned when I said:

> for each new client name we add an entry to a map (_clientEvents) even if no handlers are added

It's more realistic for an author to create 10000 dynamically named clients (even with no specific provider). In this case, we shouldn't be adding to maps. After re-reading the code, I don't think this was happening before - but I think it is now with your recent changes to fix the other issues (calling `events()` in the Client constructors).

toddbaert:
> > Does this make sense? I think if we do this lazily, at least only when we add handlers, that's good enough... then we are creating handlers which take up memory anyway, so using more memory would be quite expected.
> 
> Okay so you are saying we should do this, because this kind of helps against the problem of getting too many event handlers, but we are fine with many event handlers if every clients gets at least one event handler? 🙂

Yes, because if you are adding handlers, your already adding a resource that requires some specific cleanup beyond what the GC will do for you (the handler) and when all handlers are removed we can even remove the map entry.

lukas-reining:
> > > Does this make sense? I think if we do this lazily, at least only when we add handlers, that's good enough... then we are creating handlers which take up memory anyway, so using more memory would be quite expected.
> > 
> > 
> > Okay so you are saying we should do this, because this kind of helps against the problem of getting too many event handlers, but we are fine with many event handlers if every clients gets at least one event handler? 🙂
> 
> Yes, because if you are adding handlers, you're already adding a resource that requires some specific cleanup beyond what the GC will do for you (the handler) and when all handlers are removed we can even remove the map entry.

Okay great, I am with you here. I would propose trying to use a weak map here, that could be the cleanest and easiest way.

toddbaert:
> > > > Does this make sense? I think if we do this lazily, at least only when we add handlers, that's good enough... then we are creating handlers which take up memory anyway, so using more memory would be quite expected.
> > > 
> > > 
> > > Okay so you are saying we should do this, because this kind of helps against the problem of getting too many event handlers, but we are fine with many event handlers if every clients gets at least one event handler? slightly_smiling_face
> > 
> > 
> > Yes, because if you are adding handlers, you're already adding a resource that requires some specific cleanup beyond what the GC will do for you (the handler) and when all handlers are removed we can even remove the map entry.
> 
> Okay great, I am with you here. I would propose trying to use a weak map here, that could be the cleanest and easiest way.

It could be as easy as only creating the emitter when a handler is actually finally added (moving the call to the handler supplier function out of the constructor) but a weakmap might be a possible solution too.

lukas-reining:
> It could be as easy as only creating the emitter when a handler is actually added (moving the call to the handler supplier function call out of the constructor) but a weakmap might be a possible solution too.

Yeah thats true. Just pushed the change to do this @toddbaert.
Weak map is not even trivially possible because it could be that the provider is even set before the client which means that we would have to have the client as key in the weak map, that does not even exist yet.

I think the lazy init is good enough.

toddbaert:
@lukas-reining besides the small issues @beeme1mr found this looks good to me now. Since I actually opened the PR, you will have to approve it :smile: . If you do, I will squash all the changes into a single commit (by you, since you fixed it) and then merge.

lukas-reining (approved):
Great thanks @toddbaert!
And good that you have seen the console.log @beeme1mr 😅

lukas-reining (approved):
Great thanks @toddbaert!
And good that you have seen the console.log @beeme1mr 😅

beeme1mr (approved):

PR #471, opened 2023-06-29
chore: import fixes to fix e2e tests

toddbaert:
Both the server and client e2e tests run fine in the CI, but in some local scenarios the "deep imports" we used were causing issues with module resolution.

I've updated the e2e test to always import from the "top level" instead of the /src sub-directory in the test suite.

### DIFF
--- a/packages/server/e2e/step-definitions/setup.ts
+++ b/packages/server/e2e/step-definitions/setup.ts
@@ -1,6 +1,5 @@
-
 import assert from 'assert';
-import { OpenFeature } from '../../src';
+import { OpenFeature } from '../..';
 import { FlagdProvider } from '@openfeature/flagd-provider';
 
 const FLAGD_NAME = 'flagd Provider';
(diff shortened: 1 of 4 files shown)


packages/server/e2e/step-definitions/setup.ts
```diff
@@ -1,12 +1,14 @@
...
 const FLAGD_NAME = 'flagd Provider';
 
 // register the flagd provider before the tests.
 console.log('Setting flagd provider...');
 OpenFeature.setProvider(new FlagdProvider({ cache: 'disabled' }));
-assert(OpenFeature.providerMetadata.name === FLAGD_NAME, new Error(`Expected ${FLAGD_NAME} provider to be configured, instead got: ${OpenFeature.providerMetadata.name}`));
+assert(
+  OpenFeature.providerMetadata.name === FLAGD_NAME,
+  new Error(`Expected ${FLAGD_NAME} provider to be configured, instead got: ${OpenFeature.providerMetadata.name}`)
+);
```
toddbaert, line 13: just formatting here

toddbaert:
@jamescarr I'm not 100% sure this will fix your issue locally (I wasn't able to reproduce it) but it might.

beeme1mr (approved):

PR #472, opened 2023-06-29
fix: named client events

toddbaert:
* fixed issues where named client wouldnt get events from a default provider

### DIFF
--- a/packages/client/test/events.spec.ts
+++ b/packages/client/test/events.spec.ts
@@ -198,7 +198,11 @@ describe('Events', () => {
     });
 
     it('anonymous provider with named client should run', (done) => {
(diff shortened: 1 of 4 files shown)


toddbaert:
@lukas-reining I think this is a bug. I found it doing some integration testing. The new test failed without my change. I think what's happened is the transfers of handlers fails when the client name is null, but we still need to transfer the "unbound" emitters when the default provider is changed.

Please take a look and make sure the solution makes sense to you.

lukas-reining (commented):
Oh yes that is a problem.
Your solution looks good and makes sense.
Just refactored a bit and used `getAndCacheEventEmitterForClient` when getting the emitters.
What do you think @toddbaert?

lukas-reining (approved):

toddbaert:
@lukas-reining that makes sense, and I very much like your refactors!

PR #480, opened 2023-07-05
fix: events on anon provider/client

toddbaert:
* fix an issue where events would fail to run on the default provider anonymous client if the provider was changed

### DIFF
--- a/packages/shared/src/open-feature.ts
+++ b/packages/shared/src/open-feature.ts
@@ -191,7 +191,12 @@ export abstract class OpenFeatureCommonAPI<P extends CommonProvider = CommonProv
     const namedProviders = [...this._clientProviders.keys()];
     const eventEmitterNames = [...this._clientEvents.keys()].filter(isDefined);
     const unboundEmitterNames = eventEmitterNames.filter((name) => !namedProviders.includes(name));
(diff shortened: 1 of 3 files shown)


packages/shared/src/open-feature.ts
```diff
@@ -191,7 +191,12 @@ export abstract class OpenFeatureCommonAPI<P extends CommonProvider = CommonProv
     const namedProviders = [...this._clientProviders.keys()];
     const eventEmitterNames = [...this._clientEvents.keys()].filter(isDefined);
     const unboundEmitterNames = eventEmitterNames.filter((name) => !namedProviders.includes(name));
-    return unboundEmitterNames.map((name) => this._clientEvents.get(name)).filter(isDefined);
+    return [
+      // all unbound, named emitters
+      ...unboundEmitterNames.map((name) => this._clientEvents.get(name)),
+      // the default emitter
+      this._clientEvents.get(undefined)
+    ].filter(isDefined);
```
toddbaert, line 199: When I did [this fix](https://github.com/open-feature/js-sdk/pull/472), I missed the emitter for the default provider, which has no name.

lukas-reining (approved):
Looks good to me! 🙂

PR #490, opened 2023-07-20
feat: typesafe event emitter

### LINKED ISSUE #489: web-sdk: clientName in EventDetails is overriden
### Observed behavior
When writing a provider for the web-sdk, you can emit events and it is possible to provide `EventDetails` with the event.

```javascript
const events = new OpenFeatureEventEmitter();
events.emit(ProviderEvents.Error, {
    clientName: 'my-client-name',
    message: 'your message here'
});
```

When reading the event, from the handler we can see that the value of the client name is not what was provided in the event.

```javascript
await OpenFeature.setContext(defaultContext);
OpenFeature.setProvider('test-provider', defaultProvider);
const client = await OpenFeature.getClient('test-provider');
client.addHandler(ProviderEvents.Error, (event) => console.log(event));

// Result:
//{
//  clientName: 'test-provider',
//  message: 'Request failed with status code 404'
//}
```
The value of `clientName` is the client's name set in `setProvider` or `undefined` if we have no name for the provider.

### Expected Behavior
Not sure exactly how it should work here, but I see 2 potential solutions:
1. Don't expose `clientName` to providers to avoid confusion for people writing providers.
2. Use the `clientName` provided in the `emit` function and forward it in the event.

beeme1mr:
## This PR

- adds type safety to the event emitter

### Related Issues

Fixes #489 

### Notes

The event emitter was split into two classes. `OpenFeatureEventEmitter` is meant to be used by developers working on OpenFeature. `InternalEventEmitter` is meant for use within the SDK. Unfortunately, the internal emitter is currently being exposed publically. This is because both the client and server SDKs expose the entire shared lib.

### DIFF
--- a/packages/shared/src/events/eventing.ts
+++ b/packages/shared/src/events/eventing.ts
@@ -4,14 +4,34 @@ export type EventMetadata = {
   [key: string]: string | boolean | number;
 };
 
-export type EventDetails = {
+export type CommonEventDetails = {
   clientName?: string;
+};
+
+type CommonEventProps = {
   message?: string;
-  flagsChanged?: string[];
   metadata?: EventMetadata;
 };
 
-export type EventHandler = (eventDetails?: EventDetails) => Promise<unknown> | unknown;
+export type ReadyEvent = CommonEventProps;
+export type ErrorEvent = CommonEventProps;
+export type StaleEvent = CommonEventProps;
+export type ConfigChangeEvent = CommonEventProps & { flagsChanged?: string[] };
+
+type EventMap = {
+  [ProviderEvents.Ready]: ReadyEvent;
+  [ProviderEvents.Error]: ErrorEvent;
+  [ProviderEvents.Stale]: StaleEvent;
+  [ProviderEvents.ConfigurationChanged]: ConfigChangeEvent;
+};
+
+export type EventContext<
+  T extends ProviderEvents,
+  U extends Record<string, unknown> = Record<string, unknown>
+> = EventMap[T] & U;
+
+export type EventDetails<T extends ProviderEvents> = EventContext<T> & CommonEventDetails;
+export type EventHandler<T extends ProviderEvents> = (eventDetails?: EventDetails<T>) => Promise<unknown> | unknown;
 
 export interface Eventing {
   /**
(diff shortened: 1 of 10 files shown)


thomaspoignant (commented):
Looks good to me, I like the fact that we have typed objects for each event type it makes it explicit when to use optional fields such as `flagsChanged`.

thomaspoignant (approved):

packages/shared/src/events/eventing.ts
```diff
@@ -4,14 +4,34 @@ export type EventMetadata = {
   [key: string]: string | boolean | number;
 };
 
-export type EventDetails = {
+export type CommonEventDetails = {
```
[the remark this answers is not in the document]
toddbaert: I think this is fine.

packages/shared/src/events/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './events';
 export * from './eventing';
+// TODO restrict access to the internal event emitter while still being able to use it in the monorepo.
```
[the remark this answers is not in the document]
beeme1mr: Yeah, I'm going to try and address this now. If it can't easily be done, I'll just add a comment above the internal class that describes the intent.

packages/shared/src/open-feature.ts
```diff
@@ -23,10 +27,10 @@ export abstract class OpenFeatureCommonAPI<P extends CommonProvider = CommonProv
 
   protected abstract _defaultProvider: P;
 
-  private readonly _events = new OpenFeatureEventEmitter(() => this._logger);
-  private readonly _clientEventHandlers: Map<string | undefined, [ProviderEvents, EventHandler][]> = new Map();
+  private readonly _events = new InternalEventEmitter(() => this._logger);
+  private readonly _clientEventHandlers: Map<string | undefined, [ProviderEvents, EventHandler<any>][]> = new Map();
```
lukas-reining, line 31:
```suggestion
  private readonly _clientEventHandlers: Map<string | undefined, [ProviderEvents, EventHandler<ProviderEvents>][]> = new Map();
```

Here we can use ProviderEvents instead of any as we are never using narrower types here.
Thats not possible in the OpenFeatureEvenetEmitter, as we use the narrower types there ( https://github.com/open-feature/js-sdk/pull/490/files#diff-92bf71360d9fdb080cd82b4e1e5340f32ba26fe878ffc9529f87f32cf629931eR9)

beeme1mr: Good catch!

packages/shared/src/events/index.ts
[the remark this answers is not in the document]
beeme1mr: I commented on the event emitter classes and removed the TODO.

packages/shared/src/events/open-feature-event-emitter.ts
```diff
@@ -59,6 +62,22 @@ export class OpenFeatureEventEmitter implements ManageLogger<OpenFeatureEventEmi
...
+ * the result of the initialize method.
+ */
+export class OpenFeatureEventEmitter extends GenericEventEmitter {};
+
+/**
+ * The InternalEventEmitter should only be used within the SDK. It extends the
+ * OpenFeatureEventEmitter to include additional properties that can be included
+ * in the event details.
+ */
+export class InternalEventEmitter extends GenericEventEmitter<CommonEventDetails> {};
```
beeme1mr, line 83: Are these descriptions clear enough?

lukas-reining: I would say these are good!

lukas-reining (approved):

PR #491, opened 2023-07-21
fix: onContextChanged not running for named providers

### LINKED ISSUE #488: web-sdk: onContextChange not called for named provider
### Observed behavior
I am not sure to understand how do you change the context for a named provider?

Let's say I am doing something like this :
```javascript
await OpenFeature.setContext({targetingKey: 'foo'});
OpenFeature.setProvider('my-provider', provider);
const client = await OpenFeature.getClient('my-provider');
//...
await OpenFeature.setContext({targetingKey: 'bar'});
```

### Expected Behavior
I was expecting the function `onContextChange` of the provider to be called when changing the context.
But when looking at the code https://github.com/open-feature/js-sdk/blob/5725194393329cb65cf82a7a76c4ed2bd5d7bcf0/packages/client/src/open-feature.ts#L37-L41

 it seems that this function is called only for the default provider.

Is it something expected? Did I miss something to be sure that `onContextChange` is call on the provider.

lukas-reining:
## This PR

Fixes context propagation for the static-context-paradigm as found in #488.

### Related Issues

Fixes #488

### DIFF
--- a/packages/client/src/open-feature.ts
+++ b/packages/client/src/open-feature.ts
@@ -39,7 +39,15 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider> implements Ma
     this._context = context;
 
     const allProviders = [this._defaultProvider, ...this._clientProviders.values()];
-    await Promise.all(allProviders.map((provider) => provider.onContextChange?.(oldContext, context)));
+    await Promise.all(
+      allProviders.map(async (provider) => {
+        try {
+          return await provider.onContextChange?.(oldContext, context);
+        } catch (err) {
+          this._logger?.error(`Error running context change handler of provider ${provider.metadata.name}:`, err);
+        }
+      })
+    );
   }
 
   getContext(): EvaluationContext {
--- a/packages/client/src/open-feature.ts
+++ b/packages/client/src/open-feature.ts
@@ -37,7 +37,9 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider> implements Ma
   async setContext(context: EvaluationContext): Promise<void> {
     const oldContext = this._context;
     this._context = context;
-    await this._defaultProvider?.onContextChange?.(oldContext, context);
+
+    const allProviders = [this._defaultProvider, ...this._clientProviders.values()];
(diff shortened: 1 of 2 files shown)


thomaspoignant (approved):
Nice and simple 👍

lukas-reining:
The onContextChanged is now called for all providers.
I am not sure if some requirement is still missing for the new spec @toddbaert @beeme1mr?

I think what we still have to do is check the numbering of all tests as we mostly used requirement names for the describe blocks, and the number shifted a bit.

packages/client/src/open-feature.ts
```diff
@@ -37,7 +37,9 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider> implements Ma
   async setContext(context: EvaluationContext): Promise<void> {
     const oldContext = this._context;
     this._context = context;
-    await this._defaultProvider?.onContextChange?.(oldContext, context);
+
+    const allProviders = [this._defaultProvider, ...this._clientProviders.values()];
+    await Promise.all(allProviders.map((provider) => provider.onContextChange?.(oldContext, context)));
```
beeme1mr, line 42: Promise.all would fail fast if any of the `onContextChange` methods threw. I think we should catch and log an error or use Promise.allSettled.

beeme1mr:
I've created an issue to update the review and update test names to match the updated spec.

https://github.com/open-feature/js-sdk/issues/492

packages/client/src/open-feature.ts
toddbaert: I agree. One provider's error should not impact others. We may need to mention this in the spec.

lukas-reining: Good point, will add it. That was actually my question in the issue.

lukas-reining: @beeme1mr @toddbaert Fixed it now

beeme1mr (approved):

toddbaert (approved):
lgtm

toddbaert:
@lukas-reining I'm not sure this is a feat, I think it was more of a fix, but I'm not 100% sure it matters TBH, we are still sub 1.0/experimental.

Though TBH, we should probably publish our first non-experimental soon.

PR #501, opened 2023-07-22
fix: race adding handler during init

toddbaert:
Fixes a race condition that prevents handlers added while their provider is initializing from running.

### DIFF
--- a/packages/server/test/events.spec.ts
+++ b/packages/server/test/events.spec.ts
@@ -12,17 +12,21 @@ import {
 } from '../src';
 import { v4 as uuid } from 'uuid';
 
+const TIMEOUT = 1000;
+
 class MockProvider implements Provider {
   readonly metadata: ProviderMetadata;
   readonly events?: OpenFeatureEventEmitter;
   private hasInitialize: boolean;
   private failOnInit: boolean;
+  private initDelay?: number;
   private enableEvents: boolean;
   status?: ProviderStatus = undefined;
 
   constructor(options?: {
     hasInitialize?: boolean;
     initialStatus?: ProviderStatus;
+    initDelay?: number;
     enableEvents?: boolean;
     failOnInit?: boolean;
     name?: string;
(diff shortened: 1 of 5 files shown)


packages/server/test/events.spec.ts
```diff
@@ -299,6 +308,19 @@ describe('Events', () => {
...
+      // set the default provider
+      OpenFeature.setProvider(provider);
+      const client = OpenFeature.getClient();
+
+      // add a handler while the provider is starting
+      client.addHandler(ProviderEvents.Ready, () => {
+        done();
+      });
+    });
+
```
toddbaert, line 323: This test fails without the fix.

packages/shared/src/open-feature.ts
```diff
@@ -122,26 +122,26 @@ export abstract class OpenFeatureCommonAPI<P extends CommonProvider = CommonProv
...
-
+    const emitters = this.getAssociatedEventEmitters(clientName);
+    
     if (typeof provider.initialize === 'function') {
       provider
         .initialize?.(this._context)
         ?.then(() => {
-          emitters.forEach((emitter) => {
+          // fetch the most recent event emitters, some may have been added during init
+          this.getAssociatedEventEmitters(clientName).forEach((emitter) => {
```
toddbaert, line 132: This was the bug. We we're getting the event emitters before this block. Sometimes, after a provider had been set, but BEFORE the `initialize` was complete, a handler would be added. We wouldn't have it in the list to run (the `emitters` variable was stale).

The fix is to get the most up-to-date list of handlers in the `then`.

packages/server/test/events.spec.ts
```diff
@@ -299,6 +308,19 @@ describe('Events', () => {
...
```
toddbaert, line 323: This test fails without the fix.

packages/shared/src/open-feature.ts
lukas-reining: Ah good point!

lukas-reining (approved):
Looks good to me!

Issue #505, opened 2023-07-25
[BUG] client and server SDKs both initialize providers regardless of `status`.


toddbaert:
The SDKs currently initialize providers regardless of their state. This can mean providers could be initialized multiple times in some situations.

The spec could be a bit more strongly worded about this, but I think [1.1.2.2](https://openfeature.dev/specification/sections/flag-evaluation#requirement-1122) makes it clear (see also linked provider section).

I think we should fix this. We could also perhaps add some type guards to make this easier (for instance, force implementation of `status` if you also implement `OpenFeatureEventEmitter` since if you support events you probably need to define `initialize` and `status`.

Note both the [Java](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/ProviderRepository.java#L107-L108) and [Go](https://github.com/open-feature/go-sdk/blob/main/pkg/openfeature/api.go#L173) impls do not have this issue; they only initialize `NOT_READY` providers.

_I will happily take this issue on, I just want to me sure everyone agrees and is aware._

toddbaert:
I will happily take this issue on, I just want to me sure everyone agrees and is aware.

lukas-reining:
@toddbaert Yeah, thats true. 
If you dont find the time, I can also do it. If so, just reach out 🙂

toddbaert:
@lukas-reining I've fixed the issue, but we probably want to fix these before we release this:

- https://github.com/open-feature/js-sdk-contrib/issues/488
- https://github.com/open-feature/js-sdk-contrib/issues/489 
- https://github.com/open-feature/js-sdk-contrib/issues/490 for these.

lukas-reining:
Okay, I will fix this for config cat now. @toddbaert

PR #507, opened 2023-07-26
fix: only initialize NOT_READY providers

toddbaert:
See here. We should only run initialize if `ProviderStatus` is `NOT_READY` to prevent duplicate calls to `initialize`, as we  [Java](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/ProviderRepository.java#L107-L108) and [Go](https://github.com/open-feature/go-sdk/blob/main/pkg/openfeature/api.go#L173).

Most of this is test and formatting changes (I had to update the tests not to use the same object literal). The import change is the single line here.

Note: it's likely this will break _some_ providers that didn't implement `status`. I think the [flagd web provider](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/flagd-web/src/lib/flagd-web-provider.ts), [GoFF](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts) provider and the [config-cat](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/config-cat/src/lib/config-cat-provider.ts) could be impacted. I created issues for these:
- https://github.com/open-feature/js-sdk-contrib/issues/488
- https://github.com/open-feature/js-sdk-contrib/issues/489 
- https://github.com/open-feature/js-sdk-contrib/issues/490

Fixes: https://github.com/open-feature/js-sdk/issues/505

I've also removed all the `setTimeouts` in our test suite. Everything is working "properly" now with plain events, since all flagd providers are fully updated.

### DIFF
--- a/packages/client/e2e/step-definitions/setup.ts
+++ b/packages/client/e2e/step-definitions/setup.ts
@@ -6,14 +6,15 @@ const FLAGD_NAME = 'flagd-web';
 
 // register the flagd provider before the tests.
 console.log('Setting flagd web provider...');
-OpenFeature.setProvider(
-  new FlagdWebProvider({
-    host: 'localhost',
-    port: 8013,
-    tls: false,
-    maxRetries: -1,
-  })
-);
+const provider = new FlagdWebProvider({
+  host: 'localhost',
+  port: 8013,
+  tls: false,
+  maxRetries: -1,
+});
+// TODO: remove this when provider updated to properly use state.
+provider.initialize({});
+OpenFeature.setProvider(provider);
 assert(
   OpenFeature.providerMetadata.name === FLAGD_NAME,
   new Error(`Expected ${FLAGD_NAME} provider to be configured, instead got: ${OpenFeature.providerMetadata.name}`)
--- a/packages/client/e2e/step-definitions/setup.ts
+++ b/packages/client/e2e/step-definitions/setup.ts
@@ -2,21 +2,18 @@ import assert from 'assert';
 import { OpenFeature } from '../..';
 import { FlagdWebProvider } from '@openfeature/flagd-web-provider';
 
-const FLAGD_NAME = 'flagd-web';
+const FLAGD_WEB_NAME = 'flagd-web';
 
 // register the flagd provider before the tests.
 console.log('Setting flagd web provider...');
-const provider = new FlagdWebProvider({
+OpenFeature.setProvider(new FlagdWebProvider({
   host: 'localhost',
   port: 8013,
   tls: false,
   maxRetries: -1,
-});
-// TODO: remove this when provider updated to properly use state.
-provider.initialize({});
-OpenFeature.setProvider(provider);
+}));
 assert(
-  OpenFeature.providerMetadata.name === FLAGD_NAME,
-  new Error(`Expected ${FLAGD_NAME} provider to be configured, instead got: ${OpenFeature.providerMetadata.name}`)
+  OpenFeature.providerMetadata.name === FLAGD_WEB_NAME,
+  new Error(`Expected ${FLAGD_WEB_NAME} provider to be configured, instead got: ${OpenFeature.providerMetadata.name}`)
 );
 console.log('flagd web provider configured!');
--- a/packages/shared/src/open-feature.ts
+++ b/packages/shared/src/open-feature.ts
@@ -124,7 +124,7 @@ export abstract class OpenFeatureCommonAPI<P extends CommonProvider = CommonProv
 
     const emitters = this.getAssociatedEventEmitters(clientName);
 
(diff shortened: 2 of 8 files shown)


thomaspoignant (approved):

packages/shared/src/open-feature.ts
```diff
@@ -124,7 +124,15 @@ export abstract class OpenFeatureCommonAPI<P extends CommonProvider = CommonProv
...
     const emitters = this.getAssociatedEventEmitters(clientName);
 
-    if (typeof provider.initialize === 'function') {
+    // TODO: this can be removed after 1.0
+    if (typeof provider.initialize === 'function' && provider.status === undefined) {
+      const activeLogger = this._logger || console;
+      activeLogger.warn(
+        `Provider ${provider?.metadata?.name} implements 'initialize' but not 'status'. Please implement 'status'.`
+      );
+    }
```
toddbaert, line 133: This logs a warning if the provider doesn't implement state, but implements init.

packages/shared/src/open-feature.ts
```diff
@@ -124,7 +124,15 @@ export abstract class OpenFeatureCommonAPI<P extends CommonProvider = CommonProv
...
+
+    if (provider?.status === ProviderStatus.NOT_READY && typeof provider.initialize === 'function') {
```
toddbaert, line 135: This is the change.

thomaspoignant:
Hey, just wondering, do we already have a provider that implement the correct behavior?

packages/client/e2e/step-definitions/setup.ts
```diff
@@ -6,14 +6,15 @@ const FLAGD_NAME = 'flagd-web';
...
-    maxRetries: -1,
-  })
-);
+const provider = new FlagdWebProvider({
+  host: 'localhost',
+  port: 8013,
+  tls: false,
+  maxRetries: -1,
+});
+// TODO: remove this when provider updated to properly use state.
```
beeme1mr, line 15: Please create an issue for this so we don't forget about it 😄

packages/shared/src/open-feature.ts
```diff
@@ -124,7 +124,15 @@ export abstract class OpenFeatureCommonAPI<P extends CommonProvider = CommonProv
 
     const emitters = this.getAssociatedEventEmitters(clientName);
 
-    if (typeof provider.initialize === 'function') {
+    // TODO: this can be removed after 1.0
```
beeme1mr, line 127: Is there any harm in keeping this in post 1.0?

beeme1mr (approved):

packages/shared/src/open-feature.ts
toddbaert: I guess not really, except for that it's a bit messy.

toddbaert:
> Hey, just wondering, do we already have a provider that implement the correct behavior?

@thomaspoignant the launchdarkly provider does: https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts#L96-L99.

So does the in-progress implementation of flagd-js-server: https://github.com/open-feature/js-sdk-contrib/pull/484

packages/client/e2e/step-definitions/setup.ts
toddbaert: OK, ya. I there's a couple other minor change here too, we can get rid of some set-timeouts now that we have init.

toddbaert: https://github.com/open-feature/js-sdk/issues/518

packages/shared/src/open-feature.ts
toddbaert: I've removed this comment. I think that since people could be using pure JS, it's nice to keep warnings like this in the SDK. That something I tend to forget.

toddbaert, line 127:
```suggestion
    // warn of improper implementations
```

toddbaert:
This is fixed in all the providers we own which I know needed to be updated. 

I will merge this in the next few days and release the artifact as our first non-experimental version if that's fine with you all.

cc @thomaspoignant @beeme1mr @lukas-reining

packages/client/e2e/step-definitions/setup.ts
beeme1mr: Don't forget to remove this comment.

toddbaert: I've removed this comment _and_ removed all the `setTimeouts` in our test suite. Everything is working "properly" now with plain events, since all flagd providers are fully updated.

PR #534, opened 2023-08-07
docs: update web sdk readme

beeme1mr:
## This PR

- updates the web sdk readme to follow the new template

### DIFF
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -1,232 +1,296 @@
 <!-- markdownlint-disable MD033 -->
+<!-- x-hide-in-docs-start -->
 <p align="center">
   <picture>
-    <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/open-feature/community/0e23508c163a6a1ac8c0ced3e4bd78faafe627c7/assets/logo/horizontal/white/openfeature-horizontal-white.svg">
-    <source media="(prefers-color-scheme: light)" srcset="https://raw.githubusercontent.com/open-feature/community/0e23508c163a6a1ac8c0ced3e4bd78faafe627c7/assets/logo/horizontal/black/openfeature-horizontal-black.svg">
+    <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/open-feature/community/0e23508c163a6a1ac8c0ced3e4bd78faafe627c7/assets/logo/horizontal/white/openfeature-horizontal-white.svg" />
+    <source media="(prefers-color-scheme: light)" srcset="https://raw.githubusercontent.com/open-feature/community/0e23508c163a6a1ac8c0ced3e4bd78faafe627c7/assets/logo/horizontal/black/openfeature-horizontal-black.svg" />
     <img align="center" alt="OpenFeature Logo">
   </picture>
 </p>
 
-<h2 align="center">OpenFeature Web SDK</h2>
+<h2 align="center">OpenFeature JS Web SDK</h2>
 
-[![Project Status: WIP – Initial development is in progress, but there has not yet been a stable, usable release suitable for the public.](https://www.repostatus.org/badges/latest/wip.svg)](https://www.repostatus.org/#wip)
-[![npm version](https://badge.fury.io/js/@openfeature%2Fweb-sdk.svg)](https://www.npmjs.com/package/@openfeature/web-sdk)
-[![Specification](https://img.shields.io/static/v1?label=Specification&message=v0.6.0&color=yellow)](https://github.com/open-feature/spec/tree/v0.6.0)
+<!-- x-hide-in-docs-end -->
+<!-- The 'github-badges' class is used in the docs -->
+<p align="center" class="github-badges">
+  <a href="https://github.com/open-feature/spec/tree/v0.6.0">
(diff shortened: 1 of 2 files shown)


packages/client/README.md
```diff
@@ -1,232 +1,296 @@
...
-<h2 align="center">OpenFeature Web SDK</h2>
-
-[![Project Status: WIP – Initial development is in progress, but there has not yet been a stable, usable release suitable for the public.](https://www.repostatus.org/badges/latest/wip.svg)](https://www.repostatus.org/#wip)
-[![npm version](https://badge.fury.io/js/@openfeature%2Fweb-sdk.svg)](https://www.npmjs.com/package/@openfeature/web-sdk)
-[![Specification](https://img.shields.io/static/v1?label=Specification&message=v0.6.0&color=yellow)](https://github.com/open-feature/spec/tree/v0.6.0)
-
-## 👋 Hey there! Thanks for checking out the OpenFeature Web SDK
-
-### What is OpenFeature?
+<h2 align="center">OpenFeature JS Web SDK</h2>
```
beeme1mr, line 11: I went with "Web" over "Client-side" but I'm interested in others thoughts on this.

packages/client/README.md
```diff
@@ -1,232 +1,296 @@
...
-- support for various [providers](https://openfeature.dev/docs/reference/concepts/provider)
-- easy integration and extension via [hooks](https://openfeature.dev/docs/reference/concepts/hooks)
-- handle flags of any type: bool, string, numeric and object
-- [context-aware](https://openfeature.dev/docs/reference/concepts/evaluation-context) evaluation
+```ts
+import { OpenFeature } from '@openfeature/web-sdk';
 
-## 🚀 Usage:
+// Register your feature flag provider
+OpenFeature.setProvider(new YourProviderOfChoice());
```
beeme1mr, line 73: The current readme shows that `setProvider` returns a promise. However, that doesn't appear to be the case.

I would like this basic example to wait for initialization before performing flag evaluation. What's the recommended approach here? Is it possible to set context before initializing a new provider?

packages/client/README.md
```diff
@@ -1,232 +1,296 @@
...
 
-### Providers:
+Once you've added a provider as a dependency, it can be registered with OpenFeature like this:
 
-To develop a provider, you need to create a new project and include the OpenFeature SDK as a dependency. This can be a
-new repository or included in an existing contrib repository available under the OpenFeature organization. Finally,
-you’ll then need to write the provider itself. In most languages, this can be accomplished by implementing the provider
-interface exported by the OpenFeature SDK.
+```ts
+// TODO the current docs show this returning a promise
```
beeme1mr, line 112: This is a slight deviation from the exiting readme. This looks correct, but I would like to double-check before merging.

beeme1mr (commented):
While updating the web readme, I had a few questions that should be addressed in the docs.

1. Can you wait for a provider to be ready before evaluating a feature?
2. Can you set context before initializing a provider? If so, how?
3. Is `setProvider` synchronous or async in the static context paradigm?
4. What happens if a server-side provider is registered on the client (or vice versa)?
5. What happens if a server-side hook is registered on the client (or vice versa)?

I think we should update the SDK to throw if a hook or provider is registered on the "wrong" SDK. This would help provide subtle runtime errors by failing loudly and early if misconfiguration occurs.

FYI @thomaspoignant @lukas-reining @toddbaert

thomaspoignant:
> I think we should update the SDK to throw if a hook or provider is registered on the "wrong" SDK. This would help provide subtle runtime errors by failing loudly and early if misconfiguration occurs.

In typescript we will have no issue with that because the type will prevent you from doing it, but in JS we may have the problem, I agree this is a must to avoid confusion.

packages/client/README.md
lukas-reining: I think that is fine.
I mean it could be that the SDK is used in something like React Native but I do not think that this would lead to any problems in understanding.

packages/client/README.md
lukas-reining: > The current readme shows that setProvider returns a promise. However, that doesn't appear to be the case.

Yes, this not the case anymore. For both SDK, you get back the Api without waiting for initialization.

> Is it possible to set context before initializing a new provider?

Yes this can be set at any point. When `initialize` on the provider is called, the current context is given into it.

> I would like this basic example to wait for initialization before performing flag evaluation. What's the recommended approach here?

I think this would only be possible with a top level await, where a promise is resolved wen the ready handler is called. Without this, you can not really wait for the initialization.

lukas-reining (approved):
Looks good to me!

packages/client/README.md
toddbaert, line 73: We could add convenience methods as described [here](https://github.com/open-feature/ofep/blob/main/011-OFEP-sdk-wait-provider-ready.md), and document them here.

In fact, a PoC would be helpful, cc @thomaspoignant

packages/client/README.md
toddbaert: I think "web" is the clearest.

toddbaert (approved):

packages/client/README.md
beeme1mr: I'll update the readme once we have something available.

PR #537, opened 2023-08-12
feat: add provider compatibility check

### LINKED ISSUE #535: [FEATURE] warn an implementer if they are using a client-side hook in the server-side SDK, or vice versa
### Requirements

As suggested by @beeme1mr [here](https://github.com/open-feature/js-sdk/pull/534#pullrequestreview-1566071076), it would be really good if we could give a clear warning to an OF integrater who is accidently trying to use the server SDK in a browser context, or add a server-side provider to a client-side SDK, etc etc.

beeme1mr:
## This PR

- add provider compatibility check

### Related Issues

Fixes #535 

### Notes

This implementation throws if you set a provider that's incompatible with its intended use. This could easily be changed to a warning if that's the preferred approach. I think it's a good idea to fail early but I would be interested in others thoughts on this.

This is a non-breaking change. Providers can optionally specify their intended use and the SDK will perform a runtime check during provider registration.

### Follow up

- Add hook compatibility checks.
- Update readme to include examples.
- Update existing providers.

### DIFF
--- a/packages/shared/src/types/category.ts
+++ b/packages/shared/src/types/category.ts
@@ -1,1 +1,0 @@
-export type Category = 'server' | 'client';
--- a/packages/shared/src/types/category.ts
+++ b/packages/shared/src/types/category.ts
@@ -1,1 +1,1 @@
-export type Category = "server" | "client";
+export type Category = 'server' | 'client';
--- a/packages/shared/src/types/category.ts
+++ b/packages/shared/src/types/category.ts
@@ -1,0 +1,1 @@
+export type Category = "server" | "client";
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,4 +1,5 @@
 {
+  "packages/react": "0.0.1-experimental",
   "packages/client": "0.4.0",
   "packages/server": "1.4.2",
   "packages/shared": "0.0.10"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1476,9 +1476,9 @@
       }
     },
     "node_modules/@rollup/plugin-typescript": {
-      "version": "11.0.0",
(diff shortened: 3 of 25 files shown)


packages/shared/src/types/category.ts
```diff
@@ -0,0 +1 @@
+export type Category = "server" | "client";
```
beeme1mr, line 1: I really don't like the term "Category" here. It's what we used on the Ecosystem page, so that is what I used here. Perhaps something like "IntendedOn", "IntendedFor", "RunsOn", "CompatibleWith", or something similar would work better.

@moredip, any recommendations?

toddbaert:
I think this is a good idea, and I think throwing is fine.

packages/shared/src/types/category.ts
moredip: My vote is `Paradigm`!

beeme1mr: Do you think I should stick with the values of `server` and `client` if we go with `Paradigm`? In the spec, we use the term paradigm to describe static context vs dynamic context. I think the intent is more clear when you use the term `server` or `client` but I want to be careful mixing terminology.

@moredip @toddbaert @lukas-reining

toddbaert: I really don't mind, but I think if I had to pick one, I'd pick `RunsOn` and `client/server`.

beeme1mr: I went with a combination of the two. The provider has an optional property `runsOn` which expects a `Paradigm`.

toddbaert (approved):
I think this is a good idea. One thing to note is the React SDK currently re-exports the `/client` objects, so we get this "for free" there.

Issue #539, opened 2023-08-15
Updates to conform to latest spec


toddbaert:
With the possibility of a new [spec release](https://github.com/open-feature/spec/issues/202), we should have a PoC branch. The changes are fairly small but worth proving out before we release.

PR #541, opened 2023-08-17
feat: STALE state, minor event changes

toddbaert:
* added the "STALE" provider state
* added "providerName" to EventDetails
* run handlers immediately if provider in corresponding state
* fixes a few lint issues

Fixes: https://github.com/open-feature/js-sdk/issues/562
Fixes: https://github.com/open-feature/js-sdk/issues/539

### DIFF
--- a/packages/shared/src/provider/provider.ts
+++ b/packages/shared/src/provider/provider.ts
@@ -1,7 +1,6 @@
-import { OpenFeatureEventEmitter } from '../events';
-import { Metadata } from '../types';
 import { EvaluationContext } from '../evaluation';
-import { Paradigm } from '../types';
+import { OpenFeatureEventEmitter } from '../events';
+import { Metadata, Paradigm } from '../types';
 
 /**
  * The state of the provider.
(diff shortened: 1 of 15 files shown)


packages/shared/src/provider/provider.ts
```diff
@@ -20,6 +20,11 @@ export enum ProviderStatus {
    * The provider is in an error state and unable to evaluate flags.
    */
   ERROR = 'ERROR',
+
+  /**
+   * The provider's cached state is no longer valid and may not be up-to-date with the source of truth.
```
beeme1mr, line 25:
```suggestion
   * The provider's cached state may not be up-to-date with the source of truth.
```

I don't think it's accurate to say that it's no longer valid since we simply don't know.

toddbaert: I would say as soon as you're not sure a cache is valid, it's invalid. This is the language we use in the spec: https://github.com/open-feature/spec/blob/main/specification/types.md#provider-status

We can change it but I think if we do we need to start there.

beeme1mr: I'm fine with the language as is.

toddbaert:
Now that 0.7.0 is released marked this ready.

beeme1mr (approved):

PR #546, opened 2023-08-22
feat: react sdk groundwork

toddbaert:
This PR:

- adds all the necessary groundwork for a new react module in the monorepo (packaging, release, CI, etc)
- adds a _very_ limited implementation based on work by @moredip , see example in [README](https://github.com/open-feature/js-sdk/blob/a9104c1ca2a2fecd75b02e1d804a7bb984dd6d2d/packages/react/README.md)

Things to note:

- I've chosen NOT to bundle in the web-sdk directly into the react-sdk (which I could have done). Instead, the web-sdk is a peer dependency. Reasoning:
  - NPM 16+ automatically downloads peer deps if missing, and otherwise uses the available peer if compatible
  - Providers "don't care" whether they are used by the web-sdk, or the react-sdk, however, they all need a peerDependency on the web-sdk. If we bundled the web-sdk into the react-sdk automatically, it would mean that a consumer only using the react-sdk + some-web-provider would get a "useless" copy of the web-sdk downloaded automatically. For that reason, it's better to have the web-sdk as a peer, with the react SDK purely building on to of that dep, not duplicating/embedding a bunch of its code.
  - the web-sdk objects are re-exported from the react-sdk, so consumers only need to import from there.

relates to: https://github.com/open-feature/js-sdk/pull/546

:mega: I want to make clear that this is not a final API, and I'd love to continue discussion in the linked issue on how this could work. We can merge this PR (and subsequent ones) without releasing so that people can experiment locally with various APIs, via [yalc](https://github.com/wclr/yalc) or `npm link`.

### DIFF
--- a/packages/react/src/use-feature-flag.ts
+++ b/packages/react/src/use-feature-flag.ts
@@ -1,21 +1,11 @@
-import {
-  Client,
-  EvaluationDetails,
-  FlagValue,
-  ProviderEvents,
-  ResolutionDetails,
-} from '@openfeature/web-sdk'
-import { useEffect, useState } from 'react'
-import { useOpenFeatureClient } from './provider'
+import { Client, EvaluationDetails, FlagValue, ProviderEvents, ResolutionDetails } from '@openfeature/web-sdk';
+import { useEffect, useState } from 'react';
+import { useOpenFeatureClient } from './provider';
 
-export function useFeatureFlag<T extends FlagValue>(
-  flagKey: string,
-  defaultValue: T
-): T {
-  const [flagEvaluationDetails, setFlagEvaluationDetails] =
-    useState<ResolutionDetails<T> | null>(null)
+export function useFeatureFlag<T extends FlagValue>(flagKey: string, defaultValue: T): T {
+  const [flagEvaluationDetails, setFlagEvaluationDetails] = useState<ResolutionDetails<T> | null>(null);
 
-  const client = useOpenFeatureClient()
+  const client = useOpenFeatureClient();
 
   useEffect(() => {
     getFlag(client, flagKey, defaultValue, setFlagEvaluationDetails);
@@ -26,12 +16,12 @@ export function useFeatureFlag<T extends FlagValue>(
     client.addHandler(ProviderEvents.ConfigurationChanged, () => {
       getFlag(client, flagKey, defaultValue, setFlagEvaluationDetails);
     });
-  }, [flagKey, defaultValue, client])
+  }, [flagKey, defaultValue, client]);
 
   if (flagEvaluationDetails) {
-    return flagEvaluationDetails.value
+    return flagEvaluationDetails.value;
   } else {
-    return defaultValue
+    return defaultValue;
   }
 }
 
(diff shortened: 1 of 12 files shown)


packages/react/src/use-feature-flag.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export function useFeatureFlag<T extends FlagValue>(flagKey: string, defaultValue: T): T {
+  const [flagEvaluationDetails, setFlagEvaluationDetails] = useState<ResolutionDetails<T> | null>(null);
+
+  const client = useOpenFeatureClient();
+
+  useEffect(() => {
+    getFlag(client, flagKey, defaultValue, setFlagEvaluationDetails);
+    // adding handlers here mean that changes are immediately reflected in the UI when flags change
+    client.addHandler(ProviderEvents.Ready, () => {
```
moredip, line 13: I believe we also want to add `client.removeHandler` calls in the [cleanup method](https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/) that we return to `useEffect`

packages/react/src/use-feature-flag.ts
```diff
@@ -0,0 +1,47 @@
...
+      getFlag(client, flagKey, defaultValue, setFlagEvaluationDetails);
+    });
+    client.addHandler(ProviderEvents.ConfigurationChanged, () => {
+      getFlag(client, flagKey, defaultValue, setFlagEvaluationDetails);
+    });
```
kinyoklion, line 18: Feels like this should return a cleanup function that removes these handlers.

packages/react/src/use-feature-flag.ts
```diff
@@ -0,0 +1,47 @@
+import { Client, EvaluationDetails, FlagValue, ProviderEvents, ResolutionDetails } from '@openfeature/web-sdk';
+import { useEffect, useState } from 'react';
+import { useOpenFeatureClient } from './provider';
+
+export function useFeatureFlag<T extends FlagValue>(flagKey: string, defaultValue: T): T {
+  const [flagEvaluationDetails, setFlagEvaluationDetails] = useState<ResolutionDetails<T> | null>(null);
```
moredip, line 6: 🚨 this approach may go poorly 🚨

I took a very similar tack in an internal implementation  a `useFeatureFlag` at $DAY_JOB$, and it bit us pretty hard. Trying to maintain your own copy of the flag evaluation state separate from the OF client leads to some very confusing race conditions due to how React rendering and hooks work. 

It is MUCH better to simply call `getFlag` each time the hook is evaluated.

kinyoklion (approved):
Looks like a reasonable starting point to me.

packages/react/src/use-feature-flag.ts
toddbaert: That makes sense to me.

toddbaert:
I'll leave this open for at least 24 hrs to give people a chance to comment, but I'm inclined to merge after sooner than later since there's no intent to release this as is.

packages/react/src/use-feature-flag.ts
weyert: And maintain/manage any sort of caches within the OpenFeature provider?

packages/react/src/use-feature-flag.ts
toddbaert: absolutely.

packages/react/src/use-feature-flag.ts
toddbaert: [Great minds.](https://github.com/open-feature/js-sdk/pull/546#discussion_r1301935549)

packages/react/src/use-feature-flag.ts
toddbaert: > And maintain/manage any sort of caches within the OpenFeature provider?

I won't speak for @moredip , but that's how I took it... and I think it's consistent with all our paradigms and thinking so far. Maintaining state in 2 places is always a bad idea:tm:

weyert (approved):

packages/react/src/use-feature-flag.ts
lukas-reining: > It is MUCH better to simply call getFlag each time the hook is evaluated.

Ya, +1 on that.
Would we then just add something like a "stale" state that we set when ProviderEvents.Ready and ProviderEvents.ConfigurationChanged occur?  

Could be something like this sketch: 

```ts
 const [, updateFlag] = React.useState();
 const reevaluateFlag = React.useCallback(() => updateFlag({}), []);

  useEffect(() => {
    // adding handlers here means that changes are immediately reflected in the UI when flags change
    client.addHandler(ProviderEvents.Ready, reevaluateFlag);
    client.addHandler(ProviderEvents.ConfigurationChanged, reevaluateFlag);
    return () => {
      // be sure to cleanup the handlers
      client.removeHandler(ProviderEvents.Ready, reevaluateFlag);
      client.removeHandler(ProviderEvents.ConfigurationChanged, reevaluateFlag);
    };
  }, [flagKey, defaultValue, client]);

  return getFlag(client, flagKey, defaultValue, setFlagEvaluationDetails)

```

lukas-reining (approved):

Issue #545, opened 2023-08-04
React SDK implementation


toddbaert:
We should implement a React SDK. The SDK would provide a "React-friendly" API and features, built on top of the [Web SDK](https://github.com/open-feature/js-sdk/tree/main/packages/client).

:warning: Below is a _**tentative**_ list of features we should implement (some points include questions, _and all of these are up for debate_) :warning: 

- [x] a [context provider](https://react.dev/learn/passing-data-deeply-with-context) for initializing the SDK and allowing access to flag values across components
  - we may want to allow this context provider to be bound to a particular `client-name` to support multiple providers, provider-client mapping 
- [x] a `useStatus` [hook](https://react.dev/reference/react#your-own-hooks) which corresponds to associated provider status and last fired provider readiness event
  - useful for showing spinners/loaders while provider's init, if authors would like to wait before rendering content sensitive to flags
- [x] `useFlagValue/useFlagDetails` hooks, which can be used to impact rendering based on flag values
  - should this API be generic? Or should we do `useBooleanFlagValue`, etc?
  - should facilitate real-time, dynamic updates when providers emit `PROVIDER_CONFIGURATION_CHANGED` events

Non-functional requirements:

- consistent terminology and naming with Web SDK and OpenFeature spec
- observe React idioms and patterns
- helpful documentation (`useEffect` to update context, explanations for all the features/APIs above, configuration)
  - included in https://openfeature.dev/ecosystem/ page
- minimal implementation (only "react-ish") things should be implemented here, general enhancements should be made in the Web SDK if needed
- accurate peer dependency for the Web SDK
- helpful types

Precedents:

- featurevisor react SDK: https://featurevisor.com/docs/react/
- LaunchDarkly react SDK: https://docs.launchdarkly.com/sdk/client-side/react/react-web
- Spotify Confidence SDK: https://github.com/spotify/confidence-openfeature-provider-js/tree/main/packages/integration-react

toddbaert:
I've opened https://github.com/open-feature/js-sdk/pull/546, which mostly just lays the repo groundwork, but also has a basic implementation (NOT FINAL) by @moredip.

I've also transferred this issue to the js-sdk (out of contrib).

juanparadox:
Have we given any thought to adding some kind of `<OpenFeature>{some child}</OpenFeature>` component that takes in a feature flag boolean result and based on that result conditionally renders the `children`?

[...]

PR #549, opened 2023-08-24
feat: do not add feature flags to state

lukas-reining:
## This PR
Implements the change proposed [here](https://github.com/open-feature/js-sdk/pull/546#discussion_r1302922403) for the useFeatureFlag hook.

Instead of writing the flag to the hook state and updating it on events, the flag is just always evaluated. 
This prevents us from cache inconsistencies.

One thing I had to do was adding a forceUpdate which forces react to rerender and by this returns the new flag value when an event occurs.

### DIFF
--- a/packages/react/src/use-feature-flag.ts
+++ b/packages/react/src/use-feature-flag.ts
@@ -1,51 +1,36 @@
-import { Client, EvaluationDetails, FlagValue, ProviderEvents, ResolutionDetails } from '@openfeature/web-sdk';
+import { Client, EvaluationDetails, FlagValue, ProviderEvents } from '@openfeature/web-sdk';
 import { useEffect, useState } from 'react';
 import { useOpenFeatureClient } from './provider';
 
 export function useFeatureFlag<T extends FlagValue>(flagKey: string, defaultValue: T): T {
-  const [flagEvaluationDetails, setFlagEvaluationDetails] = useState<ResolutionDetails<T> | null>(null);
+  const [, setForceUpdateState] = useState({});
 
   const client = useOpenFeatureClient();
 
   useEffect(() => {
-    getFlag(client, flagKey, defaultValue, setFlagEvaluationDetails);
-    const handler = () => {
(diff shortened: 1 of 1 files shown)


packages/react/src/use-feature-flag.ts
```diff
@@ -1,51 +1,36 @@
-import { Client, EvaluationDetails, FlagValue, ProviderEvents, ResolutionDetails } from '@openfeature/web-sdk';
+import { Client, EvaluationDetails, FlagValue, ProviderEvents } from '@openfeature/web-sdk';
 import { useEffect, useState } from 'react';
 import { useOpenFeatureClient } from './provider';
 
 export function useFeatureFlag<T extends FlagValue>(flagKey: string, defaultValue: T): T {
-  const [flagEvaluationDetails, setFlagEvaluationDetails] = useState<ResolutionDetails<T> | null>(null);
+  const [, setForceUpdateState] = useState({});
```
toddbaert, line 6:
```suggestion
  const [setForceUpdateState] = useState({});
```

toddbaert (approved):

toddbaert:
> One thing I would also propose to change is that instead of just returning the flag value, we could return the EvaluationDetails which would also fit nicely to @moredip's [proposal](https://github.com/open-feature/js-sdk/issues/545#issuecomment-1688504758) for having the destructuring going and maybe later also adding other information if needed. But maybe we handle that in another issue?
> 
> ```ts
> const { value } = useFeatureFlag('feature-enabled', false);
> ```

I like this idea.

packages/react/src/use-feature-flag.ts
lukas-reining: Mh this does not work isn't it?
It would make the state be setState if I am not totally overseeing something @toddbaert🤔

toddbaert: Oh, I'm sorry, I misread. I though this was a typo, but I see it's an array de-structure now.

toddbaert:
I'll merge this tomorrow unless I hear objections, since it's not really an API change.

PR #555, opened 2023-08-29
feat(react-sdk): return flag evaluation details from hook

lukas-reining:
## This PR

This proposes to return the whole feature flag evaluation details from the react `useFeatureFlag` hook.

Only thing i am not sure if we should return an alias type for EvaluationDetails or as is is in this PR, the EvaluationDetails type.  Using an alias maybe like `type FeatureFlag<T> = EvaluationDetails<T>` would make clear that this may be extended later, even though we can still do it without having an alias now.

The hook becomes nicely usable as seen in the following.

```ts
const { value } = useFeatureFlag('feature-enabled', false);
```


beeme1mr (approved):

toddbaert (approved):
I like this.

Personally, I want to keep as many abstractions consistent as possible with the React SDK, so I would rather not use the alias... but I could be convinced otherwise.

PR #578, opened 2023-10-09
feat: add support for clearing providers

### LINKED ISSUE #526: Add utility function to clear providers for testing purposes
### Requirements

When writing tests we can observe some weird behavior because there is no way to unlink providers from the `Openfeature` singleton.

It could be great to have a way to clear the providers from the singleton to be able to create new tests that are not dependent on the previous situation of the singleton.
I am thinking of something similar to [`clearHooks`](https://github.com/open-feature/js-sdk/blob/95332ab5b2cc523b0ec7104d7e72c8a693887209/packages/shared/src/open-feature.ts#L45).

beeme1mr:
## This PR

- adds support for clearing providers

### Related Issues

Fixes #526 

### Notes

We may want to consider adding a reset method that calls `clearHooks`, `clearProviders`, rests the logger, and clears event listeners. There are a few tests that use a hacky method to do this already.

### DIFF
--- a/packages/client/src/open-feature.ts
+++ b/packages/client/src/open-feature.ts
@@ -75,6 +75,10 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider> implements Ma
       { name, version }
     );
   }
+
+  clearProviders() {
+    return super.clearProviders(NOOP_PROVIDER);
+  }
 }
 
 /**
--- a/packages/client/src/open-feature.ts
+++ b/packages/client/src/open-feature.ts
@@ -76,8 +76,12 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider> implements Ma
     );
   }
 
-  clearProviders() {
-    return super.clearProviders(NOOP_PROVIDER);
+  /**
+   * Clears all registered providers and resets the default provider.
+   * @returns {Promise<void>}
+   */
+  clearProviders(): Promise<void> {
+    return super.clearProvidersAndSetDefault(NOOP_PROVIDER);
   }
 }
 
--- a/packages/client/test/client.spec.ts
+++ b/packages/client/test/client.spec.ts
@@ -91,11 +91,12 @@ const MOCK_PROVIDER: Provider = {
 };
 
 describe('OpenFeatureClient', () => {
-  beforeAll(() => {
+  beforeEach(() => {
     OpenFeature.setProvider(MOCK_PROVIDER);
   });
 
-  afterEach(() => {
+  afterEach(async () => {
+    await OpenFeature.clearProviders();
     jest.clearAllMocks();
   });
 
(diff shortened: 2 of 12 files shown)


packages/client/src/open-feature.ts
```diff
@@ -75,6 +75,10 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider> implements Ma
       { name, version }
     );
   }
+
+  clearProviders() {
```
beeme1mr, line 79: JSDocs are enhanced from the abstract class.

packages/client/test/client.spec.ts
```diff
@@ -307,164 +308,165 @@ describe('OpenFeatureClient', () => {
       });
     });
   });
-});
-
-describe('Requirement 1.4.3.1', () => {
```
beeme1mr, line 312: These tests were moved inside the `OpenFeatureClient` test suite.

packages/shared/src/open-feature.ts
```diff
@@ -290,12 +290,26 @@ export abstract class OpenFeatureCommonAPI<P extends CommonProvider = CommonProv
         try {
           await provider.onClose?.();
         } catch (err) {
-          this.handleShutdownError(this._defaultProvider, err);
+          this.handleShutdownError(provider, err);
```
beeme1mr, line 293: This addresses a small issue where the wrong provider was being logged when an error occurred.

packages/client/src/open-feature.ts
lukas-reining: Are they even though abstract and concrete `clearProviders` they have different params?

packages/server/src/open-feature.ts
```diff
@@ -104,6 +103,10 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider> implements Ma
       context
     );
   }
+
+  clearProviders() {
+    return super.clearProviders(NOOP_PROVIDER);
+  }
```
lukas-reining, line 109: Is it a good idea to override this method with different params?
To me this seems to break Liskovs Substitution Principle. One could not use this method in a context where a OpenFeatureCommonApi is assumed I would say.
But for some reason Typescript seems to be fine with it, maybe because it is protected. 
It still feels a bit odd, but i am not sure.

packages/client/src/open-feature.ts
beeme1mr: Hmm, good point. Let me double-check.

packages/server/src/open-feature.ts
beeme1mr: I'll try and clean this up. I agree it's a bit odd.

packages/client/test/client.spec.ts
```diff
@@ -91,11 +91,12 @@ const MOCK_PROVIDER: Provider = {
 };
 
 describe('OpenFeatureClient', () => {
-  beforeAll(() => {
+  beforeEach(() => {
```
toddbaert, line 94: Is there a good reason for doing this each test?

Could we instead only do the `await OpenFeature.clearProviders();` in an afterAll?

packages/server/src/open-feature.ts
toddbaert: Agree this violates LSP.

packages/client/test/client.spec.ts
beeme1mr: It's cleaner to reset the providers after each test, in my opinion. Also, some of the new tests change the provider and must be reset at some point. I could handle this another way if you prefer but I think a clean reset per test is worth the impact on testing performance.

packages/client/src/open-feature.ts
beeme1mr: I've added a JSDoc to both the client and server because the method name on the abstract class has changed. That should resolve this concern and the LSP violation.

beeme1mr:
@toddbaert @lukas-reining I've addressed the concerns you identified. Please take a look when you have a moment.

lukas-reining (approved):
Looks good to me! :)

Issue #565, opened 2023-09-13
[FEATURE] In-memory provider for js (node) and web SDKs


toddbaert:
### Requirements

As outlined [in the specification](https://openfeature.dev/specification/appendix-a#in-memory-provider), the js-sdk should include an in-memory provider implementation. This implementation should take a simple object that defines flags as well as simple rules in the form of lambdas (see equivalent [Java implementation](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java), and existing [JS implementation](https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers/in-memory) in contrib) and use those to evaluate flags. This is very useful to conusmers of the SDK for testing purposes. Please carefully implement all the behaviors defined in the spec for the in-memory provider.

This is a great issue to take on if you're looking to get started in OpenFeature!

@moredip This would be pretty similar to what you implemented in the JS contrib. I think the main thing missing would be rules-based evaluation (some flag "values" would instead define lambdas to run which are passed the context). If you'd like to work on this issue, just assign it to yourself. Otherwise, I will add the `good first issue`, etc tags to it which I suspect will result in it quickly being picked up.
```[tasklist]
### Tasks
- [ ] web-sdk implementation (`packages/client`)
- [x] js-sdk (node) implementation (`packages/server`)
- [ ] remove existing partial implementation from js-sdk-contrib
```

luizgribeiro:
I was looking at both js-contrib inMemoryProvider and the Java Provider mentioned and the structure looks pretty different. The JS one uses  object literals to assign values to flags while Java's goes deeper with variants and context evaluator. This part caught my attention from the js:

```ts
  async resolveObjectEvaluation<U extends JsonValue>(flagKey: string): Promise<ResolutionDetails<U>> {
    throw new GeneralError('support for object flags has not been implemented');
  }
```
It is not clear to me if the context aware implementation of js-sdk should be backwards compatible or not.
In terms of docs, content and also getting started with openFeature it would be easier if it was.

What do you think that would be the best approach in this case @toddbaert ?

toddbaert:
I don't think we have _too many_ docs specifically talking about the current in-memory provider (and if we do have blogs/docs we can update them). I think the most important thing is to have a provider that conforms to the appendix in the spec.and is robust-enough for use in testing, which is the main purpose of the in-memory provider (replace your "real" provider with in-memory in unit testing).

[...]

PR #585, opened 2023-10-09
feat(server): add in memory provider

luizgribeiro:
## This PR

Implements the following features for `inMemoryProvider`:
- default value for flags
- reason for flag evaluation
- Context based evaluation

### Related Issues

It's part of #565 

### Follow-up Tasks

- Update docs

### How to test

Unit tests were also implemented to assure that it is working


toddbaert:
Hey @luizgribeiro I think the only thing we need here is the provider code. You can remove the `CHANGELOG.md`, `README.md`, and all the other files (like the tsconfig, etc). Basically just add the actual provider files into /packages/server and we should be good. You can also add some of the doc from `README.md` to the server `README.md`.

There's also a couple lint errors, and some imports are broken (you will need to use relative imports for the stuff from the server package, or import from `@openfeature/shared` for things common to both the web/server SDKs).

packages/server/README.md
```diff
@@ -186,17 +186,28 @@ A name is a logical identifier which can be used to associate clients with a par
 If a name has no associated provider, the global provider is used.
 
 ```ts
-import { OpenFeature } from "@openfeature/js-sdk";
+import { OpenFeature, InMemoryProvider } from "@openfeature/js-sdk";
```
toddbaert, line 189: I've made this look pretty much identical to the Java readme.

beeme1mr: We should either release this right after it's merged or wait to update the docs until after the feature is live.

packages/server/src/provider/in-memory-provider/flag-configuration.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Don't export types from this file publicly.
+ * It might cause confusion since these types are not a part of the general API,
+ * but just for the in-memory provider.
```
beeme1mr, line 4: Don't we `export * from '@openfeature/shared` already in the SDK?

packages/server/src/provider/in-memory-provider/flag-configuration.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+import { EvaluationContext, JsonValue } from '@openfeature/shared';
+
+type Variants<T> = Record<string, T>;
+
+export type Flag = {
+  variants: Variants<boolean> | Variants<string> | Variants<number> | Variants<JsonValue>;
+  defaultVariant: string;
+  disabled: boolean;
+  contextEvaluator?: (ctx: EvaluationContext) => string;
```
beeme1mr, line 14: Could you include some JSDocs above each property?

packages/server/src/provider/in-memory-provider/flag-configuration.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Don't export types from this file publicly.
+ * It might cause confusion since these types are not a part of the general API,
+ * but just for the in-memory provider.
+ */
+import { EvaluationContext, JsonValue } from '@openfeature/shared';
+
+type Variants<T> = Record<string, T>;
+
+export type Flag = {
```
beeme1mr, line 10: Could this live in shared or will it be slightly different on the client?

packages/server/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -0,0 +1,152 @@
...
+  ProviderEvents,
+  ResolutionDetails,
+  StandardResolutionReasons,
+  TypeMismatchError,
+} from '@openfeature/shared';
+import { FlagConfiguration, Flag } from './flag-configuration';
+import { Provider } from '../provider';
+
+/**
+ * A simple OpenFeature provider, intended for simple demos and as a test stub.
```
beeme1mr, line 19:
```suggestion
 * A simple OpenFeature provider intended for demos and as a test stub.
```

packages/server/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -0,0 +1,152 @@
...
+      let errorCode;
+      switch (true) {
+        case error instanceof TypeMismatchError:
+          errorCode = ErrorCode.TYPE_MISMATCH;
+          break;
+        case error instanceof FlagNotFoundError:
+          errorCode = ErrorCode.FLAG_NOT_FOUND;
+          break;
+        case error instanceof GeneralError:
+          errorCode = ErrorCode.PARSE_ERROR;
```
beeme1mr, line 106: Why are you mapping a general error to a parse error?

beeme1mr (approved):

packages/server/src/provider/in-memory-provider/flag-configuration.ts
Kavindu-Dodan: I think what this comment means is to avoid exposing the `FlagConfiguration` type.

Kavindu-Dodan (approved):
Approving the overal completeness of the implementation. Please check language specific reviews prior to merging.

packages/server/src/provider/in-memory-provider/flag-configuration.ts
toddbaert: Exactly

packages/server/src/provider/in-memory-provider/flag-configuration.ts
toddbaert: Good idea

packages/server/src/provider/in-memory-provider/flag-configuration.ts
toddbaert: I think this will live in shared when we implement it in web. We can do some commonization then

packages/server/src/provider/in-memory-provider/in-memory-provider.ts
toddbaert: I didn't see this, I'll look into it.

packages/server/README.md
toddbaert: Ya I plan on releasing immediately after this is merged

packages/server/src/provider/in-memory-provider/flag-configuration.ts
beeme1mr: Yeah, understood. I'm just trying to clarify if the comment is accurate.

packages/server/src/provider/in-memory-provider/flag-configuration.ts
luizgribeiro: From what I understood from the web version there is no need for the `EvaluationContext`. We could either have a base `Flag` type in shared and it being internally extended as needed or have one with all properties and use `Omit` to specialize.

packages/server/src/provider/in-memory-provider/in-memory-provider.ts
luizgribeiro: I did it based on [this comment](https://github.com/open-feature/js-sdk-contrib/pull/607#discussion_r1350716854).
I'm not really happy with it. Since it's rather difficult to make sure that an existing variant will always be available in a context based evaluation a `VariantNotFound` error would fit better in this case.

packages/server/src/provider/in-memory-provider/flag-configuration.ts
toddbaert: @luizgribeiro there's still `EvaluationContext` in web, its just not passed as a parameter from the flag evaluation API that users use. Instead it's set globally, representing a logged-in user, typically. The backend system and provider still receives a context though.

packages/server/src/provider/in-memory-provider/flag-configuration.ts
toddbaert: Looks like @luizgribeiro did this (thanks!)

toddbaert (approved):

beeme1mr (approved):

packages/server/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -0,0 +1,139 @@
...
+import { VariantFoundError } from './variant-not-found-error';
+
+/**
+ * A simple OpenFeature provider intended for demos and as a test stub.
+ */
+export class InMemoryProvider implements Provider {
+  public readonly events = new OpenFeatureEventEmitter();
+  public readonly runsOn = 'server';
+  readonly metadata = {
+    name: 'In-Memory Provider',
```
toddbaert, line 26:
```suggestion
    name: 'in-memory',
```

Issue #575, opened 2023-10-04
[proposal] rename @openfeature/js-sdk to @openfeature/server-sdk


toddbaert:
As came up in the last community meeting, `@openfeauter/js-sdk` is somewhat vague. We've heard of a few cases were a lot of time is wasted because devs import it in a client side web-app instead of `@openfeature/web-sdk`.

I propose we rename the package `@openfeature/server-sdk`. Reasoning:

- `server-sdk` makes clear it's not for web
- `server-sdk` doesn't restrict runtime to just node (deno/bun could be used)
- "JS" is redundant, and implied with npm/yarn.

We can publish a final version of the `@openfeaature/js-sdk` with a readme pointing to this new artifact, and a deprecation notice, and then deprecate the package.

EDIT: We can also decide _not_ to do this!

---

CONCLUSION:

It seems like there's a consensus for `server-sdk`.
I will push a final version of the @openfeature/js-sdk, with an updated readme referencing the new location.

justinabrahms:
I would prefer we don't name it `server-sdk` without putting javascript in there somewhere like `js-server-sdk`. `nodejs-sdk` would fit the bill, in my mind. I would guess (though I don't know) that deno/bun developers think "Oh, it says nodejs. It'll probably work for me", rather than "Oh, nodejs? That's definitely not going to work w/ deno/bun".

All of that said, maybe I'm actually resistant to the mismatch between github.com/open-feature/js-sdk and `@openfeature/server-sdk` ? I agree that as long as it's within the context of npm/yarn, it is much clearer.

liran2000:
In my opinion, something like _js-server-sdk_ make it more clear that it is _js_ SDK. Otherwise it may imply that this is the main server sdk, where it is similar to _go-sdk_, _java-sdk_ etc'.

thomaspoignant:
Agree with others here `@openfeature/js-server-sdk` and `@openfeature/web-sdk` would be better.
That being said it will probably cause breaking changes to the people already using it. 
Do we want to deprecate directly? or should we continue to deploy the package in the 2 names for some times?

tcarrio:
Putting the language in the name of a package that supports only that language is redundant. I would prefer we remove "JS" from the npm package name.

For the GitHub repository, "JS" in the name _does_ make sense.

|GitHub|NPM|
|-|-|
|js-web-sdk|@openfeature/web-sdk|
|js-server-sdk|@openfeature/server-sdk|

[...]

PR #589, opened 2023-10-11
fix: rename, deprecate js-sdk

toddbaert:
This is the first step in https://github.com/open-feature/js-sdk/issues/575. It's not really a fix, but this will force a new release so we can publish the readme.

This is my plan:

- publish this
- in a new PR:
  - rename the package and revert this readme change.
  - modify the bootstrap hash in release please to the HEAD hash (otherwise when we release the new package release please will look all the way back in time to get changes for the new artifact)
  - publish the new package
- use the console to deprecate the old package in npm (when you do this via console, you can add a custom message that prints on install, it will display: `npm WARN deprecated @openfeature/js-sdk: This package has been moved to @openfeature/server-sdk`)
- immediately update all server providers in [js-sdk-contribs](https://github.com/open-feature/js-sdk-contrib) to use the new package as their peer dependency, and release them all with a major version increase.


beeme1mr (approved):
This PR proves that naming is hard. It's a bummer we need to do this, but I think it's the right decision and the plan looks solid. Please make sure to update the docs and include a migration plan in the release notes.

kinyoklion (approved):

lukas-reining (approved):
As discussed, I find this a good step to reduce/remove the confusion about these package names! :)

PR #590, opened 2023-10-11
chore: expose transaction context propagation to the server sdk only

beeme1mr:
## This PR

- removes transaction context propagation from web SDK
- refactored the transaction context propagation logic to live under the server SDK

### Notes

Transaction context propagation only makes sense on the server and was simply ignored by the client. This removes the methods from the client sdk to avoid confusion.

This is a **breaking change** in the web SDK. However, it's sub 1.0 and this feature effectively had no effect if it was used. Currently, this is not marked as a breaking change to avoid releasing a major version. @toddbaert do you know if it's possible to mark the server sdk changes as a chore and the web sdk changes as a breaking fix?


toddbaert:
> @toddbaert do you know if it's possible to mark the server sdk changes as a chore and the web sdk changes as a breaking fix?

I don't think there's a way to do this other than manually editing the release PR, so that's the way I'd handle this. It wouldn't be hard - just manual changes to the PR description, the release notes, and the version number.

beeme1mr:
Hey @toddbaert @lukas-reining @thomaspoignant could one of you take a look at this PR when you have a moment?

packages/shared/tsconfig.json
```diff
@@ -98,7 +98,9 @@
   },
   "include": [
     "./src/**/*"
-  ],
+,
+"../server/src/transaction-context"
+],
```
toddbaert, line 103: Sorry I'm late to this PR, but I'm a bit confused by this.

It seems odd (and maybe even problematic?) that we need to compile source from server in shared.

PR #614, opened 2023-10-24
fix: add setProviderAndWait examples

### LINKED ISSUE #613: [DOC] update readme to include setProviderAndWait examples
### Change in the documentation

Update the server and client READMEs to include an example of how to use `setProviderAndWait`.

This should be modeled off the docs added to Java.
https://github.com/open-feature/java-sdk/commit/818131b77e48f985bc9e115085f49f228891393b

beeme1mr:
## This PR

- adds `setProviderAndWait` example on the server readme
- adds `setProviderAndWait` example on the client readme
- fixed invalid link on the server readme

### Related Issues

Fixes #613 

### Notes

The React SDK wasn't updated in this PR.


packages/client/README.md
```diff
@@ -130,22 +145,7 @@ sequenceDiagram
...
-
-// OpenFeature API
-OpenFeature.addHandler(ProviderEvents.Ready, () => {
-  const client = OpenFeature.getClient();
-  const stringFlag = client.getStringValue('string-flag', "default value"))
-
-  //use stringFlag from this point
-});
-```
+In order to prevent flag evaluations from defaulting while the provider is initializing, it is highly recommended to only look for flag value after the provider is ready. This can be done using the `setProviderAndWait` method or using the `setProvider` method and listening for the `READY` [event](#eventing).
```
toddbaert, line 148:
```suggestion
In order to prevent flag evaluations from defaulting while the provider is initializing, it is highly recommended to evaluate flags only after the provider is ready. This can be done using the `setProviderAndWait` method or using the `setProvider` method and listening for the `READY` [event](#eventing).
```

packages/server/README.md
```diff
@@ -106,10 +106,25 @@ If the provider you're looking for hasn't been created yet, see the [develop a p
...
+To register a provider in a synchronous manner, you can use the `setProvider` method as shown below:
+
 ```ts
-OpenFeature.setProvider(new MyProvider())
+OpenFeature.setProvider(new MyProvider());
 ```
 
+> [!NOTE]  
+> The status of the provider can be tracked using [events](#eventing).
+
```
toddbaert, line 127: Just outta curiosity, how do these sort of notes render in docusaurus?

toddbaert (approved):
approved with questions / minor nits.

packages/server/README.md
beeme1mr: They don't work... it appears GitHub make up this syntax and it's not widely supported. Removing the fancy rendering for now.

packages/client/README.md
```diff
@@ -109,10 +109,24 @@ If the provider you're looking for hasn't been created yet, see the [develop a p
...
 
+#### Awaitable
+
+To register a provider and ensure it is ready before further actions are taken, you can use the `setProviderAndWait` method as shown below:
+
+```ts
+await OpenFeature.setProviderAndWait(new MyProvider());
+```  
+
+#### Synchronous
```
thomaspoignant, line 120: Should it be Asynchronous here ?

beeme1mr: I really struggled with the name here. Techically, the method is synchronous but the provider may not be ready immediately. Calling it asynchronous would imply that it would return a promise or accept a callback. However, I do agree that this is confusing. Can you think of a better term we could use?

packages/server/README.md
thomaspoignant: To do the same thing in docusaurus you can use 
```markdown
:::note

The status of the provider can be tracked using [events](#eventing).
```

https://docusaurus.io/docs/next/markdown-features/admonitions

packages/client/README.md
thomaspoignant: Yes, the naming is super complex on this one.
We can use something like **blocking** / **non blocking** maybe ?

packages/server/README.md
```diff
@@ -81,7 +81,7 @@ if (v2Enabled) {
 
 ### API Reference
 
-See [here](https://open-feature.github.io/js-sdk/modules/OpenFeature_JS_SDK.html) for the complete API documentation.
+See [here](https://open-feature.github.io/js-sdk/modules/OpenFeature_Server_SDK.html) for the complete API documentation.
```
beeme1mr, line 84: This has changed

packages/client/README.md
toddbaert: I personally think this is accurate. The method returns immediately. The fact is starts some asynchronous jobs doesn't make the method itself asynchronous.

toddbaert:
Changed the title so it prompts a release to update npm notes.

PR #617, opened 2023-10-26
feat: client in memory provider

### LINKED ISSUE #565: [FEATURE] In-memory provider for js (node) and web SDKs

luizgribeiro:
## This PR

- Adds in memory provider implementation for client (web) sdk
- Move files from server sdk to shared

### Related Issues

Closes #565 

### Notes

It's a first try. Any feedback and enhancement proposals are welcome :)

### How to test

Automated/unit testes were implemented, but e2e tests wouldn't hurt.


packages/client/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -0,0 +1,186 @@
...
+   * Overwrites the configured flags.
+   * @param { FlagConfiguration } flagConfiguration new flag configuration
+   */
+  putConfiguration(flagConfiguration: FlagConfiguration) {
+    const flagsChanged = Object.entries(flagConfiguration)
+      .filter(([key, value]) => this._flagConfiguration[key] !== value)
+      .map(([key]) => key);
+
+    this.status = ProviderStatus.STALE;
+    this.events.emit(ProviderEvents.Stale);
```
beeme1mr, line 69: This behavior doesn't match what was done in the server SDK.

https://github.com/open-feature/js-sdk/blob/main/packages/server/src/provider/in-memory-provider/in-memory-provider.ts

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -0,0 +1,186 @@
...
+    this._flagConfiguration = { ...flagConfiguration };
+    this.events.emit(ProviderEvents.ConfigurationChanged, { flagsChanged });
+  }
+
+  /**
+   * Overwrites the configured flags & updates internal cache (resolved flags).
+   * @param { FlagConfiguration } flagConfiguration new flag configuration
+   * @param { EvaluationContext } [context] flag evaluation context
+   */
+  async putConfigurationAndUpdateResolution(flagConfiguration: FlagConfiguration, context?: EvaluationContext | undefined): Promise<void> {
```
beeme1mr, line 80: When should a user use this method instead of `putConfiguration`?

beeme1mr (commented):
I didn't have a chance to give it a detailed review but I wanted to provide some quick feedback. Overall, it looks good but I had a few questions. Also, would it be possible to share some of the logic across client and server? @toddbaert is working on publishing the shared components right now. Perhaps that work could be leveraged here.

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
luizgribeiro: The idea here was to mimic the state of a "real" client side providers, since it evaluates all flags at once.
I think the server version of the in memory provider can also set it's state to stale and emit the stale event when it begins to process the new config file, but since it's a synchronous process it wouldn't be much valuable. It would also need to emit Ready and set state when it finishes. 

In terms of behavior that's ok, but I don't know how valuable this is in practice.

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
luizgribeiro: I think this question is also related to the previous one.

My idea was to:
1. Provide a method in which people could force provider's state to be stale (probably valuable while testing)
2. Provide another straight forward method to change flag configuration

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -0,0 +1,186 @@
...
+  } as const;
+  private _flagConfiguration: FlagConfiguration;
+  private _flagEvalCache: Record<string, ResolutionDetails<number | string | boolean | JsonValue>> = {};
+
+  constructor(flagConfiguration: FlagConfiguration = {}) {
+    this._flagConfiguration = { ...flagConfiguration };
+  }
+
+  async initialize(context?: EvaluationContext | undefined): Promise<void> {
+    this.status = ProviderStatus.READY;
```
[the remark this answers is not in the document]
luizgribeiro: I tried this before and ended up doing it this way :sweat_smile: 

Because of how `lookUpFlagValue` is implemented the tests suite broke.
The provider state is `stale`, and it gives `cached` as reason.

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
lukas-reining: I also find this a little confusing. 
To me it feels that besides from changing the behavior in [line 130](https://github.com/open-feature/js-sdk/pull/617/files#diff-fa6dedaedba26b152e2aa7fa49d5e27aa197def6d2334aef668dba9b8a5d4a0aR130) if some flag is disabled,`putConfiguration` does not change the flag resolution, which means that I would alwys only use `putConfigurationAndUpdateResolution`. 
My feeling is that we only need one method that always changes the resolved flags as this method does.

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -0,0 +1,186 @@
...
+    defaultValue: T, logger?: Logger): ResolutionDetails<T> {
+    if (!(flagKey in this._flagEvalCache)) {
+      const message = `no flag found with key ${flagKey}`;
+      logger?.debug(message);
+      throw new FlagNotFoundError(message);
+    }
+
+    if (this._flagConfiguration[flagKey].disabled) {
+      return { value: defaultValue, reason: StandardResolutionReasons.DISABLED };
+    }
```
lukas-reining, line 132: This is somewhat odd. As a user of this provider, I do not know of the difference between cache and flagConfiguration.
Here, the configuration might change the resolution, even though the cache might not be disabled.
My feeling is that we should only have one source of truth for resolving.
If we want to keep the cache, then we should use the `caches disabled value` of the flag here.

lukas-reining (commented):
This looks good to me!
Left some comments, mainly regarding the usage of the cache.
Please correct me if I see something wrong here.

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
lukas-reining: > Provide a method in which people could force provider's state to be stale (probably valuable while testing)

This actually behaves a bit odd here. As stated in the other comment, the only thing that is change here is the disabled state. The rest of the definitions does not change for the user. Maybe we could just have a method for setting the provider to stale without changing the data, if we want to support the use case  you described.

lukas-reining: > Provide another straight forward method to change flag configuration

The problem is, that `putConfiguration` does not change the flag configuration observed by the user (besides disabled status) if I see correctly.

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -0,0 +1,186 @@
...
+ */
+export class InMemoryProvider implements Provider {
+  public readonly events = new OpenFeatureEventEmitter();
+  public readonly runsOn = 'client';
+  status: ProviderStatus = ProviderStatus.NOT_READY;
+  readonly metadata = {
+    name: 'in-memory',
+  } as const;
+  private _flagConfiguration: FlagConfiguration;
+  private _flagEvalCache: Record<string, ResolutionDetails<number | string | boolean | JsonValue>> = {};
```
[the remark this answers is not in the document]
lukas-reining: After reading your comments I get the point of the cache.
My feeling is that we might then would be better of just being able to control the stale and ready state more fine grained. Currently yes, we are stale for an extremely short time but we are not really in control of it as only calling `putConfiguration` will lead to the stale state but we will never get back to ready without changing the `_flagConfiguration` again.

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
luizgribeiro: That's right! It does not change the cache, so the evaluation `value` is the same.
But since it sets `status=STALE` the reason will be different

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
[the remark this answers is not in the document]
luizgribeiro: I get your point.

> we are not really in control of it as only calling putConfiguration will lead to the stale state but we will never get back to ready without changing the _flagConfiguration again.

The same flagConfiguration could be provided to get it back to ready state, but it's confusing

luizgribeiro:
Hey folks! Thanks for the feedback.

Based on what you said, I'm planning on:
1. Getting rid of flag cache: I'd rather keep `initialize` and use it to guarantee that nothing breaks with context evaluation and set provider's status as ready
2. Unifying `putConfiguration` and `putConfigurationAndUpdateResolution`: I think @toddbaert said something about making the update and status cycle change invisible. It should probably call `initialize` internally then
3. Provide a way to change/control provider's status 

What do you think about that? @beeme1mr @lukas-reining

beeme1mr:
> 3\. Provide a way to change/control provider's status

Is this for supporting various test use cases?

lukas-reining:
Sound good to me @luizgribeiro :)

toddbaert:
> Hey folks! Thanks for the feedback.
> 
> Based on what you said, I'm planning on:
> 
> 1. Getting rid of flag cache: I'd rather keep `initialize` and use it to guarantee that nothing breaks with context evaluation and set provider's status as ready

:+1: 

> 2. Unifying `putConfiguration` and `putConfigurationAndUpdateResolution`: I think @toddbaert said something about making the update and status cycle change invisible. It should probably call `initialize` internally then

:+1: 

> 3. Provide a way to change/control provider's status

I think this might add confusion. It's easy enough to make a testing provider to test this stuff, but because this is all "in-memory", I don't expect a lot of complexity around statuses. The only things I would expect in this care are `NOT_READY` (before init) `READY` (after succeessful init) and perhaps `ERROR` if somebody tries to put an invalid flag definition, but its not required for you to validate ahead of time - you could just throw during the eval if you'd like.

beeme1mr:
Hey @luizgribeiro, it looks like there are some merge conflicts since we moved some files to `shared`. Would you be able to update this PR to absorb those changes?

luizgribeiro:
Hey @beeme1mr! Sure, I'll probably have some time to work on this during the weekend.

luizgribeiro:
I've just pushed a newer version that instead of caching flag resolution it works based on the points mentioned above and store context internally to make it easier to test (even though it can also use context provided in resolution). 
Let me know wyt.

Sorry for taking too long on this

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      for (const key in this._flagConfiguration) {
+        this.resolveFlagWithReason(key, context);
+      }
+
+      this._context = context;
+      this.status = ProviderStatus.READY;
+    } catch (error) {
+      this.status = ProviderStatus.ERROR;
+      throw error;
```
toddbaert, line 48: Originally you were manually emitting ready/error events here, which the SDK doesn't based on the resolve/reject of this promise. This caused duplicate events.

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    this._flagConfiguration = { ...flagConfiguration };
+    this.events.emit(ProviderEvents.ConfigurationChanged, { flagsChanged });
+
+    try {
+      await this.initialize(this._context);
+      this.events.emit(ProviderEvents.Ready);
+    } catch (err) {
+      this.events.emit(ProviderEvents.Error);
+      throw err;
```
toddbaert, line 72: We do have to emit our own events here, which we do after the init call.

toddbaert (approved):
Hey @luizgribeiro ! This latest work looks great. I did however revert all changes that weren't `/client`. My reasoning is threefold:

- this is a `feat` only for client, so we need to scope our changes there so the release automation only makes changes there
- artifacts specific to the in-memery provider, such as the `VariantFoundError` and the `FlagConfiguration` and `Flag` types cannot be exported from shared, or I think they will _really_ confuse people. These are only implementation details for the in-memory provider, and not part of the actual SDK API. The only thing we should export is the in-memory provider itself.
- in addition to being useful in tests, this serves as an example, so a bit of duplication isn't bad in this case


I made only a few actual changes to the provider itself, which I've noted [here](https://github.com/open-feature/js-sdk/pull/617/files#r1400957519).

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
beeme1mr: It may be worth adding this as a comment to the provider since people may use the in-memory provider as a point of reference.

packages/client/src/provider/in-memory-provider/variant-not-found-error.ts
```diff
@@ -0,0 +1,15 @@
+import { ErrorCode, OpenFeatureError } from '@openfeature/core';
+
+/**
+ * A custom error for the in-memory provider.
+ * Indicates the resolved or default variant doesn't exist.
+ */
+export class VariantFoundError extends OpenFeatureError {
```
beeme1mr, line 7:
```suggestion
export class VariantNotFoundError extends OpenFeatureError {
```

packages/client/src/provider/in-memory-provider/variant-not-found-error.ts
```diff
@@ -0,0 +1,15 @@
...
+  code: ErrorCode;
+  constructor(message?: string) {
+    super(message);
+    Object.setPrototypeOf(this, VariantFoundError.prototype);
+    this.name = 'VariantFoundError';
```
beeme1mr, line 12:
```suggestion
    this.name = 'VariantNotFoundError';
```

beeme1mr (approved):

toddbaert:
> It may be worth adding this as a comment to the provider since people may use the in-memory provider as a point of reference.

@beeme1mr good idea. I will do this generally.

toddbaert:
Thanks again @luizgribeiro . I will merge this if you don't have any objections to my changes to your branch.

packages/client/src/provider/in-memory-provider/variant-not-found-error.ts
luizgribeiro: This will probably need renaming in the server provider as well.

Issue #616, opened 2023-10-26
Publish shared components


toddbaert:
Currently we do not publish a "shared" module for common type/interfaces/classes between the web and server SDKs. This hasn't been too much of an issue so far, particularly since providers don't share exactly the same interface between both SDKs. However, recently, there's been a demonstrable need for a common artifact, for example, the flagd-in-process lib [here](https://github.com/open-feature/js-sdk-contrib/pull/620). This artifact could be useful in both web/server contexts. There's probably more use-cases as well (maybe even some hooks?).

Now that we know there's a use case, we should:

- publish the shared module (we can decide on a good name)
- set it as a dependency of the web/server/react SDKs
- re-export all dependencies from it in the server/web SDKs (this will prevent us from requiring users to update all their imports)

toddbaert:
@thomaspoignant @lukas-reining you guys have any concerns with this?

lukas-reining:
Well, sounds like a good idea to me.
My only concern is that we could not have any shared "SDK internals" then. If we put all the stuff the SDKs share in this package, anyone using it may depend on implementation details that are not meant to be part of the public api.
We could avoid this by having the published shared package and shared internals separated but I fear this may lead to inconsistencies🤔
Or maybe this is just not a problem because we do not have such "implementation details" that we would expose, I have no time to check right now but will have a look tomorrow.

toddbaert:
> Well, sounds like a good idea to me. My only concern is that we could not have any shared "SDK internals" then. If we put all the stuff the SDKs share in this package, anyone using it may depend on implementation details that are not meant to be part of the public api. We could avoid this by having the published shared package and shared internals separated but I fear this may lead to inconsistencies🤔 Or maybe this is just not a problem because we do not have such "implementation details" that we would expose, I have no time to check right now but will have a look tomorrow.

Currently, we bundle and publish the contents of "shared" already, as part of the web/server modules - if they are exported, they are already exposed publicly through those modules.

I can't think of anything that fits into this category though, besides maybe the `OpenFeatureCommonAPI` abstract class... and I'm not sure it's really much of a problem. :thinking:

thomaspoignant:
I like the idea a lot, and it can avoid mixing things in a provider.
It happened to me for example with the `EvaluationContext` see https://github.com/open-feature/js-sdk-contrib/pull/583.

Using a shared component may help to avoid that.

lukas-reining:
>  if they are exported, they are already exposed publicly through those modules

[...]

PR #650, opened 2023-11-02
fix: remove events.js from core module

toddbaert:
Optimization to `@openfeature/core` package. Both server and web need the node-js `Events` interfaces, but obviously only the web needs them poly-filled (they come with node).

This PR:

- marks `events.js` as an external dep in `@openfeature/core` meaning it won't be bundled there (see esbuild [doc](https://esbuild.github.io/api/#external))

This means that now `@openfeature/core` no longer includes `events.js` (an un-needed polyfill for the server). The bundling pictured below only happens in the `dist/` of `@openfeature/web-sdk`, and no longer in `@openfeature/core` :

image

I've tested both the server and web SDKs based on this change locally in the js-contribs by running the e2e test suite. 

Relates to: https://github.com/open-feature/js-sdk/issues/616


packages/shared/package.json
```diff
@@ -19,8 +19,8 @@
     "lint": "eslint ./",
     "clean": "shx rm -rf ./dist",
     "type": "tsc --project ./tsconfig.json --declaration --emitDeclarationOnly",
-    "build:esm": "esbuild src/index.ts --bundle --sourcemap --target=es2016 --format=esm --outfile=./dist/esm/index.js --analyze",
-    "build:cjs": "esbuild src/index.ts --bundle --sourcemap --target=es2016 --format=cjs --outfile=./dist/cjs/index.js --analyze",
+    "build:esm": "esbuild src/index.ts --bundle --external:events --sourcemap --target=es2016 --format=esm --outfile=./dist/esm/index.js --analyze",
```
toddbaert, line 22: We do this same thing already in web, except it's `--external:@openfeature/core`, which prevent us from bundling the core module there.

lukas-reining (approved):
Looks good to me!

Kavindu-Dodan (approved):

Issue #659, opened 2023-11-07
[BUG] Importing the @openfeature/launchdarkly-client-provider throws a TypeError


juanparadox:
### Observed behavior

I installed the `@openfeature/web-sdk` and `@openfeature/launchdarkly-client-provider` packages using `pnpm install`. 

Upon trying to simply import the `LaunchDarklyClientProvider` via `import { LaunchDarklyClientProvider } from "@openfeature/launchdarkly-client-provider";` and running my local app (it's the boilerplate Vite React project that the Vite install provides) a TypeError.

```
open-feature-event-emitter.ts:11 Uncaught TypeError: import_events.default is not a constructor
    at new GenericEventEmitter (open-feature-event-emitter.ts:11:35)
    at new InternalEventEmitter (open-feature-event-emitter.ts:83:8)
    at new OpenFeatureCommonAPI (open-feature.ts:21:30)
    at new OpenFeatureAPI (open-feature.ts:18:5)
    at OpenFeatureAPI.getInstance (open-feature.ts:32:22)
    at open-feature.ts:92:43
```

### Expected Behavior

I expected the provider to be able to be imported without a `TypeError` being thrown so that it can be instantiated into the `OpenFeature.setProvider` method.

### Steps to reproduce

1. Install the `@openfeature/web-sdk` library (This may not be necessary as the issue occurs when just importing the provider library below)
2. Install the `@openfeature/launchdarkly-client-provider` library
3. Create a new file, say `featureFlags.ts` with the following (this is as bare of code you can write to reproduce the issue, I understand it's not creating a LaunchDarklyClientProvider instance or doing anything with the import, you could skip this step and just import the provider library directly into a file your app is already using)

```
import { LaunchDarklyClientProvider } from "@openfeature/launchdarkly-client-provider";
```

4. Import that file into your app, something like

```
import './featureFlags';
```
5. TypeError is thrown upon running the application

juanparadox:
@toddbaert Just created the issue about the constructor error. Please let me know if you need more details.

toddbaert:
Thanks @juanparadox . Can you confirm the versions you're using (I'm assuming both @ latest)?

And are you using yarn or npm? (They behave a bit differently with peer dependencies, which may be related to the problem).

juanparadox:
@toddbaert Yeah! They are both at the latest versions and I am using pnpm.

toddbaert:
[...]

PR #660, opened 2023-11-08
fix: missing events bundled dep

toddbaert:
Fixes: https://github.com/open-feature/js-sdk/issues/659

This PR fixes an issue created with [this change](https://github.com/open-feature/js-sdk/pull/650), which removed the `events` polyfill package from `@openfeature/core` where is wasn't always needed (see that issue for details). The problem was that we still imported `events` in the `@openfeature/core` module, but can't use the `events` bundled in the `@openfeature/web-sdk` since the bundled package there isn't accessible from imports in `@openfeature/core`.

This PR _**removes all imports of `events`**_ from `@openfeature/core`, and instead only imports types. Imports of `events` only now occur in the web-sdk (where it's bundled) and server-sdk (where it's made available by the node runtime), not in the common module.

Unfortunately this issue was a bit tough to track down, because `events` is VERY common, and lots of bundlers, etc will add it, so it's frequently available "accidentally".

Thanks to @juanparadox for the report.

### DIFF
--- a/packages/client/tsconfig.json
+++ b/packages/client/tsconfig.json
@@ -28,7 +28,9 @@
     // "rootDir": "./",                                  /* Specify the root folder within your source files. */
     "moduleResolution": "node", /* Specify how TypeScript looks up a file from a given module specifier. */
     // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
-    // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
+    "paths": {
+      "@openfeature/core": [ "../shared/src" ]
+    },                                                   /* Specify a set of entries that re-map imports to additional lookup locations. */
     // "rootDirs": [],                                   /* Allow multiple folders to be treated as one when resolving modules. */
     // "typeRoots": [],                                  /* Specify multiple folders that act like `./node_modules/@types`. */
     // "types": [],                                      /* Specify type package names to be included without being referenced in a source file. */
--- a/packages/shared/src/events/generic-event-emitter.ts
+++ b/packages/shared/src/events/generic-event-emitter.ts
@@ -1,21 +1,17 @@
-import EventEmitter from 'events';
 import { Logger, ManageLogger, SafeLogger } from '../logger';
-import { CommonEventDetails, EventContext, EventDetails, EventHandler } from './eventing';
+import { EventContext, EventDetails, EventHandler } from './eventing';
 import { ProviderEvents } from './events';
 
-abstract class GenericEventEmitter<AdditionalContext extends Record<string, unknown> = Record<string, unknown>>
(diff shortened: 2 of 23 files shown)


packages/client/tsconfig.json
```diff
@@ -28,7 +28,9 @@
     // "rootDir": "./",                                  /* Specify the root folder within your source files. */
     "moduleResolution": "node", /* Specify how TypeScript looks up a file from a given module specifier. */
     // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
-    // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
+    "paths": {
+      "@openfeature/core": [ "../shared/src" ]
+    },                                                   /* Specify a set of entries that re-map imports to additional lookup locations. */
```
toddbaert, line 33: this forces resolution of src/ instead of dist/ during development. No prod impact.

packages/shared/src/events/generic-event-emitter.ts
```diff
@@ -1,21 +1,17 @@
-import EventEmitter from 'events';
```
toddbaert, line 1: This import failed, and was the cause of the bug. Even though we bundle `events` in the `@openfeature/web-sdk`, it's not available here.

packages/shared/src/events/generic-event-emitter.ts
```diff
@@ -1,21 +1,17 @@
...
 import { Logger, ManageLogger, SafeLogger } from '../logger';
-import { CommonEventDetails, EventContext, EventDetails, EventHandler } from './eventing';
+import { EventContext, EventDetails, EventHandler } from './eventing';
 import { ProviderEvents } from './events';
 
-abstract class GenericEventEmitter<AdditionalContext extends Record<string, unknown> = Record<string, unknown>>
+export abstract class GenericEventEmitter<AdditionalContext extends Record<string, unknown> = Record<string, unknown>>
```
toddbaert, line 5:
```suggestion
/**
 * The GenericEventEmitter should only be used within the SDK. It supports additional properties that can be included
 * in the event details.
 */
export abstract class GenericEventEmitter<AdditionalContext extends Record<string, unknown> = Record<string, unknown>>
```

juanparadox:
@toddbaert This is awesome. Thank you so much for moving on this so quickly. We were doing some discovery work on available feature flagging tools for the web and this is at the top of our adoption list 😄

toddbaert:
> @toddbaert This is awesome. Thank you so much for moving on this so quickly. We were doing some discovery work on available feature flagging tools for the web and this is at the top of our adoption list 😄

Thanks for finding the issue.

FWIW, I think if you use the last published versions of `@openfeature/core` and `@openfeature/web-sdk`, you will not have this issue. I will try to get this fix released tonight though, once it's reviewed.

beeme1mr (approved):

lukas-reining (approved):

PR #671, opened 2023-11-20
fix: make hooks in client sdk only return void

### LINKED ISSUE #630: [BUG] Web SDK hooks can return a promise?
Both the web and server SDK use a hook interface defined in `/shared` aka `core`.
In web, the hook calls are not awaited, so the promise is somewhat useless.

I'm fairly certain this is a bug. We can allow the server to return both promises and non, but the web SDK cannot return Promises in hooks because it doesn't await them and has a synchronous API.

cc @beeme1mr @lukas-reining

lukas-reining:
## This PR

Fixes return type of hooks in client SDK. 

Maybe we should make this more clear in the spec @toddbaert.

### Related Issues
Fixes #630


beeme1mr (commented):
This looks good but I want to manually test before I approve. The change was bigger than I expected 😄

lukas-reining:
> This looks good but I want to manually test before I approve. The change was bigger than I expected 😄

Thanks! Well It was not much but the hook types just touched many files. 
And many of these were not formatted with the attached prettier which is why this might seem a bit verbose.

packages/client/src/client/open-feature-client.ts
```diff
@@ -22,6 +21,7 @@ import { OpenFeature } from '../open-feature';
 import { Provider } from '../provider';
 import { InternalEventEmitter } from '../events/internal/internal-event-emitter';
 import { Client } from './client';
+import { Hook } from '../hooks/hook';
```
toddbaert, line 24:
```suggestion
import { Hook } from './hooks';
```

This kind of thing can sometimes lead to import cycles, for prod code it's usually best to import from the barrel file.

I wonder if we can somehow add linting for this :thinking:

packages/server/src/client/open-feature-client.ts
```diff
@@ -16,13 +15,14 @@ import {
...
   StandardResolutionReasons,
-  statusMatchesEvent
+  statusMatchesEvent,
 } from '@openfeature/core';
 import { FlagEvaluationOptions } from '../evaluation';
 import { OpenFeature } from '../open-feature';
 import { Provider } from '../provider';
 import { Client } from './client';
 import { InternalEventEmitter } from '../events/internal/internal-event-emitter';
+import { Hook } from '../hooks/hook';
```
toddbaert, line 25:
```suggestion
import { Hook } from '../hooks';
```

packages/server/src/evaluation/evaluation.ts
```diff
@@ -1,4 +1,5 @@
-import { EvaluationContext, EvaluationDetails, Hook, HookHints, JsonValue } from '@openfeature/core';
+import { EvaluationContext, EvaluationDetails, HookHints, JsonValue } from '@openfeature/core';
+import { Hook } from '../hooks/hook';
```
toddbaert, line 2:
```suggestion
import { Hook } from '../hooks';
```

packages/server/src/provider/provider.ts
```diff
@@ -1,4 +1,5 @@
-import { CommonProvider, EvaluationContext, Hook, JsonValue, Logger, ResolutionDetails } from '@openfeature/core';
+import { CommonProvider, EvaluationContext, JsonValue, Logger, ResolutionDetails } from '@openfeature/core';
+import { Hook } from '../hooks/hook';
```
toddbaert, line 2:
```suggestion
import { Hook } from '../hooks';
```

packages/server/test/hooks.spec.ts
```diff
@@ -1,4 +1,5 @@
-import { OpenFeature, Provider, ResolutionDetails, Client, FlagValueType, EvaluationContext, Hook } from '../src';
+import { OpenFeature, Provider, ResolutionDetails, Client, FlagValueType, EvaluationContext } from '../src';
+import { Hook } from '../src/hooks/hook';
```
toddbaert, line 2:
```suggestion
import { Hook } from '../src/hooks';
```

packages/server/test/hooks.spec.ts
```diff
@@ -1,4 +1,5 @@
-import { OpenFeature, Provider, ResolutionDetails, Client, FlagValueType, EvaluationContext, Hook } from '../src';
+import { OpenFeature, Provider, ResolutionDetails, Client, FlagValueType, EvaluationContext } from '../src';
```
toddbaert, line 1:
```suggestion
import { OpenFeature, Provider, ResolutionDetails, Client, FlagValueType, EvaluationContext, Hook } from '../src';
```

toddbaert, line 1:
```suggestion
import { OpenFeature, Provider, ResolutionDetails, Client, FlagValueType, EvaluationContext, Hook } from '../src';
```

packages/client/src/client/open-feature-client.ts
lukas-reining: Ah added the barrel later and missed that. Thanks!
And yes I think that should not be a problem to lint this, maybe in conjunction with import optimizations.

packages/server/test/hooks.spec.ts
toddbaert, line 2:
```suggestion
import { OpenFeature, Provider, ResolutionDetails, Client, FlagValueType, EvaluationContext, Hook } from '../src';
```

packages/client/src/open-feature.ts
```diff
@@ -2,6 +2,7 @@ import { EvaluationContext, ManageContext, OpenFeatureCommonAPI } from '@openfea
 import { Client, OpenFeatureClient } from './client';
 import { NOOP_PROVIDER, Provider } from './provider';
 import { OpenFeatureEventEmitter } from './events';
+import { Hook } from './hooks/hook';
```
toddbaert, line 5:
```suggestion
import { Hook } from './hooks';
```

packages/client/src/provider/provider.ts
```diff
@@ -1,4 +1,5 @@
-import { CommonProvider, EvaluationContext, Hook, JsonValue, Logger, ResolutionDetails } from '@openfeature/core';
+import { CommonProvider, EvaluationContext, JsonValue, Logger, ResolutionDetails } from '@openfeature/core';
+import { Hook } from '../hooks/hook';
```
toddbaert, line 2:
```suggestion
import { Hook } from '../hooks';
```

packages/client/test/hooks.spec.ts
```diff
@@ -4,10 +4,10 @@ import {
   Client,
   FlagValueType,
   EvaluationContext,
-  Hook,
   GeneralError,
   OpenFeature,
 } from '../src';
+import { Hook } from '../src/hooks/hook';
```
toddbaert, line 10: Add this to the other imports from `../src`

packages/server/src/open-feature.ts
```diff
@@ -14,6 +14,7 @@ import {
 } from './transaction-context';
 import { Client, OpenFeatureClient } from './client';
 import { OpenFeatureEventEmitter } from './events';
+import { Hook } from './hooks/hook';
```
toddbaert, line 17:
```suggestion
import { Hook } from './hooks';
```

packages/shared/src/hooks/hook.ts
```diff
@@ -1,17 +1,14 @@
 import { BeforeHookContext, HookContext, HookHints } from './hooks';
-import { EvaluationContext, EvaluationDetails, FlagValue } from '../evaluation';
+import { EvaluationDetails, FlagValue } from '../evaluation';
 
-export interface Hook<T extends FlagValue = FlagValue> {
+export interface BaseHook<T extends FlagValue = FlagValue, BeforeHookReturn = unknown, HooksReturn = unknown> {
```
toddbaert, line 4: I'm glad you went with a generic solution! Thsi is great!

Interestingly, the client hook generic constraints are a subset of the server hook's constraints. So basically, it seems like you could use the server hook's types here, and then restrict them in the client

```ts
export interface BaseHook<
T extends FlagValue = FlagValue,
BeforeHookReturn = Promise<EvaluationContext | Promise<void>> | EvaluationContext | void,
HooksReturn = Promise<void> | void> {
```

toddbaert (approved):
Awesome work. Besides the imports, I made [this comment](https://github.com/open-feature/js-sdk/pull/671/files#r1400673771), which is just an interesting observation.

packages/shared/src/hooks/hook.ts
lukas-reining: Would be okay to me too.
In my view, the current type is more explicit. 
We just have two invariants and both are defined "in-place" where your proposal implies to me that the client is a "special case" and the server is the "normal one".

toddbaert:
@lukas-reining I'm sure one of my import changes broke the build, I was only doing them in github. I guess that in general you should be able to import from the barrel files though. :sweat_smile:

lukas-reining:
> @lukas-reining I'm sure one of my import changes broke the build, I was only doing them in github. I guess that in general you should be able to import from the barrel files though. 😅

Fixed, it was one of them missing a dot 😅 Good observation.

toddbaert:
Oh, I should also mention, I packaged this locally and linked it in the `js-contribs`: everything built fine there, so exports/types/etc seem good. cc @beeme1mr

beeme1mr (approved):
I tested the API doc generation as well. Looks good to me. Thank you @lukas-reining 🙇

PR #681, opened 2023-11-22
fix: rm NodeJS type from core, rm react core peer

toddbaert:
* remove a NodeJS type from core module which needs to be platform agnostic
* remove core peer dep from react sdk, which gets core through web-sdk

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -24,8 +24,7 @@
     "postbuild": "shx cp ./../../package.esm.json ./dist/esm/package.json",
     "current-version": "echo $npm_package_version",
     "prepack": "shx cp ./../../LICENSE ./LICENSE",
-    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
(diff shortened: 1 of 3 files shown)


packages/shared/src/events/generic-event-emitter.ts
```diff
@@ -9,7 +9,7 @@ import { ProviderEvents } from './events';
 export abstract class GenericEventEmitter<AdditionalContext extends Record<string, unknown> = Record<string, unknown>>
   implements ManageLogger<GenericEventEmitter<AdditionalContext>>
 {
-  protected abstract readonly eventEmitter: NodeJS.EventEmitter;
```
toddbaert, line 12: This could create an issue in some TS consumers which perform type-checking in libs when they compile/bundle.

Fixed by adding the private type polyfill below.

beeme1mr (approved):

packages/react/package.json
```diff
@@ -47,12 +46,11 @@
   },
   "homepage": "https://github.com/open-feature/js-sdk#readme",
   "peerDependencies": {
-    "@openfeature/core": "0.0.19",
```
toddbaert, line 50: This is provided via the web-sdk, and specifying it here increased the likelyhood of peer conflicts.

lukas-reining (approved):
Looks good to me!
`PlatformEventEmitter` seems like a good way of avoiding the bundling issue as the EventEmitter interface will also be stable enough.

PR #691, opened 2023-11-27
chore: fix React SDK README.md to use the resolved value instead of the resolution details

juanparadox:
## This PR

The documentation conditionally renders something if the returned value from the hook is truthy, but this implies that it is simply checking for the entirety of the `EvaluationDetails<T>` returned from `useFeatureFlag` which seems off since we will always get an object back according to the typings. It feel like it should be checking against the `.value` to conditionally render the elements.


lukas-reining (approved):
Thanks for fixing this! Looks good to me.

toddbaert (approved):
Yes we changed this and missed the doc update. Nice catch!

PR #693, opened 2023-11-27
fix: type support for Promise<void> before hook

### LINKED ISSUE #689: [BUG] typings don't support returning Promise<void>
### Observed behavior

It looks like we have an error with the typings for the hooks when it comes to returning a `Promise<void>` from the server-sdk:

image

'Promise<void>' is not assignable to type 'Promise<EvaluationContext>' is the error.

I think the problem is [here](https://github.com/open-feature/js-sdk/blob/cc9fe79c556365d9cd6542d9c443a14166de4e08/packages/server/src/hooks/hook.ts#L5). We have `Promise<EvaluationContext | Promise<void>>`. I think this should read: `Promise<EvaluationContext> | Promise<void>` instead, or I'm missing something.

cc @beeme1mr @lukas-reining 

### Expected Behavior

_No response_

### Steps to reproduce

_No response_

beeme1mr:
## This PR

- Updates the server-side before hook type to perform async work without modifying context.

### Related Issues

Fixes #689

### Notes

I added a test but the test runner isn't not validating types.

### DIFF
--- a/packages/server/test/hooks.spec.ts
+++ b/packages/server/test/hooks.spec.ts
@@ -207,7 +207,8 @@ describe('Hooks', () => {
       const clientPropToOverwrite434 = 'clientPropToOverwrite';
       const invocationProp434 = 'invocationProp';
       const invocationPropToOverwrite434 = 'invocationPropToOverwrite';
-      const hookProp434 = 'hookProp';
+      const syncHookProp434 = 'syncHookProp';
+      const asyncHookProp434 = 'asyncHookProp';
 
       OpenFeature.setContext({
         [globalProp434]: true,
@@ -223,9 +224,12 @@ describe('Hooks', () => {
         [invocationPropToOverwrite434]: false,
         [clientPropToOverwrite434]: true,
       };
-      const hookContext = {
+      const syncHookContext = {
         [invocationPropToOverwrite434]: true,
-        [hookProp434]: true,
+        [syncHookProp434]: true,
+      };
+      const asyncHookContext = {
+        [asyncHookProp434]: true,
       };
 
       const localClient = OpenFeature.getClient('merge-test', 'test', clientContext);
@@ -234,7 +238,23 @@ describe('Hooks', () => {
         hooks: [
           {
             before: () => {
-              return hookContext;
+              // synchronous hook that doesn't modify context
+            },
+          },
+          {
+            before: () => {
+              return syncHookContext;
+            },
+          },
+          {
+            before: async () => {
+              // asynchronous hook that doesn't modify context
+              await Promise.resolve();
+            },
+          },
+          {
+            before: async () => {
+              return Promise.resolve(asyncHookContext);
             },
           },
         ],
@@ -250,7 +270,8 @@ describe('Hooks', () => {
           [clientPropToOverwrite434]: true,
           [invocationProp434]: true,
           [invocationPropToOverwrite434]: true,
-          [hookProp434]: true,
+          [syncHookProp434]: true,
+          [asyncHookProp434]: true,
         }),
         expect.anything(),
       );
--- a/packages/server/test/hooks.spec.ts
+++ b/packages/server/test/hooks.spec.ts
@@ -234,30 +234,33 @@ describe('Hooks', () => {
 
       const localClient = OpenFeature.getClient('merge-test', 'test', clientContext);
 
+      const syncVoidHook: Hook = {
+        before: () => {
+          // synchronous hook that doesn't modify context
+        },
+      };
+
+      const syncContextHook: Hook = {
+        before: () => {
+          return syncHookContext;
+        },
+      };
+
+      const asyncVoidHook: Hook = {
+        before: async () => {
+          // asynchronous hook that doesn't modify context
+          await Promise.resolve();
+        },
+      };
+
+      const asyncContextHook: Hook = {
+        before: () => {
+          return Promise.resolve(asyncHookContext);
+        },
+      };
+
       await localClient.getBooleanValue(FLAG_KEY, false, invocationContext, {
-        hooks: [
-          {
-            before: () => {
-              // synchronous hook that doesn't modify context
-            },
-          },
-          {
-            before: () => {
-              return syncHookContext;
-            },
-          },
-          {
-            before: async () => {
-              // asynchronous hook that doesn't modify context
-              await Promise.resolve();
-            },
-          },
-          {
-            before: async () => {
-              return Promise.resolve(asyncHookContext);
-            },
-          },
-        ],
+        hooks: [syncVoidHook, syncContextHook, asyncVoidHook, asyncContextHook],
       });
       expect(MOCK_PROVIDER.resolveBooleanEvaluation).toHaveBeenCalledWith(
         expect.anything(),
--- a/packages/server/src/hooks/hook.ts
+++ b/packages/server/src/hooks/hook.ts
@@ -2,6 +2,6 @@ import { BaseHook, EvaluationContext, FlagValue } from '@openfeature/core';
 
 export type Hook = BaseHook<
   FlagValue,
-  Promise<EvaluationContext | Promise<void>> | EvaluationContext | void,
+  Promise<EvaluationContext | void> | EvaluationContext | void,
   Promise<void> | void
 >;


packages/server/test/hooks.spec.ts
```diff
@@ -234,7 +238,23 @@ describe('Hooks', () => {
...
+          {
+            before: () => {
+              return syncHookContext;
+            },
+          },
+          {
+            before: async () => {
+              // asynchronous hook that doesn't modify context
+              await Promise.resolve();
+            },
```
beeme1mr, line 253: Unfortunately, this ran successfully before I made the change to the hook. It looks like our tsconfig ignores test files. I was able to confirm that the change worked with a manual test.

FYI, @toddbaert @lukas-reining

toddbaert: I'll see what I can do about that

packages/server/test/hooks.spec.ts
```diff
@@ -234,7 +238,23 @@ describe('Hooks', () => {
         hooks: [
           {
             before: () => {
-              return hookContext;
+              // synchronous hook that doesn't modify context
```
toddbaert, line 241:
```suggestion
              // synchronous hook that doesn't modify context
              return undefined;
```

There's no issue with our jest tsconfig. If you change `before` to `beefor`, you will get a compiler error.

This is something stranger. If you get rid of the implicit `void` return here, by, for instance applying this suggestion and returning `undefined`, you will see that without your fix, the compiler complains on the async hook. It's like the presence of the void return stops type checking for the other return values in the array.

beeme1mr: > Contextual typing with a return type of void does not force functions to not return something. Another way to say this is a contextual function type with a void return type (type voidFunc = () => void), when implemented, can return any other value, but it will be ignored.

https://www.typescriptlang.org/docs/handbook/2/functions.html#return-type-void

Perhaps we should switch from void to undefined in the hook types.

toddbaert: Maybe. I think void is just very good at communicating that a function only has side effects.

I understand `=> void` is pretty open in terms of what you can assign to it, what I don't understand is why returning void in a single element of the array prevents the compiler from checking the others.

toddbaert: I think removing void would also be an awkward breaking change. You'd need to add a return statement to all hooks, I think.

lukas-reining: Ya that would also be weird.
Semantically what we want to say is void.
But currently if there is a function with implicit void, we losse type checking completely for the array?

toddbaert: > But currently if there is a function with implicit void, we losse type checking completely for the array?

Implicit OR explicit void, yes. It's very strange. TS seems to lose all type-checking for return values for all elements if you include a single function with a void return value. It must be either a compiler bug or a very strange nuance I've never come across before.

We could put the hooks in different arrays, with the correct annotated type, and combine them perhaps... that would at least make the test valuable.

lukas-reining: For me it seems to be working as expected. Just that returning anything from the void hook is maybe unexpected but just allowed by typescript as you pointed out.

lukas-reining: https://github.com/microsoft/TypeScript/wiki/FAQ#why-are-functions-returning-non-void-assignable-to-function-returning-void

I think we can really not change that right? Good question what to do here as we are also consuming the result in the server SDK. In this case it could also be anything.

toddbaert: For me, if I add a void function to the array, I can then return _anything_ I want from _other_ functions in the array without TS complaining. If I remove the void return function, TS starts checking the other elements again.

lukas-reining: > I understand the assignability of void functions, I just don't understand why that assignability impacts other elements in the array.

Same. Still looking into it and it feels really like the compiler not getting it.

dyladan: What is the inferred type of `hooks` after you add the `void` return? I think it is most likely `{ before: () => void }` because it is the minimal type that satisfies the constraints. All of your promise versions are also assignable to that type. The type of the `getBooleanValue` argument is overridden by the inferred type. If you add an `as Parameters<typeof localClient["getBooleanValue"]>[3]` (this syntax might not be exactly right?) to the end of the object I would not be surprised if you got your type checking back.

dyladan: image

the popover hides the first array value but it is non-void

dyladan: I think instead of removing void, you should make the checking of return values more permissive. `return 1` shouldn't hurt anything, it just also doesn't really provide value. There's no real advantage to strictly checking the type at compile time. Who cares if the user returns `1`? Just ignore unexpected values.

lukas-reining: >  I think it is most likely { before: () => void }[] because it is the minimal type that satisfies the constraints

Ya, the only question is, why it "works" when the function is async. 

In the end I also think we should make it "ignore" unexpected types but for arbitrary objects that means we will have to check if it is an `EvaluationContext`.

lukas-reining: > Your point is correct: the entire reason for void's existence is to indicate the return value of a function isn't observed.

@toddbaert The following code from the **server** SDK consumes these types then.
```ts
  private async beforeHooks(hooks: Hook[], hookContext: HookContext, options: FlagEvaluationOptions) {
    for (const hook of hooks) {
      // freeze the hookContext
      Object.freeze(hookContext);

      // use Object.assign to avoid modification of frozen hookContext
      Object.assign(hookContext.context, {
        ...hookContext.context,
        ...(await hook?.before?.(hookContext, Object.freeze(options.hookHints))),
      });
    }

    // after before hooks, freeze the EvaluationContext.
    return Object.freeze(hookContext.context);
  }
```

If someone is able to return an arbitrary value, this value is assigned to the `EvaluationContext`.
Most times thats fine because the `EvaluationContext` accepts arbitrary keys and many value types, but something like this also works, so this method ends up in the context:

```ts
 hooks: [
          {
            before: () => {
              // synchronous hook that doesn't modify context
              return;
            },
          },
          {
            before: () => {
              return {
                myMethod: () => {
                  console.log('This ends in the EvaluationContext');
                },
              };
            },
          },
        ],
```

toddbaert: I realize that, but I think forcing users to return `undefined` or something like that would be too awkward. Hooks defined in-line in an array like this is probably not the most common way they're implemented, so I'm not sure it's worth it.

lukas-reining: Ya thats fine, just wanted to mention that we are not ignoring these unwanted values :)

beeme1mr: > I realize that, but I think forcing users to return `undefined` or something like that would be too awkward. Hooks defined in-line in an array like this is probably not the most common way they're implemented, so I'm not sure it's worth it.

There's an implicit `undefined` returned if nothing is explicitly returned in the function. I just tested this in the TypeScript playground and it behaved as expected.

```ts
type MyFun = () => undefined;

// invalid
const myFun: MyFun = () => {
    return "hi"
} 

// valid
const myFun: MyFun = () => {
    console.log("here")
} 
```

Neither are type issues when using void.

```ts
type MyFun = () => void;

// valid
const myFun: MyFun = () => {
    return "hi"
} 

// valid
const myFun: MyFun = () => {
    console.log("here")
} 
```

I actually think undefined may be more accurate because the return value is merged with the evaluation context.

dyladan: Think I agree with @beeme1mr if you're actually doing anything with the return value it is `undefined`. `void` implies you're not even checking if it exists which is why it's so permissive.

toddbaert: > Think I agree with @beeme1mr if you're actually doing anything with the return value it is `undefined`. `void` implies you're not even checking if it exists which is why it's so permissive.

Ya that makes sense - we _are_ actually checking the value. Honestly I don't think I've ever explicitly annotated a function to have an `undefined` return type, but I think when I briefly tried it in this case with the hook return type it required an explicit return statement.

beeme1mr: This feature was introduced in TypeScript 5.1. We're currently using TypeScript 4.7.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-1.html

toddbaert: > This feature was introduced in TypeScript 5.1. We're currently using TypeScript 4.7.
> 
> https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-1.html

Well that explains why you saw something different than me when you used the playground.

lukas-reining: > This feature was introduced in TypeScript 5.1. We're currently using TypeScript 4.7.

Ya this was also my initial thought. Should we then rely on this? Pre 5.1 consumers would have to return undefined explicitly then if they e.g. have a return statement for an early out.
Maybe that is still the best tradeoff right?🙂

toddbaert: I'm hesitant to upgrade our TS version. It can cause issues for consumers (as this change demonstrates). My vote would be to continue with `void` and extract the hooks into separate declarations to make the test valuable... but I'm open to other opinions.

I certainly think that if we upgrade to TS 5.1, `undefined` is a better return type, so it really just comes down to if we want to do that upgrade or not.

toddbaert (approved):

packages/server/test/hooks.spec.ts
lukas-reining: > I certainly think that if we upgrade to TS 5.1, undefined is a better return type, so it really just comes down to if we want to do that upgrade or not.

Sounds good!
But then we would have the problem that consumers pre 5.1 would get different behavior than we expect right?

toddbaert: Yes, this is why I think if we do this upgrade, we should do it in a few months once most people will have already updated, if we decide to do it at all.

lukas-reining (approved):
LGTM!

toddbaert:
@beeme1mr I made slight changes to the title. Hope you don't mind. Feel free to merge if you're OK with it.

PR #698, opened 2023-11-29
feat: suspense support, client scoping, context-sensitive re-rendering

toddbaert:
Adds a few react features:

- `<Suspense />` support: components using feature flags will trigger suspense for easy loaders/spinners
- Ability to specify name for provider scope: `<OpenFeatureProvider name="my-provider">`

> [!IMPORTANT]  
> Please see [here](https://github.com/open-feature/react-test-app/pull/2) for the latest changes to the demo application using these features (note that we won't be able to merge these demos until this is released).

> [!IMPORTANT]  
> Also note I've added no tests, which is certainly not my MO. I will add them to this PR once there's an agreement on this behavior and implementation.

gif from the demo app:

demo

:warning: I want to add another feature here to support re-rendering on context changes. That requires [this](https://github.com/open-feature/js-sdk/pull/731) to be merged.


packages/client/src/client/open-feature-client.ts
```diff
@@ -49,6 +49,10 @@ export class OpenFeatureClient implements Client {
     };
   }
 
+  get providerStatus(): ProviderStatus {
+    return this.providerAccessor()?.status || ProviderStatus.READY;
+  }
+
```
toddbaert, line 55: This is the only change outside react. It makes it possible to get he associated provider state immediately, which prevents us from doing some extra work in cases when the provider is already ready.

We could do this in a separate PR if we don't want to pollute the web-sdk changelog with a "feat" that doesn't really apply to it.

packages/react/src/provider.tsx
```diff
@@ -3,14 +3,15 @@ import { Client, OpenFeature } from '@openfeature/web-sdk';
 
 type ProviderProps = {
   client?: Client;
+  name?: string,
```
toddbaert, line 6: Allow the provider to be scoped by name.

packages/react/src/use-feature-flag.ts
```diff
@@ -1,20 +1,62 @@
-import { Client, EvaluationDetails, FlagValue, ProviderEvents } from '@openfeature/web-sdk';
-import { useEffect, useState } from 'react';
+import { Client, EvaluationDetails, FlagEvaluationOptions, FlagValue, ProviderEvents, ProviderStatus } from '@openfeature/web-sdk';
+import { Dispatch, SetStateAction, useEffect, useState } from 'react';
 import { useOpenFeatureClient } from './provider';
 
-export function useFeatureFlag<T extends FlagValue>(flagKey: string, defaultValue: T): EvaluationDetails<T> {
-  const [, setForceUpdateState] = useState({});
+type ReactFlagEvaluationOptions = {
```
toddbaert, line 5: I've defaulted both of these to true. I think that's what people would want. Reasoning:

- suspending if you don't have `<Suspend>` doesn't really have a negative impact
- if you are getting change events in the UI, you have probably set up your provider to do that, and therefor you probably want to update the UI. If your provider never gets readltime changes, then it's no problem.

packages/react/tsconfig.json
```diff
@@ -29,7 +29,10 @@
     // "rootDir": "./",                                  /* Specify the root folder within your source files. */
     "moduleResolution": "node", /* Specify how TypeScript looks up a file from a given module specifier. */
     // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
-    // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
+    "paths": {
+      "@openfeature/core": [ "../shared/src" ],
+      "@openfeature/web-sdk": [ "../client/src" ]
```
toddbaert, line 34: Ready from the src, not the dist, locally.

beeme1mr:
I know React uses the term `Provider` but we may want to come up with an alternative since a provider already has a specific meaning in OpenFeature.

image

toddbaert:
> I know React uses the term `Provider` but we may want to come up with an alternative since a provider already has a specific meaning in OpenFeature.

@beeme1mr I think I disagree. Conveniently the React provider is easily associated with a OpenFeature Provider via the name. I actually kinda thinks it works well. Moreover, in React, providers are a very explicit thing, and if we don't use that name I think people would be confused.

weyert:
> > I know React uses the term `Provider` but we may want to come up with an alternative since a provider already has a specific meaning in OpenFeature.
> 
> @beeme1mr I think I disagree. Conveniently the React provider is easily associated with a OpenFeature Provider via the name... so in some respects it actually works well to blur these lines. Moreover, in React, providers are a very explicit thing, and if we don't use that name I think people would be confused.
> 
> Maybe somebody else has an opinion on this point specifically. cc @lukas-reining @thomaspoignant @kinyoklion

No strong feelings about it but you could always call it `OpenFeatureContextProvider`

toddbaert:
> > > I know React uses the term `Provider` but we may want to come up with an alternative since a provider already has a specific meaning in OpenFeature.
> > 
> > 
> > @beeme1mr I think I disagree. Conveniently the React provider is easily associated with a OpenFeature Provider via the name... so in some respects it actually works well to blur these lines. Moreover, in React, providers are a very explicit thing, and if we don't use that name I think people would be confused.
> > Maybe somebody else has an opinion on this point specifically. cc @lukas-reining @thomaspoignant @kinyoklion
> 
> No strong feelings about it but you could always call it `OpenFeatureContextProvider`

Ya, I'd be fine with that, but unfortunately we also use `Context` in OpenFeature :grimacing:

lukas-reining:
> No strong feelings about it but you could always call it OpenFeatureContextProvider.
Also no strong feelings. 
`OpenFeatureProvider` feels the best to me. For people working on React it seems to be the most intuitive and as it is only used in the JSX context, I also feel that people wont confuse this for the OpenFeature provider concept.

beeme1mr:
You've convinced me. I'm fine with `OpenFeatureProvider`.

packages/client/src/client/open-feature-client.ts
beeme1mr: Should we add this to the spec? It may be useful on the server too.

What would you label it if you didn't go with feature?

packages/react/src/provider.tsx
beeme1mr: Could you add a JSDoc to both properties?

thomaspoignant:
I agree with @lukas-reining `OpenFeatureProvider` is simple enough.

lukas-reining (approved):
Change looks good to me and is a nice addition.
Left two comments with open questions.

packages/react/src/use-feature-flag.ts
```diff
@@ -1,20 +1,62 @@
...
 
-    // adding handlers here means that an update is triggered, which leads to the change directly reflecting in the UI
-    client.addHandler(ProviderEvents.Ready, forceUpdate);
-    client.addHandler(ProviderEvents.ConfigurationChanged, forceUpdate);
+    if (client.providerStatus !== ProviderStatus.READY) {
+      client.addHandler(ProviderEvents.Ready, forceUpdate); // update the UI when the provider is ready
+      if (defaultedOptions.suspend) {
+        suspend(client, updateState);
+      }
+    }
```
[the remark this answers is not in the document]
toddbaert: This is a good point, and I blocker I think. We may need some options or something. I will think more about it.

packages/client/src/client/open-feature-client.ts
toddbaert: > Should we add this to the spec? It may be useful on the server too.

I expected somebody to ask this question, and I don't feel too strongly about it. Honestly I've considered adding it before, but even with all the previous event stuff, this is the first time I've needed it. Since handlers run immediately, you can always just add one and get the state, and in most cases that's the better choice because it will cover the case where the provider is _about_ to become ready as well. In fact we might want to discourage it's use generally, to avoid confusing race conditions.

The only reason is really helpful here is because, in the case of UIs, the page is refreshed and components mount/unmount a lot. That would mean a lot of handler additions/removals. This helps use avoid un-needed ones. However, based on how it hasn't been needed thus far, I'm not 100% sure it's worth spec'ing. I would guess it will only be used in web (maybe android/iOS).

All that said, it's obviously easy to implement.

packages/react/src/use-feature-flag.ts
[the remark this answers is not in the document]
lukas-reining: I think we just have a problem with the SDK implementation.
https://openfeature.dev/specification/sections/events#requirement-533 tells that all handlers of an associated state must run also if they are attached after the provider reached this specific state.
The SDK currently only does this for the `READY` state. When we change/fix this, the problem would be gone as we could also just listen for error events. @toddbaert 
https://github.com/open-feature/js-sdk/blob/main/packages/client/src/client/open-feature-client.ts#L52-L64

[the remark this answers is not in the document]
toddbaert: I will fix this now.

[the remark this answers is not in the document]
toddbaert: https://github.com/open-feature/js-sdk/issues/700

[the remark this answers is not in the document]
toddbaert: Yes, I think for now it's OK. Of course if we decide we need different error semantics, we can always change things since this SDK is experimental for now.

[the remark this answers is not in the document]
lukas-reining: Sounds good!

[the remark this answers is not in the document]
toddbaert: fix: https://github.com/open-feature/js-sdk/pull/701

packages/react/src/provider.tsx
toddbaert: Done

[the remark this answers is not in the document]
toddbaert: done

packages/client/src/client/open-feature-client.ts
toddbaert: This was done in the SDK outside of this PR.

weyert (approved):
LGTM

toddbaert:
> > @toddbaert / @beeme1mr one question we had taking a look at this and thinking through the suspense behaviour, what happens when the context is updated? does the SDK go back into a "pending" state?
> 
> That's a very interesting question...
> 
> When the context is set, the SDK doesn't change the provider state (and has no such state itself) - but some providers might go into the `ProviderStatus.STALE` state while they await new flag values for the updated context. Others (particularly those that cache rulesets) may not.
> 
> This has a few implications, and I think we are missing some guidance and possibly some semantics around the `on context changed` handler and provider state/events. I will think about this some more today, and possibly create an issue and a PR in spec. Thanks for the thoughtful consideration @jonathannorris !

OK, so I've implemented another feature which allows for re-renders with the emission of `ProviderEvents.ContextChanged` events. This means that once the context is reconciled, we can re-render components, which is a _very_ important thing to have. Currently, there's no "pending" state; the aforementioned event is only emitted once the provider has fully reconciled it's new context (or it fails to, in which case an error is emitted).

packages/react/README.md
```diff
@@ -65,3 +77,93 @@ function Page() {
...
+
+This is analogous to:
+
+```ts
+OpenFeature.getClient('myClient');
+```
+
+### Re-rendering with Context Changes
+
+By default, if the OpenFeature [evaluation context](https://openfeature.dev/docs/reference/concepts/evaluation-context) is modified, components will be re-rendered.
```
Kavindu-Dodan, line 104: Minor - should we also highlight that react sdk implements static context paradigm [1] of the spec? 

[1] - https://openfeature.dev/specification/glossary#static-context-paradigm

toddbaert: Good call, yes!

packages/react/src/provider.tsx
```diff
@@ -2,15 +2,23 @@ import * as React from 'react';
...
 type ProviderProps = {
+  /**
+   * OpenFeature client to use.
+   */
   client?: Client;
+  /**
+   * The name of the client.
+   * @see OpenFeature.setProvider() and overloads.
+   */
+  clientName?: string,
```
beeme1mr, line 13: These are mutually exclusive. Perhaps we should make a union type.

beeme1mr (approved):

packages/react/src/provider.tsx
toddbaert: That's a good point.

packages/react/README.md
toddbaert: Done

packages/react/src/provider.tsx
toddbaert: Done

Issue #700, opened 2023-11-30
[BUG] client handlers do not run immediately if associated provider is in matching state.


toddbaert:
### Observed behavior

client handlers do not run immediately if associated provider is in matching state.

### Expected Behavior

client handlers should run immediately if associated provider is in matching state: https://openfeature.dev/specification/sections/events#requirement-533

PR #701, opened 2023-11-30
fix: handlers should run immediately

toddbaert:
* web-sdk handlers should run immediately if provider in matching state
* added tests for both server and client

Fixes: https://github.com/open-feature/js-sdk/issues/700

### DIFF
--- a/packages/client/test/events.spec.ts
+++ b/packages/client/test/events.spec.ts
@@ -504,28 +504,78 @@ describe('Events', () => {
 
   describe('Requirement 5.3.3', () => {
     describe('API', () => {
-      it('Handlers attached after the provider is already in the associated state, MUST run immediately.', (done) => {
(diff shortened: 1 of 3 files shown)


packages/client/test/events.spec.ts
```diff
@@ -504,28 +504,78 @@ describe('Events', () => {
...
+          OpenFeature.setProvider(clientId, provider);
+          expect(provider.initialize).not.toHaveBeenCalled();
+  
+          client.addHandler(ProviderEvents.Error, () => {
+            done();
+          });
+        });
 
-        client.addHandler(ProviderEvents.Ready, () => {
-          done();
```
toddbaert, line 528: The reason I missed this bug last time was because I only had a test for `READY` (which _did_ run immediately). I've added tests for all relevant states now, which failed before the fix.

lukas-reining (approved):
Looks good!

beeme1mr (approved):

PR #704, opened 2023-12-04
feat: add evaluation context management to the web SDK

### LINKED ISSUE #685: Allow evaluation context to be managed per provider in the web SDK
### Observed behavior

The lib allows for setting named providers and getting the client associated with that provider by name also, but removes (from `js-sdk` and `server-sdk`) the ability to set context on the client level (`client.setContext(...)`), rather than global level (`OpenFeature.setContext(...)`). This means that when I instantiate two different providers/clients and then set some context which is specific to the first one, followed by setting some different context that is specific to the second one, the second `setContext` overwrites the context set for the first provider/client.

### Expected Behavior

I have two proposals:

1. The `setContext` function should be added to the `Client` object, similar to how this is handled in the other JS SDKs.
2. The `OpenFeature.setContext` function could take a name property, similar to the `setProvider` function, so that the context is only associated with the provider and client of the same name.

### Steps to reproduce

See "Observed behavior" above

beeme1mr:
## This PR

- adds the ability to manage named context in the web SDK

### Related Issues

Fixes #685 

### Notes

This change requires additions to the spec before it can be merged. I'll mark this PR as a draft until the spec has been updated.

I also noticed that we currently use the term "named clients" to describe the scoped providers. I believe this terminology is confusing and made writing the JSDocs difficult because the client name shouldn't be important when setting context. I think we should consider using the term "Provider Namespace". In my opinion, this more accurately describes the behavior and could be introduced in a non-breaking way.

### DIFF
--- a/packages/client/src/open-feature.ts
+++ b/packages/client/src/open-feature.ts
@@ -65,8 +65,7 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider, Hook> impleme
   async setContext(clientName: string, context: EvaluationContext): Promise<void>;
   async setContext<T extends EvaluationContext>(nameOrContext: T | string, contextOrUndefined?: T): Promise<void> {
     const clientName = stringOrUndefined(nameOrContext);
-    const context =
-      objectOrUndefined<T>(nameOrContext) ?? objectOrUndefined(contextOrUndefined) ?? this._defaultContext;
+    const context = objectOrUndefined<T>(nameOrContext) ?? objectOrUndefined(contextOrUndefined) ?? {};
 
     if (clientName) {
       const provider = this._clientProviders.get(clientName);
@@ -142,7 +141,7 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider, Hook> impleme
         this._namedProviderContext.delete(clientName);
       }
     } else {
-      return this.setContext(this._defaultContext);
+      return this.setContext({});
     }
   }
 
--- a/packages/client/src/open-feature.ts
+++ b/packages/client/src/open-feature.ts
@@ -80,12 +80,12 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider, Hook> impleme
       const oldContext = this._context;
       this._context = context;
 
-      const providersWithoutContextOverride = [];
-      for (const [name, provider] of this._clientProviders.entries()) {
-        if (!this._namedProviderContext.has(name)) {
-          providersWithoutContextOverride.push(provider);
-        }
-      }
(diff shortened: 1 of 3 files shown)


toddbaert:
> adds the ability to manage context >> _named_ << in the web SDK

packages/client/src/open-feature.ts
```diff
@@ -38,26 +44,126 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider, Hook> impleme
...
+      this._context = context;
+
+      const providersWithoutContextOverride = [];
+      for (const [name, provider] of this._clientProviders.entries()) {
+        if (!this._namedProviderContext.has(name)) {
+          providersWithoutContextOverride.push(provider);
         }
-      }),
-    );
+      }
```
toddbaert, line 89: I'd prefer a reduce here, but that's just opinion.

packages/shared/src/open-feature.ts
```diff
@@ -16,8 +16,9 @@ export abstract class OpenFeatureCommonAPI<P extends CommonProvider = CommonProv
   protected abstract _defaultProvider: P;
   protected abstract readonly _events: GenericEventEmitter;
 
+  protected readonly _defaultContext = {};
   protected _hooks: H[] = [];
-  protected _context: EvaluationContext = {};
+  protected _context: EvaluationContext = this._defaultContext;
```
toddbaert, line 21: I might be better to leave this out since it can probably be done with just `{}`, and that way you've isolated your changes to just the web-sdk, but that's minor.

toddbaert (approved):
I don't see any issues witht this. Makes sense to me. Minor comments.

packages/shared/src/open-feature.ts
beeme1mr: Yeah, I had the same concern. I'll remove this.

packages/client/src/open-feature.ts
```diff
@@ -38,26 +44,125 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider, Hook> impleme
...
+   */
+  async clearContexts(): Promise<void> {
+    // Default context must be cleared first to avoid calling the onContextChange
+    // handler multiple times for named clients.
+    await this.clearContext();
+
+    const clearContextPromises: Promise<void>[] = [];
+    for (const name of this._clientProviders.keys()) {
+      clearContextPromises.push(this.clearContext(name));
+    }
```
lukas-reining, line 160: I would do a .map here, but as @toddbaert just an opinion.

lukas-reining (approved):
Looks good!

PR #718, opened 2023-12-12
feat: implement draft for a Nest.js SDK

### LINKED ISSUE #705: Nest.js SDK
We should build a [Nest.js Module](https://docs.nestjs.com/modules), as Nest.js is a really popular backend framework for JavaScript.

The following is the current idea of what feature could be interesting:

- [ ] Dynamic module with forRoot receiving a list of OpenFeature clients 
- [ ] A decorator to inject a client (named and unnamed)
- [ ] A decorator to inject a feature flag into a method (if good to implement)
- [ ] It should be possible to easily set the context of the feature flag evaluation in the scope of a request

Non-functional requirements:

- consistent terminology and naming with Web SDK and OpenFeature spec
- observe Nest.js idioms and patterns
- accurate peer dependency for the Web SDK

lukas-reining:
## This PR

This is a first draft of a Nest.js SDK.
Currently it only handles injecting clients and feature flags.
I also added a context factory for the FeatureFlag decorators to be able to use request information for the execution context. 

### Related Issues

Fixes #705 

### Notes

@toddbaert maybe we can do it as we did with the React SDK, merge this fast and release as experimental to let people experiment and try out?

### Follow-up Tasks

We will have to see if we want to stick with the injected feature flags being wrapped in an observable.
To me this feels the most idiomatic for Nest.js.

### How to test
Tests for the current functionality are included.

### DIFF
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -1,0 +1,103 @@
+<!-- markdownlint-disable MD033 -->
+<!-- x-hide-in-docs-start -->
+<p align="center">
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/open-feature/community/0e23508c163a6a1ac8c0ced3e4bd78faafe627c7/assets/logo/horizontal/white/openfeature-horizontal-white.svg" />
+    <img align="center" alt="OpenFeature Logo" src="https://raw.githubusercontent.com/open-feature/community/0e23508c163a6a1ac8c0ced3e4bd78faafe627c7/assets/logo/horizontal/black/openfeature-horizontal-black.svg" />
+  </picture>
+</p>
+
+<h2 align="center">OpenFeature Nest.js SDK</h2>
+
+<!-- x-hide-in-docs-end -->
+<!-- The 'github-badges' class is used in the docs -->
+<p align="center" class="github-badges">
+  <a href="https://github.com/open-feature/spec/releases/tag/v0.7.0">
+    <img alt="Specification" src="https://img.shields.io/static/v1?label=specification&message=v0.7.0&color=yellow&style=for-the-badge" />
+  </a>
+  <br/>
+  <a href="https://codecov.io/gh/open-feature/js-sdk">
+    <img alt="codecov" src="https://codecov.io/gh/open-feature/js-sdk/branch/main/graph/badge.svg?token=3DC5XOEHMY" />
+  </a>
+</p>
+<!-- x-hide-in-docs-start -->
+
+[OpenFeature](https://openfeature.dev) is an open specification that provides a vendor-agnostic, community-driven API for feature flagging that works with your favorite feature flag management tool.
+
+🧪 This is SDK is experimental.
+
+
+
+#### Here's a basic example of how ot use the current API with flagd.
+#### Registering the Nest.js SDK module in the App Module:
+
+```ts
+import { Module } from '@nestjs/common';
+import { FlagdProvider } from '@openfeature/flagd-provider';
+import { OpenFeatureModule } from '@openfeature/nestjs-sdk';
+
+@Module({
+  imports: [
+    OpenFeatureModule.forRoot({
+      defaultProvider: new FlagdProvider({
+        host: 'localhost',
+        port: 8013,
+        tls: false,
+      }),
+      providers: {
+        differentServer: new FlagdProvider({
+          host: 'different-server.com',
+          port: 8013,
+          tls: false,
+        }),
+      },
+    }),
+  ],
+})
+export class AppModule {}
+```
+
+#### Injecting the default and a named client into a service:
+```ts
+import { Controller, Get } from '@nestjs/common';
+import { map, Observable } from 'rxjs';
+import { BooleanFeatureFlag } from '@openfeature/nestjs-sdk';
+import { EvaluationDetails } from '@openfeature/server-sdk';
+
+@Controller()
+export class OpenFeatureController {
+  @Get('/welcome')
+  public async welcome(
+    @BooleanFeatureFlag({ flagKey: 'testBooleanFlag', defaultValue: false })
+    feature: Observable<EvaluationDetails<boolean>>,
+  ) {
+    return feature.pipe(
+      map((details) =>
+        details.value ? 'Welcome to this OpenFeature-enabled Nest.js app!' : 'Welcome to this Nest.js app!',
+      ),
+    );
+  }
+}
+```
+
+#### Injecting a feature flag into an endpoint handler method
+```ts
+import { Injectable } from '@nestjs/common';
+import { Client } from '@openfeature/server-sdk';
+import { OpenFeatureClient } from '@openfeature/nestjs-sdk';
+
+@Injectable()
+export class OpenFeatureTestService {
+  constructor(
+    @OpenFeatureClient() private defaultClient: Client,
+    @OpenFeatureClient({ name: 'differentServer' }) private namedClient: Client,
+  ) {}
+
+  public async getMessage() {
+    const companyName = await this.defaultClient.getStringValue('companyName', 'Unknown Company');
+    return `Hey User from ${companyName}`;
+  }
+}
+```
+
+
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -108,13 +108,13 @@ export class OpenFeatureController {
 ```ts
 import { Injectable } from '@nestjs/common';
 import { Client } from '@openfeature/server-sdk';
-import { OpenFeatureClient } from '@openfeature/nestjs-sdk';
+import { FeatureClient } from '@openfeature/nestjs-sdk';
 
 @Injectable()
 export class OpenFeatureTestService {
   constructor(
-    @OpenFeatureClient() private defaultClient: Client,
-    @OpenFeatureClient({ name: 'differentServer' }) private namedClient: Client,
+    @FeatureClient() private defaultClient: Client,
+    @FeatureClient({ name: 'differentServer' }) private namedClient: Client,
   ) {
   }
 
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -27,7 +27,7 @@ for feature flagging that works with your favorite feature flag management tool.
 
 🧪 This SDK is experimental.
 
-#### Here's a basic example of how to use the OpenFeature NestJS API with flagd.
+#### Here's a basic example of how to use the OpenFeature NestJS API with `InMemoryProvider`.
 
 #### Registering the Nest.js SDK module in the App Module:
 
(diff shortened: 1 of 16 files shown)


weyert (approved):
Looks similar to what I am having at work

lukas-reining:
Just also added the `contextFactory` parameter to the FeatureFlag decorators for dynamically setting the execution context based on request parameters such as Headers, IP, UserId, .... 
Maybe someone has a nicer way than this :)

packages/nest/README.md
```diff
@@ -0,0 +1,125 @@
...
+#### Injecting a feature flag with header value in evakuation context into an endpoint handler method
+
+```ts
+import { Controller, ExecutionContext, Get } from '@nestjs/common';
+import { map, Observable } from 'rxjs';
+import { BooleanFeatureFlag } from '@openfeature/nestjs-sdk';
+import { EvaluationDetails } from '@openfeature/server-sdk';
+import { Request } from 'express';
+
+function getContext(executionContext: ExecutionContext) {
```
luizgribeiro, line 70: Got it!

One concern here is how it can be propagated to different layers (modules/services) since.
`ExecutionContext` can only be obtained through  at controller level, right?

Maybe using [async local storage](https://docs.nestjs.com/recipes/async-local-storage) with [interceptors](https://docs.nestjs.com/interceptors) this can be a bit more ergonomic and accessible within different layers. 

Getting the type of execution context we could check for specific metadata fields (in http case headers) to check and configure an async local storage

packages/nest/README.md
```diff
@@ -0,0 +1,125 @@
...
+  <a href="https://codecov.io/gh/open-feature/js-sdk">
+    <img alt="codecov" src="https://codecov.io/gh/open-feature/js-sdk/branch/main/graph/badge.svg?token=3DC5XOEHMY" />
+  </a>
+</p>
+<!-- x-hide-in-docs-start -->
+
+[OpenFeature](https://openfeature.dev) is an open specification that provides a vendor-agnostic, community-driven API
+for feature flagging that works with your favorite feature flag management tool.
+
+🧪 This is SDK is experimental.
```
chenhunghan, line 28:
```suggestion
🧪 This SDK is experimental.
```

packages/nest/typedoc.json
```diff
@@ -0,0 +1,8 @@
+{
+    "$schema": "https://typedoc.org/schema.json",
+    "name": "@openfeature/react-sdk",
```
chenhunghan, line 3: probably need to update this
```suggestion
    "name": "@openfeature/nestjs-sdk",
```

packages/nest/tsconfig.json
```diff
@@ -0,0 +1,117 @@
...
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
+    /* Language and Environment */
+    "target": "ES2015",
+    /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
+    "lib": [
+      "ES2020",
+      "DOM"
+    ],
```
chenhunghan, line 17: Probably don't need `DOM`, same as server-sdk https://github.com/open-feature/js-sdk/blob/main/packages/server/tsconfig.json#L14
```suggestion
    "lib": [
      "ES2020"
    ],
```

packages/nest/README.md
lukas-reining: My feeling is that the FeatureFlag decorators only make sense for the endpoint handlers.
In the services, it feels like one would use the injected clients. To get request information I would assume one would use request scopes services that build the context themselves.

> Maybe using [async local storage](https://docs.nestjs.com/recipes/async-local-storage) with [interceptors](https://docs.nestjs.com/interceptors) this can be a bit more ergonomic and accessible within different layers.

You could implement a PR for this or maybe try it out. I am not sure how I would do this or if this could be made with good ergonomics.

packages/nest/src/open-feature.module.ts
```diff
@@ -0,0 +1,45 @@
+import { DynamicModule, Module, FactoryProvider as NestValueProvider } from '@nestjs/common';
+import { Client, OpenFeature, Provider } from '@openfeature/server-sdk';
+
+@Module({})
+export class OpenFeatureModule {
+  static forRoot(options: OpenFeatureModuleOptions): DynamicModule {
+    const clientValueProviders: NestValueProvider<Client>[] = [];
+
```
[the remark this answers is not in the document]
lukas-reining: Not sure, as the SDKs themselves also do not throw, but rather use the No-Op Provider.
If we throw or enforce a provider, we would behave differently than the SDKs I would say.
But ya, using the No-Op Provider does not make much sense but still I would rather behave like the SDK.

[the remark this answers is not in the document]
luizgribeiro: I see.

What do you think about configuring No-Op as default provider if no named provider or default provider is passed?
It might not throw at this point, but if the provider injection decorator is being used the app will break anyway

packages/nest/README.md
```diff
@@ -0,0 +1,125 @@
...
+  </a>
+</p>
+<!-- x-hide-in-docs-start -->
+
+[OpenFeature](https://openfeature.dev) is an open specification that provides a vendor-agnostic, community-driven API
+for feature flagging that works with your favorite feature flag management tool.
+
+🧪 This SDK is experimental.
+
+#### Here's a basic example of how to use the current API with flagd.
```
toddbaert, line 30:
```suggestion
#### Here's a basic example of how to use the OpenFeature NestJS API with flagd.
```

packages/nest/README.md
```diff
@@ -0,0 +1,125 @@
...
+        port: 8013,
+        tls: false,
+      }),
+      providers: {
+        differentServer: new FlagdProvider({
+          host: 'different-server.com',
+          port: 8013,
+          tls: false,
+        }),
+      },
```
toddbaert, line 53: flagd is OK, but I think the in-memory provider would be a better example.

packages/nest/README.md
toddbaert: > Maybe using async local storage with interceptors this can be a bit more ergonomic and accessible within different layers.

I really like the decorator approach, and I think it still would be cool to have them in lower-level modules. I think this might be possible in all layers if you used nest [factory dependency](https://docs.nestjs.com/fundamentals/custom-providers#factory-providers-usefactory) which provided the context, is injected in `Scope.REQUEST`, and itself received the `req` dependency. In my experience with NestJS, whenever I wanted request data lower down the stack, I'd just add `Scope.REQUEST` dependencies to the lower modules and use it there to avoid passing whatever data I need down the entire class-stack manually. As long as the objects you're constructing can be created per-request with low memory cost it works well.

I'm not exactly sure how much of these we'd want to implement, and how much we'd just give examples for and leave to the consumer.

toddbaert:
> @toddbaert maybe we can do it as we did with the React SDK, merge this fast and release as experimental to let people experiment and try out?

Yep that works for me! You'll have to manually add an entry in https://github.com/open-feature/js-sdk/blob/main/.release-please-manifest.json though.

packages/nest/src/open-feature.module.ts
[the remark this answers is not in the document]
lukas-reining: > What do you think about configuring No-Op as default provider if no named provider or default provider is passed?

Yes thats good. We should just always inject the default provider regardless of what it is set to.
Added that in the last commit.

packages/nest/README.md
lukas-reining: Okay I can do this, just stole this from React 😄

lukas-reining: Changed

packages/nest/package.json
```diff
@@ -0,0 +1,63 @@
...
+  ],
+  "author": "",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/open-feature/js-sdk/issues"
+  },
+  "homepage": "https://github.com/open-feature/js-sdk#readme",
+  "peerDependencies": {
+    "@nestjs/common": "^8.0.0 || ^9.0.0 || ^10.0.0",
+    "@openfeature/server-sdk": ">=0.4.0",
```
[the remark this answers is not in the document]
lukas-reining: Ya thats fine to me

packages/nest/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "nest"
+  ],
+  "author": "",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/open-feature/js-sdk/issues"
+  },
+  "homepage": "https://github.com/open-feature/js-sdk#readme",
+  "peerDependencies": {
+    "@nestjs/common": "^8.0.0 || ^9.0.0 || ^10.0.0",
```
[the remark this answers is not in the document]
lukas-reining: It should thats right.

lukas-reining:
> Yep that works for me! You'll have to manually add an entry in https://github.com/open-feature/js-sdk/blob/main/.release-please-manifest.json though.

Great, changed that.

packages/nest/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './open-feature.module';
```
toddbaert, line 1: You might want to consider adding something like this: https://github.com/open-feature/js-sdk/blob/c7098c365a3723c12eb3b1b35b1f5c58e0cc231f/packages/react/src/index.ts#L3C1-L3C1

This allows users to import all the things they need from the Nest SDK (it just re-exports the exports of the peer). There is no runtime impact, all the same modules are resolved, etc, it's just a nice experience to be able to import everything from the "highest level" sdk.

toddbaert (approved):
Looks good to me as an experimental release!

I added a couple more suggestions.

packages/nest/tsconfig.json
```diff
@@ -0,0 +1,116 @@
...
+    // "noLib": true,                                    /* Disable including any library files, including the default lib.d.ts. */
+    // "useDefineForClassFields": true,                  /* Emit ECMAScript-standard-compliant class fields. */
+    /* Modules */
+    "module": "ES2020",
+    /* Specify what module code is generated. */
+    // "rootDir": "./",                                  /* Specify the root folder within your source files. */
+    "moduleResolution": "node",
+    /* Specify how TypeScript looks up a file from a given module specifier. */
+    // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
+    // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
```
toddbaert, line 36: You might want to add the sources here in paths, as we did in a react/web. This will tell TSC to import from source files instead of dist. This makes the local dev experience must nicer and less confusing.

toddbaert: I would suggest doing it for `"@openfeature/core"` and `"@openfeature/server-sdk"` so that when these modules are changed, the changes can be used here without rebuilding.

packages/nest/src/index.ts
lukas-reining: Ya, I thought we might not need, other Nest Packages also do not do this but I will add it.

packages/nest/tsconfig.json
lukas-reining: Totally, forgot that and added it now.

packages/nest/README.md
lukas-reining: I would propose that we make a new issue for this, that we can work on then :)

lukas-reining:
I would merge this now, so we can work together on different questions.
@luizgribeiro lets open an issue for all open questions, e.g. the feature flag injection to services.

PR #733, opened 2023-12-15
temp: add custom decorators and context interceptor

luizgribeiro:
## This PR

Has experiments on decorators ergonomics and context propagation for Nest.js SDK

### Notes

This is WIP and shouldn't be merged.

### Follow-up Tasks

Define how:
- Context Should be propagated on a Nest.js app
- OpenFeature Nestjs decorators API should look like

### How to test
N/A (yet)

### DIFF
--- a/packages/nest/src/feature.decorator.ts
+++ b/packages/nest/src/feature.decorator.ts
@@ -1,7 +1,9 @@
 import { createParamDecorator, ExecutionContext, Inject } from '@nestjs/common';
 import { EvaluationContext, FlagValue, JsonValue, OpenFeature } from '@openfeature/server-sdk';
 import { getOpenFeatureClientToken } from './open-feature.module';
-import { from } from 'rxjs';
+import { from, range } from 'rxjs';
+import { asyncLocalStorage } from './feature.interceptor';
+import 'reflect-metadata';
 
 interface FeatureClientProps {
   name?: string;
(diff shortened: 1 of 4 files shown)


luizgribeiro:
Hey @lukas-reining, I've expressed a few ideas and possibilities on this branch.
I started by the context interceptor and discovered that the current decorators would not work with service's methods.
So, a new pair of params decorator and method decorator was created to provide a simpler API to OF lookups and it already uses injected context. 
The sad thing is that it doesn't work with controllers/routes at this point. I'll probably need to dive deeper into Nest decorators to come up with a way to do this (if possible).

Let me know what you think :wink:

packages/nest/src/feature.decorator.ts
```diff
@@ -17,46 +19,79 @@ interface FeatureProps<T extends FlagValue> {
   contextFactory?: (executionContext: ExecutionContext) => EvaluationContext | undefined;
 }
 
+/**
+ *
+ * @param clientName
+ * @param context
+ */
 function getClientForEvaluation(clientName?: string, context?: EvaluationContext) {
+  //is context thread safe? setting it here might result in context from different requests mixing up
```
lukas-reining, line 28: Yes, getting a named client creates a new instance.
If a provider is bound to the name, all clients with this name get it, but the context is per-instance.

packages/nest/test/test-app.ts
```diff
@@ -81,15 +91,21 @@ export class OpenFeatureController {
...
     })
     feature: Observable<EvaluationDetails<number>>,
   ) {
     return feature.pipe(map((details) => this.testService.serviceMethod(details)));
   }
+
+  @Get('/dynamic-context/service')
+  public handleDynamicContextForService() {
+    return this.testService.injectedFlagWithContextMethod("dummy");
+  }
```
lukas-reining, line 107: This is, why I think service method decorators could be hard to do with good ergonomics.
Having these makes callers of the service methods having to provide "dummy" arguments as service methods are typically called by a controller or other services, while controllers are called by Nest.
My feeling is that FeatureFlag injection will only work good for injecting FeatureFlags into request scoped services via the constructor because this is the only guaranteed to be called by Nest.

packages/nest/src/feature.interceptor.ts
```diff
@@ -0,0 +1,25 @@
...
+export class OpenFeatureContextInterceptor implements NestInterceptor {
+  intercept(context: ExecutionContext, next: CallHandler): Observable<any> {
+
+    if (context.getType() === 'http') {
+      const req = context.switchToHttp().getRequest();
+      const targetingKey = req.get('x-user-id');
+      if (targetingKey) {
+
+        return asyncLocalStorage.run({targetingKey}, () => next.handle());
+      }
```
lukas-reining, line 18: Here we would have go get a mechanism choose what is injected as context.

lukas-reining (commented):
The asyncLocalStorage is a good idea to get the context down.
We will really have to think about where we can inject the flags though, as my feeling is that service methods do not work well.
What do you think @luizgribeiro?🙂

packages/nest/test/test-app.ts
luizgribeiro: I understand your concern. It's not really a great experience from the caller side.

> My feeling is that FeatureFlag injection will only work good for injecting FeatureFlags into request scoped services via the constructor because this is the only guaranteed to be called by Nest.
I'm not a big fan of request scoped services but we cant try it and see what we get.

packages/nest/src/feature.interceptor.ts
luizgribeiro: I've been thinking about using a context key -> header mapping on instantiation for this.
Something like:

```ts
new Interceptor({targetKey: 'x-user-id', tenant: 'x-tenat-id'});
```

packages/nest/test/test-app.ts
lukas-reining: > I'm not a big fan of request scoped services but we cant try it and see what we get.

Generally I thinks they can be a good option, but we could also just decide that services just can not get injected feature flags nicely. 
I dont know, we should play around with and see if we find a good option or if it is not possible to do nicely.

packages/nest/src/feature.interceptor.ts
lukas-reining: This would be good, but e.g. other request info like the IP address could also be needed.

toddbaert: In general, I think a functional approach would be more flexible. Instead of a key/value config, you could accept a function that takes a request, and run that... then users can extract whatever they want.

lukas-reining: This would be something like the `contextFactory` that we have in the current draft.

lukas-reining:
Should we close this because of #736 @luizgribeiro?

luizgribeiro:
Just for sake of completeness, this one is being closed since context propagation is being handled on #736 and I could not find a nice and ergonomic way to inject flag evaluation results as service parameters.

PR #736, opened 2023-12-17
feat: context propagation for nestjs

lukas-reining:
## This PR

Implements context propagation in the Nest.js SDK.
This relates to the discussions from https://github.com/open-feature/js-sdk/pull/733#discussion_r1427747322.

Open questions are:
- How to scope the `contextFactory`?  It seems that we should be able to define a common and client specific context factories.
- Should the `AsyncLocalStorage` be used in the `FeatureFlagDecorators` too as it is in this draft?
- Should we merge the injected `OpenFeatureClient` and the `OpenFeatureContextService` so that you do not have to inject both into the consuming services as we do in the `TestService`?

@luizgribeiro this is the draft for what we discussed about, what do you think?


packages/nest/package.json
```diff
@@ -60,5 +60,8 @@
     "@types/supertest": "^2.0.16",
     "express": "^4.18.2",
     "supertest": "^6.3.3"
+  },
+  "dependencies": {
+    "nestjs-cls": "^3.6.0"
```
luizgribeiro, line 65: Looks like cls and express are not being used. We should remove it later.

packages/nest/test/open-feature-sdk.spec.ts
```diff
@@ -160,4 +161,16 @@ describe('OpenFeature SDK', () => {
       expect(evaluationSpy).toHaveBeenCalledWith('testBooleanFlag', false, { targetingKey: '123' }, {});
     });
   });
+
+  describe('evaluation context service should', () => {
+    it('inject the evaluation context from contex factory', async function () {
+      const evaluationSpy = jest.spyOn(defaultProvider, 'resolveBooleanEvaluation');
+      await supertest(app.getHttpServer())
+        .get('/dynamic-context-in-service')
+        .set('x-user-id', 'dynamic-user')
```
luizgribeiro, line 170: I see x-user-id being used multiple times.
We would probably benefit from sketching conventions about OF context propagation between services to make flag evaluation consistent and this SDK might be a good starting point for this.
Marking @toddbaert for visibility

luizgribeiro (commented):
That's great!

Multiple relevant ideas that were implemented in a nice way.
I still think about the following points though:
* Should we always use a global interceptor? I think this might be the default but it would be good to make it possible to disable it and provide users the ability to use the interceptor at Controller level or route
* Should we provide a default interceptor that looks for specific (maybe spec wise) headers?

packages/nest/test/open-feature-sdk.spec.ts
lukas-reining: What exactly do you mean by this?

packages/nest/test/test-app.ts
```diff
@@ -8,19 +7,27 @@ import {
...
+    @FeatureClient({ name: 'namedClient' }) public namedClient: OpenFeatureClient,
+    @FeatureContextService() public contextService: OpenFeatureContextService,
   ) {}
 
   public async serviceMethod(flag: EvaluationDetails<FlagValue>) {
     return flag.value;
   }
+
+  public async serviceMethodWithDynamicContext(flagKey: string): Promise<boolean> {
+    return this.defaultClient.getBooleanValue(flagKey, false, this.contextService.getContext());
```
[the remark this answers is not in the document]
lukas-reining: > My ideia here is that we don´t need to provide the injected context explicitly.

How would you do this? Merging the OF Client and the ContextService?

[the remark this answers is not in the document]
lukas-reining: > I was thinking about passing additional context info here. Some date might not be available at interceptor level. It would still work by using destructuring the context from the interceptor.

Where would you pass this data?

packages/nest/src/open-feature.module.ts
```diff
@@ -25,11 +49,31 @@ export class OpenFeatureModule {
       });
     }
 
+    const alsProvider: ValueProvider = {
+      provide: AsyncLocalStorage,
```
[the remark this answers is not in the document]
lukas-reining: Ya totally!
I fixed this, by using a Symbol for this. 🙂

lukas-reining:
> Should we always use a global interceptor? I think this might be the default but it would be good to make it possible to disable it and provide users the ability to use the interceptor at Controller level or route

Ya, this may be interesting for users! Maybe you can sketch out something?

> Should we provide a default interceptor that looks for specific (maybe spec wise) headers?

Not sure if this makes sense. I think the data used (from headers, tokens,...) is to diferent from case to case to provide a good default.
Do you have anything concrete in mind?

packages/nest/src/feature.decorator.ts
```diff
@@ -1,62 +1,52 @@
...
 
 export const StringFeatureFlag = createParamDecorator(
-  (
-    { clientName, flagKey, defaultValue, context, contextFactory }: FeatureProps<string>,
-    executionContext: ExecutionContext,
-  ) => {
+  ({ clientName, flagKey, defaultValue, context }: FeatureProps<string>) => {
     const client = getClientForEvaluation(clientName, context);
-    return from(client.getStringDetails(flagKey, defaultValue, contextFactory?.(executionContext)));
+    return from(client.getStringDetails(flagKey, defaultValue, SharedAsyncLocalStorage.getStore()?.context));
```
beeme1mr, line 36: You could use the `setTransactionContextPropagator` to automatically extract transaction scoped context within the SDK itself. You can see an example in the playground.

https://github.com/open-feature/playground/blob/main/packages/app/src/app/app.module.ts#L46

You'll just need to create a context propagator [like this](https://github.com/open-feature/playground/blob/main/packages/openfeature-extra/src/lib/transaction-context/async-local-storage.ts).

beeme1mr (commented):
This is looking really nice. I'm just wondering if we should use the transaction propagator that's already built into the server SDK.

packages/nest/src/feature.decorator.ts
lukas-reining: Oh ya this could be a cool idea!
I thought about it, but thought that this API is not continued.
I will make a PR using this tomorrow or so to try I really like it!

beeme1mr: It's marked as experimental but we could remove that disclaiming if we start using it in this SDK.

lukas-reining: Cool, I will give it a try and add it tomorrow or so🙃

packages/nest/test/test-app.ts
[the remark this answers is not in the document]
luizgribeiro: I've opened a PR to this branch to give an example.
Here's it https://github.com/open-feature/js-sdk/pull/738

packages/nest/test/open-feature-sdk.spec.ts
luizgribeiro: If we have a convention for open feature context building and passing in requests we can propagate it across multiple services to make flag resolution consistent without much effort. 
Eg: Considering that service A (Node) has access to userId/tragetingKey, it can then forward this info as request metadata to services B (Golang), C (Java).
A -> B
   -> C
   
 I don´t  think I made myself clear enough though. Did you get the idea?

packages/nest/src/feature.decorator.ts
luizgribeiro: This looks great!

[It's merging context](https://github.com/open-feature/js-sdk/blob/main/packages/server/src/client/open-feature-client.ts#L202) so, [this PR](https://github.com/open-feature/js-sdk/pull/738) is not necessary :rocket:

lukas-reining: @beeme1mr @luizgribeiro 
I changed to choose the context propagator for all context propagation.
This makes the implementation way cleaner, I really like it. 
I also removed the ContextService as I think this would not be needed in this implementation.
What do you think?

packages/nest/src/evaluation-context-propagator.ts
```diff
@@ -0,0 +1,14 @@
+import { TransactionContextPropagator } from '@openfeature/server-sdk';
+import { AsyncLocalStorage } from 'async_hooks';
+import { EvaluationContext } from '@openfeature/web-sdk';
```
beeme1mr, line 3:
```suggestion
import { EvaluationContext } from '@openfeature/server-sdk';
```

packages/nest/src/feature.decorator.ts
beeme1mr: It looks good so far.

packages/nest/src/feature.decorator.ts
```diff
@@ -14,49 +14,36 @@ interface FeatureProps<T extends FlagValue> {
...
 export const BooleanFeatureFlag = createParamDecorator(
-  (
-    { clientName, flagKey, defaultValue, context, contextFactory }: FeatureProps<boolean>,
-    executionContext: ExecutionContext,
-  ) => {
-    const client = getClientForEvaluation(clientName, context);
-    return from(client.getBooleanDetails(flagKey, defaultValue, contextFactory?.(executionContext)));
+  ({ clientName, flagKey, defaultValue }: FeatureProps<boolean>) => {
+    const client = getClientForEvaluation(clientName);
+    return from(client.getBooleanDetails(flagKey, defaultValue));
```
luizgribeiro, line 26:
```suggestion
  ({ clientName, flagKey, defaultValue, context }: FeatureProps<boolean>) => {
    const client = getClientForEvaluation(clientName, context);
    return from(client.getBooleanDetails(flagKey, defaultValue));
```

luizgribeiro (commented):
It looks great!

Do you have anything else in mind for this? If not, I can start to work on controller interceptors based on this

lukas-reining:
>  If not, I can start to work on controller interceptors based on this

What interceptors are you thinking of @luizgribeiro?

luizgribeiro:
Oh, this one @lukas-reining, my bad.

> > Should we always use a global interceptor? I think this might be the default but it would be good to make it possible to disable it and provide users the ability to use the interceptor at Controller level or route
> Ya, this may be interesting for users! Maybe you can sketch out something?

lukas-reining:
> Oh, this one @lukas-reining, my bad.

Ah yes, this would be cool!

luizgribeiro:
I couldn´t find a straightforward way of doing it at module level. By the docs and from [this issue](https://github.com/nestjs/nest/issues/1488) I'd say it's not native to Nestjs.

I've opened the #739 for this branch to make it possible to switch off the global OF context interceptor

packages/nest/test/open-feature-sdk.spec.ts
lukas-reining: Baggage could be a cool solution for context propagation.
To me, this would be a special version of a `contextFactory`, that we could provide later if we spec the usage of something like baggage I would say. I feel like this is too early to include into this SDK.
What do you think @luizgribeiro?

lukas-reining:
I feel like we could bring that into the experimental SDK now.
What do you think @luizgribeiro @beeme1mr @toddbaert?

beeme1mr:
Sounds good to me. I would like to try it out in the playground app to see how it feels to work with.

lukas-reining:
> Sounds good to me. I would like to try it out in the playground app to see how it feels to work with.

Are you doing it? Or should I make a PR for this?

beeme1mr:
Feel free to get it started if you have time. I've been experimenting with a cli to generate type safe clients. I would like to get that in a better place before shifting my focus.

packages/nest/test/open-feature-sdk.spec.ts
luizgribeiro: > I feel like this is too early to include into this SDK.
I agree.

I'd say that since we are dealing with a context factory it will probably be affected in the future is baggage is included in OF. It would be necessary to provide a way to read it and insert in current context (may a higher level contextFactory that also runs the user provided context factory :thinking:).

It would also be necessary to use the provided baggage in different interfaces with other services (HTTP, gRPC, etc). I think we might learn a lot from how Otel is handling this, but it's out of scope from this SDK perspective.

luizgribeiro (approved):

beeme1mr (approved):
Please address linting issues before merging. It would also be nice to have an example included in the NestJS readme.

toddbaert (approved):
I haven't fully dug into this yet (still catching up from holidays) but I'm dropping an approval since it's experimental!

PR #738, opened 2023-12-19
Feat/nestjs auto context propagation

luizgribeiro:
## This PR

Shows an idea of how to pass context implicitly on the Nestjs SDK.

### Notes

There's probably a cleaner way to do this

### Follow-up Tasks

- Provide a OF Client Service (Nest) that has all OF client capabilities


luizgribeiro:
Closing this in favor of #736

PR #739, opened 2023-12-26
feat: add optout global context interceptors

luizgribeiro:
## This PR

- Makes it possible to deactivate the global OF context interceptor

### Notes

It currently supports passing a default context factory that is injected whenever the interceptor class is passed to `UseInterceptors`.


lukas-reining (approved):
This looks good to me, I will merge it :)

PR #750, opened 2024-01-10
docs: Nest SDK

luizgribeiro:
## This PR
- Adds docs to the NestJS SDK.

### Notes

I think we're probably reaching a stable point for the NestJS SDK.
It would be great to have some feedback on the docs before it.

### Follow-up Tasks
- Blog post & announcement regarding the new SDK (maybe?)

### DIFF
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -63,7 +63,7 @@ yarn add @openfeature/nestjs-sdk
 
 ### Usage
 
-The example bellow shows how to use the `OpenFeatureModule` with OpenFeature's `InMemoryProvider`.
+The example below shows how to use the `OpenFeatureModule` with OpenFeature's `InMemoryProvider`.
 
 ```ts
 import { Module } from '@nestjs/common';
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -142,6 +142,6 @@ export class OpenFeatureTestService {
 
 ### Flag evaluation context injection
 
-Whenever a flag evaluation occours, context can be provided with information like user e-mail, role, targeting key, etc in order to trigger specific evaluation rules or logic. The `OpenFeatureModule` provides a way to configure context for each request using the `contextFactory` option.
+Whenever a flag evaluation occurs, context can be provided with information like user e-mail, role, targeting key, etc in order to trigger specific evaluation rules or logic. The `OpenFeatureModule` provides a way to configure context for each request using the `contextFactory` option.
 The `contextFactory` is ran in a NestJS interceptor scope to configure the evaluation context and than it is used in every flag evaluation related to this request.
 By default the interceptor is cofigured globally, but it can be changed by setting the `useGlobalInterceptor` to `false`. In this case it is still possible to configure a `contextFactory` that can be injected into route, module or controller bound interceptors.
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -56,11 +56,11 @@ It's main capabilities are:
 npm install --save @openfeature/nestjs-sdk
 ```
 
-#### yarn
+#### Required peer dependencies
 
-```sh
-yarn add @openfeature/nestjs-sdk @openfeature/server-sdk @openfeature/core
-```
+The OpenFeature SDK is required as peer dependency.
+
+The minimum required version of `@openfeature/server-sdk` currently is `1.7.5`.
 
 ### Usage
 
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -31,7 +31,9 @@ for feature flagging that works with your favorite feature flag management tool.
 
 ## Overview
 
-The OpenFeature NestJS SDK is a package that provides a NestJS wrapper for the [OpenFeature Server SDK](). It's main capabilities are:
+The OpenFeature NestJS SDK is a package that provides a NestJS wrapper for the [OpenFeature Server SDK]().
+
+It's main capabilities are:
 - Provide a NestJS global module to simplify OpenFeature configuration and usage within NestJS;
 - Supply custom parameter decorators (for controllers), that inject observables with feature flags resolution details;
 - Inject context for flag evaluation seamlessly using NestJS interceptors
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -60,7 +60,12 @@ npm install --save @openfeature/nestjs-sdk
 
 #### Required peer dependencies
 
-The OpenFeature SDK is required as peer dependency.
+The following list contains the peer dependencies of `@openfeature/nestjs-sdk` with it's expected and compatible versions:
+
+* `@openfeature/server-sdk`: >=1.7.5
+* `@nestjs/common`: ^8.0.0 || ^9.0.0 || ^10.0.0
+* `@nestjs/core`: ^8.0.0 || ^9.0.0 || ^10.0.0
+* `rxjs`: ^6.0.0 || ^7.0.0
 
 The minimum required version of `@openfeature/server-sdk` currently is `1.7.5`.
 
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -65,7 +65,7 @@ The following list contains the peer dependencies of `@openfeature/nestjs-sdk` w
 * `@openfeature/server-sdk`: >=1.7.5
 * `@nestjs/common`: ^8.0.0 || ^9.0.0 || ^10.0.0
 * `@nestjs/core`: ^8.0.0 || ^9.0.0 || ^10.0.0
-* `rxjs`: ^6.0.0 || ^7.0.0
+* `rxjs`: ^6.0.0 || ^7.0.0 || ^8.0.0
 
 The minimum required version of `@openfeature/server-sdk` currently is `1.7.5`.
 
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -15,16 +15,14 @@
   <a href="https://github.com/open-feature/spec/releases/tag/v0.7.0">
     <img alt="Specification" src="https://img.shields.io/static/v1?label=specification&message=v0.7.0&color=yellow&style=for-the-badge" />
   </a>
-  <br/>
-  <a href="https://codecov.io/gh/open-feature/js-sdk">
-    <img alt="codecov" src="https://codecov.io/gh/open-feature/js-sdk/branch/main/graph/badge.svg?token=3DC5XOEHMY" />
-  </a>
-  <br/>
     <!-- x-release-please-start-version -->
   <a href="https://github.com/open-feature/js-sdk/releases/tag/nestjs-sdk-v0.1.0">
     <img alt="Release" src="https://img.shields.io/static/v1?label=release&message=v0.1.0&color=blue&style=for-the-badge" />
   </a>
   <!-- x-release-please-end -->
+  <a href="https://codecov.io/gh/open-feature/js-sdk">
+    <img alt="codecov" src="https://codecov.io/gh/open-feature/js-sdk/branch/main/graph/badge.svg?token=3DC5XOEHMY" />
+  </a>
 </p>
 <!-- x-hide-in-docs-start -->
 
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -25,12 +25,43 @@
 [OpenFeature](https://openfeature.dev) is an open specification that provides a vendor-agnostic, community-driven API
 for feature flagging that works with your favorite feature flag management tool.
 
+<!-- x-hide-in-docs-end -->
+
 🧪 This SDK is experimental.
 
-#### Here's a basic example of how to use the OpenFeature NestJS API with `InMemoryProvider`.
+## Overview
 
-#### Registering the Nest.js SDK module in the App Module:
+The OpenFeature NestJS SDK is a package that provides a NestJS wrapper for the [OpenFeature Server SDK](). It's main capabilities are:
+- Provide a NestJS global module to simplify OpenFeature configuration and usage within NestJS;
+- Supply custom parameter decorators (for controllers), that inject observables with feature flags resolution details;
+- Inject context for flag evaluation seamlessly using NestJS interceptors
+- Provide decorators for OpenFeature Client injection into NestJS services and controllers
+- TODO: map other features
 
+## 🚀 Quick start
+
+### Requirements
+
+- Node.js version 16+
+- NestJS
+
+### Install
+
+#### npm
+
+```sh
+npm install --save @openfeature/nestjs-sdk
+```
+
+#### yarn
+
+```sh
+yarn add @openfeature/nestjs-sdk
+```
+
+### Usage
+
+The example bellow shows how to use the `OpenFeatureModule` with OpenFeature's `InMemoryProvider`.
 ```ts
 import { Module } from '@nestjs/common';
 import { FlagdProvider } from '@openfeature/flagd-provider';
@@ -46,11 +77,6 @@ import { InMemoryProvider } from '@openfeature/web-sdk';
           variants: { default: true },
           disabled: false
         },
-        companyName: {
-          defaultVariant: 'default',
-          variants: { default: "BigCorp" },
-          disabled: false
-        }
       }),
       providers: {
         differentProvider: new InMemoryProvider()
@@ -61,7 +87,7 @@ import { InMemoryProvider } from '@openfeature/web-sdk';
 export class AppModule {}
 ```
 
-#### Injecting a feature flag with header value in evaluation context into an endpoint handler method
+With the `OpenFeatureModule` configured it is now possible to inject flag evaluation details into route handlers like in the following code snippet.
 
 ```ts
 import { Controller, ExecutionContext, Get } from '@nestjs/common';
@@ -70,19 +96,6 @@ import { BooleanFeatureFlag } from '@openfeature/nestjs-sdk';
 import { EvaluationDetails } from '@openfeature/server-sdk';
 import { Request } from 'express';
 
-function getContext(executionContext: ExecutionContext) {
-  const request = executionContext.switchToHttp().getRequest<Request>();
-  const userId = request.header('x-user-id');
-
-  if (!userId) {
-    return undefined;
-  }
-
-  return {
-    targetingKey: userId,
-  };
-}
-
 @Controller()
 export class OpenFeatureController {
   @Get('/welcome')
@@ -90,7 +103,6 @@ export class OpenFeatureController {
     @BooleanFeatureFlag({
       flagKey: 'testBooleanFlag',
       defaultValue: false,
-      contextFactory: getContext,
     })
       feature: Observable<EvaluationDetails<boolean>>,
   ) {
@@ -103,7 +115,7 @@ export class OpenFeatureController {
 }
 ```
 
-#### Injecting the default and a named client into a service:
+It is also possible to inject the default and a named client into a service via it's constructor.
 
 ```ts
 import { Injectable } from '@nestjs/common';
@@ -114,15 +126,22 @@ import { FeatureClient } from '@openfeature/nestjs-sdk';
 export class OpenFeatureTestService {
   constructor(
     @FeatureClient() private defaultClient: Client,
-    @FeatureClient({ name: 'differentServer' }) private namedClient: Client,
+    @FeatureClient({ name: 'differentProvider' }) private namedClient: Client,
   ) {
   }
 
-  public async getMessage() {
-    const companyName = await this.defaultClient.getStringValue('companyName', 'Unknown Company');
-    return `Hey User from ${companyName}`;
+  public async getBoolean() {
+    return await this.defaultClient.getBooleanValue('testBooleanFlag', false);
   }
 }
 ```
 
+## Module details
 
+### Configuration
+
+#### Flag evaluation context injection
+
+Whenever a flag evaluation occours, context can be provided with information like user e-mail, role, targeting key, etc in order to trigger specific evaluation rules or logic. The `OpenFeatureModule` provides a way to configure context for each request using the `contextFactory` option.
+The `contextFactory` is ran in a NestJS interceptor scope to configure the evaluation context and than it is used in every flag evaluation related to this request.
+By default the interceptor is cofigured globally, but it can be changed by setting the `useGlobalInterceptor` to `false`. In this case it is still possible to configure a `contextFactory` that can be injected into route, module or controller bound interceptors.
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -36,10 +36,10 @@ The OpenFeature NestJS SDK is a package that provides a NestJS wrapper for the [
 It's main capabilities are:
 
 - Provide a NestJS global module to simplify OpenFeature configuration and usage within NestJS;
-- Supply custom parameter decorators (for controllers), that inject observables with feature flags resolution details;
-- Inject context for flag evaluation seamlessly using NestJS interceptors
-- Provide decorators for OpenFeature Client injection into NestJS services and controllers
-- TODO: map other features
+- Injecting feature flags directly into controller route handlers by using decorators
+- Injecting transaction evaluation context for flag evaluations directly from [execution context](https://docs.nestjs.com/fundamentals/execution-context) (HTTP header values, client IPs, etc.)
+- Injecting OpenFeature clients into NestJS services and controllers by using decorators
+- Setting up logging, event handling, hooks and providers directly when registering the module
 
 ## 🚀 Quick start
 
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -19,6 +19,12 @@
   <a href="https://codecov.io/gh/open-feature/js-sdk">
     <img alt="codecov" src="https://codecov.io/gh/open-feature/js-sdk/branch/main/graph/badge.svg?token=3DC5XOEHMY" />
   </a>
+  <br/>
+    <!-- x-release-please-start-version -->
+  <a href="https://github.com/open-feature/js-sdk/releases/tag/nestjs-sdk-v0.1.0">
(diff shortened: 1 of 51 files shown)


packages/nest/README.md
```diff
@@ -25,11 +25,43 @@
...
+#### npm
+
+```sh
+npm install --save @openfeature/nestjs-sdk
+```
+
+#### yarn
+
+```sh
+yarn add @openfeature/nestjs-sdk
```
beeme1mr, line 59: Yarn unfortunately doesn't automatically install peer dependencies. Please make sure to list any extra dependencies that are required.

packages/nest/README.md
```diff
@@ -25,11 +25,43 @@
...
 for feature flagging that works with your favorite feature flag management tool.
 
+<!-- x-hide-in-docs-end -->
+
 🧪 This SDK is experimental.
 
-#### Here's a basic example of how to use the OpenFeature NestJS API with `InMemoryProvider`.
+## Overview
+
+The OpenFeature NestJS SDK is a package that provides a NestJS wrapper for the [OpenFeature Server SDK](). It's main capabilities are:
```
beeme1mr, line 34:
```suggestion
The OpenFeature NestJS SDK is a package that provides a NestJS wrapper for the [OpenFeature Server SDK]().

It's main capabilities are:
```
Looks like there's a link missing.

packages/nest/README.md
```diff
@@ -25,11 +25,43 @@
...
+- Inject context for flag evaluation seamlessly using NestJS interceptors
+- Provide decorators for OpenFeature Client injection into NestJS services and controllers
+- TODO: map other features
+
+## 🚀 Quick start
+
+### Requirements
+
+- Node.js version 16+
+- NestJS
```
beeme1mr, line 46: Is there a minimum version?

packages/nest/README.md
```diff
@@ -114,15 +127,19 @@ import { FeatureClient } from '@openfeature/nestjs-sdk';
...
+    return await this.defaultClient.getBooleanValue('testBooleanFlag', false);
   }
 }
 ```
 
+## Module aditional information
+
+### Flag evaluation context injection
 
+Whenever a flag evaluation occours, context can be provided with information like user e-mail, role, targeting key, etc in order to trigger specific evaluation rules or logic. The `OpenFeatureModule` provides a way to configure context for each request using the `contextFactory` option.
```
beeme1mr, line 143:
```suggestion
Whenever a flag evaluation occurs, context can be provided with information like user e-mail, role, targeting key, etc in order to trigger specific evaluation rules or logic. The `OpenFeatureModule` provides a way to configure context for each request using the `contextFactory` option.
```

packages/nest/src/open-feature.module.ts
```diff
@@ -13,6 +13,9 @@ import { APP_INTERCEPTOR } from '@nestjs/core';
 import { AsyncLocalStorageTransactionContext } from './evaluation-context-propagator';
 import { EvaluationContextInterceptor } from './evaluation-context-interceptor';
 
+/**
+ * NestJS OpenFeatureModule is a wraper for OpenFeature Server-SDK compatible with NestJS.
```
beeme1mr, line 17:
```suggestion
 * NestJS OpenFeatureModule is a wrapper for OpenFeature Server-SDK compatible with NestJS.
```

packages/nest/README.md
```diff
@@ -70,27 +99,13 @@ import { BooleanFeatureFlag } from '@openfeature/nestjs-sdk';
...
 @Controller()
 export class OpenFeatureController {
   @Get('/welcome')
   public async welcome(
     @BooleanFeatureFlag({
       flagKey: 'testBooleanFlag',
       defaultValue: false,
-      contextFactory: getContext,
     })
       feature: Observable<EvaluationDetails<boolean>>,
```
beeme1mr, line 110: Can this inject an observable instead of a promise? Personally, I don't love working with observables, especially if you're not dealing with a data stream.

FYI @lukas-reining

packages/nest/README.md
lukas-reining: Could be something like this: https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers/config-cat#installation

packages/nest/README.md
```diff
@@ -25,11 +25,43 @@
...
+#### yarn
+
+```sh
+yarn add @openfeature/nestjs-sdk
+```
+
+### Usage
 
-#### Registering the Nest.js SDK module in the App Module:
+The example bellow shows how to use the `OpenFeatureModule` with OpenFeature's `InMemoryProvider`.
```
lukas-reining, line 66:
```suggestion
The example below shows how to use the `OpenFeatureModule` with OpenFeature's `InMemoryProvider`.
```

packages/nest/README.md
lukas-reining: You mean a `Promise` instead of an `Observable` @beeme1mr?
I am fine with using a promise, but my feeling is that Nest.js tends more to use observables for such cases. (e.g. https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers/config-cat#installation)

packages/nest/README.md
```diff
@@ -103,7 +116,7 @@ export class OpenFeatureController {
 }
 ```
 
-#### Injecting the default and a named client into a service:
+It is also possible to inject the default and a named OpenFeature client into a service via it's constructor with Nest dependency injection system.
```
lukas-reining, line 121:
```suggestion
It is also possible to inject the default and named OpenFeature clients into a service via Nest dependency injection system.
```

packages/nest/README.md
```diff
@@ -114,15 +127,19 @@ import { FeatureClient } from '@openfeature/nestjs-sdk';
...
   }
 }
 ```
 
+## Module aditional information
+
+### Flag evaluation context injection
 
+Whenever a flag evaluation occours, context can be provided with information like user e-mail, role, targeting key, etc in order to trigger specific evaluation rules or logic. The `OpenFeatureModule` provides a way to configure context for each request using the `contextFactory` option.
+The `contextFactory` is ran in a NestJS interceptor scope to configure the evaluation context and than it is used in every flag evaluation related to this request.
```
lukas-reining, line 146: We should use `Nest`, `Nest.js` or `NestJS` consistently.
I would go with one of the first two.

packages/nest/src/evaluation-context-interceptor.ts
```diff
@@ -3,6 +3,28 @@ import { ContextFactory, ContextFactoryToken } from './context-factory';
 import { Observable } from 'rxjs';
 import { OpenFeature } from '@openfeature/server-sdk';
```
lukas-reining, line 4: We need to import `OpenFeatureModule` here, to make the linter happy.

packages/nest/src/evaluation-context-interceptor.ts
```diff
@@ -3,6 +3,28 @@ import { ContextFactory, ContextFactoryToken } from './context-factory';
...
 
+/** NestJS interceptor used in {@link OpenFeatureModule}
+ *  to configure flag evaluation context.
+ *
+ *  This interceptor is configured globally by dafault, but if
+ *  `useGlobalInterceptor` is set to `false` in {@link OpenFeatureModule} it needs to be configured for the specific controllers or routes.
```
lukas-reining, line 10:
```suggestion
 *  This interceptor is configured globally by default.
 *  If `useGlobalInterceptor` is set to `false` in {@link OpenFeatureModule} it needs to be configured for the specific controllers or routes.
```

packages/nest/src/open-feature.module.ts
lukas-reining: Maybe double, that the Nest SDK is compatible with Nest.

packages/nest/README.md
beeme1mr: haha, yeah.

beeme1mr: Thinking about this a bit more, I'm not sure if there's a use case for injecting the value of an individual feature flag. It seems much more likely that a user would inject the client into the class.

lukas-reining: @beeme1mr Ya that link was wrong, changed it😅

Regarding the feature:
I could imagine that e.g. changing if the endpoint should be "enabled" or changing the whole request would be interesting.

packages/nest/README.md
toddbaert: This is probably a good question. If you can't identify a specific min version, I would suggest going back at least one version (maybe 2) and see if it works, then call that the min.

packages/nest/README.md
toddbaert: I think "the default and named OpenFeature clients" should read "the default or named OpenFeature clients".

packages/nest/README.md
toddbaert: Based on the docs, I see [NestJS](https://docs.nestjs.com/)

packages/nest/README.md
```diff
@@ -114,15 +129,19 @@ import { FeatureClient } from '@openfeature/nestjs-sdk';
 export class OpenFeatureTestService {
   constructor(
     @FeatureClient() private defaultClient: Client,
-    @FeatureClient({ name: 'differentServer' }) private namedClient: Client,
-  ) {
-  }
+    @FeatureClient({ name: 'differentProvider' }) private namedClient: Client,
```
toddbaert, line 132: I think we should rename this to `OpenFeatureClient` before we remove the `experimental`. I think just `FeatureClient` is a bit disjointed, since we have tended to either add the entire "OpenFeature" prefix, or leave it off. I think `OpenFeatureClient` would be best, because the client inteface is called `Client` so it won't clash, and it's consistent with the equivalent in [react](https://github.com/open-feature/js-sdk/blob/main/packages/react/README.md#multiple-providers-and-scoping) (`</OpenFeatureProvider>`).

packages/nest/README.md
```diff
@@ -25,11 +25,45 @@
...
+## Overview
+
+The OpenFeature NestJS SDK is a package that provides a NestJS wrapper for the [OpenFeature Server SDK]().
+
+It's main capabilities are:
+- Provide a NestJS global module to simplify OpenFeature configuration and usage within NestJS;
+- Supply custom parameter decorators (for controllers), that inject observables with feature flags resolution details;
+- Inject context for flag evaluation seamlessly using NestJS interceptors
+- Provide decorators for OpenFeature Client injection into NestJS services and controllers
+- TODO: map other features
```
toddbaert, line 41: Do we want to remove this TODO?

toddbaert (commented):
I think you should add a release version tag like [this example](https://github.com/open-feature/js-sdk/blob/3ad61cc0553c8f92eb512004001b6843e9208691/packages/client/README.md?plain=1#L18-L21) for auto-inclusion of this SDK in docs.

You can also add a features table like [this](https://github.com/open-feature/js-sdk/blob/3ad61cc0553c8f92eb512004001b6843e9208691/packages/client/README.md?plain=1#L91-L101). The odd situation with this SDK and the React SDK (which I'm working on) is that they get some features "inhereted" from the associated parent SDK - so we may want to indicate that with some new symbol in the table. Any idas? @luizgribeiro  @lukas-reining ?

We also don't have to settle everything in this PR. I will happily come back and add these little enhancements for doc inclusions later.

We could also "hide" the features tables entirely for these "integration-level" SDKs, since they have unique feature-sets related to their framework. cc @beeme1mr

packages/nest/README.md
lukas-reining: I would say there is even more `Nest`, but `NestJS` seems fine.
At least the second was wrong 😅

lukas-reining: I would say there is even more `Nest`, but `NestJS` seems fine.
At least the second was wrong 😅

toddbaert: I mostly care that we're consistent.

packages/nest/README.md
lukas-reining: Actually we have it defined in peerDeps ("^8.0.0 || ^9.0.0 || ^10.0.0").
I think NestJS could be remove from this as it will be in the peerDeps table anyways and it seems doubled with this table.

packages/nest/README.md
lukas-reining: It has been `OpenFeatureClient` before, but [`OpenFeatureClient` is used for the implementation](https://github.com/open-feature/js-sdk/blob/3ad61cc0553c8f92eb512004001b6843e9208691/packages/server/src/client/open-feature-client.ts) of the `Client` and it is also exported from the Server SDK.

We could remove the export of `OpenFeatureClient` from the Web SDK, as no one should use and import this directly anyways I would say. I have seen it several times that this type was used instead of `Client` in projects.
If we remove this I would be happy calling the interface `OpenFeatureClient`here.

packages/nest/README.md
toddbaert: I think for transparency it might be worth just putting 8+ here.

The other thing is maybe we should use no upper bound? I guess that's a risk if there's some related breaking change, but on the other hand, this means we'll have to quickly expand this with every major Nest version

packages/nest/README.md
luizgribeiro: I've already done something pretty similar to this at controller level changing from older to newer/refactored version. Even though I'm not a big fan of observables either, it might be useful.

packages/nest/README.md
luizgribeiro: This is actually a reminder :sweat_smile:

@lukas-reining, is there anything else you think is worth adding to the docs at this point?
I'd say that from JSDocs + Readme it's pretty much all covered.

packages/nest/README.md
luizgribeiro: I agree. It's easier to just fix if any major version breaks it (and maybe pinning the max version) til it's compatible instead of updating on every release.

packages/nest/README.md
luizgribeiro: Alright, 

I'll keep using NestJS since the package is `@openfeature/nestjs-sdk`

luizgribeiro:
> We could also "hide" the features tables entirely for these "integration-level" SDKs, since they have unique feature-sets related to their framework... we could consider some SDKs "language-sdks", and some "framework-sdks" and they could behave differently in the docs. cc

@toddbaert, what do you think about using something like a SDK label to make the difference between language/runtime level SDKs and framework level SDKs? We're facing this in JS right now, but will probably be a valid point in other languages and frameworks as well

packages/nest/README.md
lukas-reining: @luizgribeiro What do you think?

packages/nest/README.md
lukas-reining: @beeme1mr What do you think? I still feel Observables are the most idiomatic for Nest but if others are leaning more to Promises, I would be fine changing it.

packages/nest/README.md
lukas-reining: What do you think @toddbaert?

packages/nest/README.md
beeme1mr: I'm fine with using an Observable here. 👍

packages/nest/README.md
toddbaert: Sorry for the late reply - if we can remove that export, I'm fine with it!

packages/nest/README.md
lukas-reining: I think this is fine, maybe something like "No need to access the global OF api for configuring OF" might be cool but is not required.

packages/nest/README.md
```diff
@@ -25,11 +25,46 @@
...
 
-#### Here's a basic example of how to use the OpenFeature NestJS API with `InMemoryProvider`.
+## Overview
+
+The OpenFeature NestJS SDK is a package that provides a NestJS wrapper for the [OpenFeature Server SDK](https://openfeature.dev/docs/reference/technologies/server/javascript/).
+
+It's main capabilities are:
+
+- Provide a NestJS global module to simplify OpenFeature configuration and usage within NestJS;
+- Supply custom parameter decorators (for controllers), that inject observables with feature flags resolution details;
```
lukas-reining, line 39:
```suggestion
- Injecting feature flags into directly controller route handlers
```

packages/nest/README.md
```diff
@@ -25,11 +25,46 @@
...
+- Inject context for flag evaluation seamlessly using NestJS interceptors
```
lukas-reining, line 40:
```suggestion
- Injecting transaction evaluation context for flag evaluations directly from [execution context](https://docs.nestjs.com/fundamentals/execution-context) (HTTP header values, client IPs, etc.)
```

packages/nest/README.md
```diff
@@ -25,11 +25,46 @@
...
+- Provide decorators for OpenFeature Client injection into NestJS services and controllers
```
lukas-reining, line 41:
```suggestion
- Injecting OpenFeature clients into NestJS services and controllers by using decorators
```

packages/nest/README.md
lukas-reining: I think, this is more about the feature, so the addition that it is using Observables is an implementation detail.

packages/nest/README.md
lukas-reining: I would like to tell that execution context is the source so people familiar with Nest can see directly which information can be expected to be injectable  by the interceptor. I again think that the interceptor is just an implemenation detail.

packages/nest/README.md
lukas-reining: Something like this @luizgribeiro: 

```suggestion
- Setting up logging, event handling, hooks and providers directly when registering the module
```

lukas-reining: I think the semicolons at the end of the points are not needed but not 100% sure :)

packages/nest/README.md
lukas-reining: Hey @toddbaert, removing this should be considered a breaking change I would say.
I am not sure how we want to continue on this then. What do you think?

packages/nest/src/evaluation-context-interceptor.ts
```diff
@@ -2,7 +2,30 @@ import { CallHandler, ExecutionContext, Inject, Injectable, NestInterceptor } fr
...
+ *  ```
+ *
+ *  A different `contextFactory` can also be provided, but the interceptor instance has to be instantiated like in the following example.
+ *  ```ts
+ *  //route interceptor
+ *  @UseInterceptors(new EvaluationContextInterceptor(<context factory>))
+ *  @Get('/user-info')
+ *  getUserInfo(){}
+ *  ```
+ */
```
lukas-reining, line 28:
```suggestion
/** 
 *  NestJS interceptor used in {@link OpenFeatureModule}
 *  to configure flag evaluation context.
 *
 *  This interceptor is configured globally by default.
 *  If `useGlobalInterceptor` is set to `false` in {@link OpenFeatureModule} it needs to be configured for the specific controllers or routes.
 *
 *  If just the interceptor class is passed to the `UseInterceptors` like below, the `contextFactory` provided in the {@link OpenFeatureModule} will be injected and used in order to create the context.
 *  ```ts
 *  //route interceptor
 *  @UseInterceptors(EvaluationContextInterceptor)
 *  @Get('/user-info')
 *  getUserInfo(){}
 *  ```
 *
 *  A different `contextFactory` can also be provided, but the interceptor instance has to be instantiated like in the following example.
 *  ```ts
 *  //route interceptor
 *  @UseInterceptors(new EvaluationContextInterceptor(<context factory>))
 *  @Get('/user-info')
 *  getUserInfo(){}
 *  ```
 */
```

lukas-reining:
> You can also add a features table like [this](https://github.com/open-feature/js-sdk/blob/3ad61cc0553c8f92eb512004001b6843e9208691/packages/client/README.md?plain=1#L91-L101). The odd situation with this SDK and the React SDK (which I'm working on) is that they get some features "inhereted" from the associated parent SDK - so we may want to indicate that with some new symbol in the table. Any idas? @luizgribeiro @lukas-reining ?

@toddbaert yes, maybe we could give this a `Framework integration` semantic. For the feature tables, I think we could maybe leave it out, as the (at least my) assumption is that the base SDK will always provide the feature set that is salwys available and these `Integration SDKs` are just gonna make life better and maybe even add additional specific features. What do you think?

lukas-reining:
> I think you should add a release version tag like [this example](https://github.com/open-feature/js-sdk/blob/3ad61cc0553c8f92eb512004001b6843e9208691/packages/client/README.md?plain=1#L18-L21) for auto-inclusion of this SDK in docs.

This is good, @luizgribeiro could you add this?

packages/nest/README.md
luizgribeiro: It's a nice alternative. I'll rewrite this

luizgribeiro:
I've tried to come up with a nice feature table but it ended up pretty repetitive from what we already have from the server-sdk, so I decided to remove it.

In this framework integration we could use something like a "compatibility table" instead of a feature table.
By doing this it would be possible to express which features from the OF base SDK are already covered and which are not or wip.

beeme1mr:
@toddbaert and I were talking about something similar for the react SDK. We'll just need to come up with a pattern for these framework specific SDKs.

packages/nest/README.md
```diff
@@ -25,11 +29,44 @@
...
+```sh
+npm install --save @openfeature/nestjs-sdk
+```
+
+#### Required peer dependencies
+
+The OpenFeature SDK is required as peer dependency.
 
-#### Here's a basic example of how to use the OpenFeature NestJS API with `InMemoryProvider`.
+The minimum required version of `@openfeature/server-sdk` currently is `1.7.5`.
```
lukas-reining, line 65: We are missing rxjs and Nest here.

beeme1mr: Should `@nestjs/core` also be a peer dependency?

lukas-reining: Yes, this slipped through when we added the interceptors, before we only had `@nestjs/common`.
I will add that @beeme1mr.

packages/nest/README.md
```diff
@@ -25,11 +29,49 @@
...
 for feature flagging that works with your favorite feature flag management tool.
 
-🧪 This SDK is experimental.
+<!-- x-hide-in-docs-end -->
+
+## Overview
+
+The OpenFeature NestJS SDK is a package that provides a NestJS wrapper for the [OpenFeature Server SDK](https://openfeature.dev/docs/reference/technologies/server/javascript/).
+
+It's main capabilities are:
```
beeme1mr, line 38:
```suggestion
Capabilities include:
```

packages/nest/README.md
```diff
@@ -44,24 +86,19 @@ import { InMemoryProvider } from '@openfeature/web-sdk';
...
+        differentProvider: new InMemoryProvider(),
+      },
+    }),
+  ],
 })
 export class AppModule {}
 ```
 
-#### Injecting a feature flag with header value in evaluation context into an endpoint handler method
+With the `OpenFeatureModule` configured it is now possible to inject flag evaluation details into route handlers like in the following code snippet.
```
beeme1mr, line 101:
```suggestion
With the `OpenFeatureModule` configured, it's possible to inject flag evaluation details into route handlers like in the following code snippet.
```

packages/nest/README.md
```diff
@@ -114,15 +137,19 @@ import { FeatureClient } from '@openfeature/nestjs-sdk';
...
 }
 ```
 
+## Module aditional information
+
+### Flag evaluation context injection
 
+Whenever a flag evaluation occurs, context can be provided with information like user e-mail, role, targeting key, etc in order to trigger specific evaluation rules or logic. The `OpenFeatureModule` provides a way to configure context for each request using the `contextFactory` option.
+The `contextFactory` is ran in a NestJS interceptor scope to configure the evaluation context and than it is used in every flag evaluation related to this request.
+By default the interceptor is cofigured globally, but it can be changed by setting the `useGlobalInterceptor` to `false`. In this case it is still possible to configure a `contextFactory` that can be injected into route, module or controller bound interceptors.
```
beeme1mr, line 155:
```suggestion
By default, the interceptor is configured globally, but it can be changed by setting the `useGlobalInterceptor` to `false`. In this case, it is still possible to configure a `contextFactory` that can be injected into route, module or controller bound interceptors.
```

beeme1mr (approved):

beeme1mr:
Awesome job @luizgribeiro and @lukas-reining.

packages/nest/README.md
```diff
@@ -25,11 +29,49 @@
...
+#### Required peer dependencies
 
-#### Here's a basic example of how to use the OpenFeature NestJS API with `InMemoryProvider`.
+The following list contains the peer dependencies of `@openfeature/nestjs-sdk` with it's expected and compatible versions:
 
-#### Registering the Nest.js SDK module in the App Module:
+* `@openfeature/server-sdk`: >=1.7.5
+* `@nestjs/common`: ^8.0.0 || ^9.0.0 || ^10.0.0
+* `@nestjs/core`: ^8.0.0 || ^9.0.0 || ^10.0.0
+* `rxjs`: ^6.0.0 || ^7.0.0
```
lukas-reining, line 68:
```suggestion
* `@openfeature/server-sdk`: >=1.7.5
* `@nestjs/common`: ^8.0.0 || ^9.0.0 || ^10.0.0
* `@nestjs/core`: ^8.0.0 || ^9.0.0 || ^10.0.0
* `rxjs`: ^6.0.0 || ^7.0.0 || ^8.0.0
```

lukas-reining (approved):
This looks really good now @luizgribeiro!
I will merge it and we can refine more details later if needed.

packages/nest/README.md
toddbaert: hmm, well, if I understand correctly, it would be technically breaking, but I think it could be argued it was a misuse. All our doc recommends using the interface.

If we want to be really strict, we could consider it a break and include it with some others I have in mind in a 2.0.0 (meaning to ask you guys about this soon... there's a small list of breaking improvements I think might be worth considering)...

PR #751, opened 2024-01-10
chore: update renovate to use best practices

beeme1mr:
## This PR

- enables renovate best practices

### Notes

The current configuration seems to be invalid. This PR updates Renovate to use best practices recommended by the Renovate team.

https://docs.renovatebot.com/presets-config/#configbest-practices

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,16 +1,12 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:best-practices"
+    "config:recommended",
+    ":automergeTypes",
+    ":automergeStableNonMajor",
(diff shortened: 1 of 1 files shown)


toddbaert:
Does this pin shas of actions? That's my only concern, I hate that mode.

toddbaert:
> > Does this pin shas of actions? That's my only concern, I hate that mode.
> 
> Yes, it does. Perhaps we should use [config-js-lib](https://docs.renovatebot.com/presets-config/#configjs-lib) or https://docs.renovatebot.com/presets-config/#configjs-app instead.

Ya, or maybe we can override it. I'm sure the rest is fine, but I can't stand the breaking changes the action sha pinning introduces.

PR #752, opened 2024-01-10
fix: wrong context passed to named providers

toddbaert:
Fixed an issue with implementations of https://github.com/open-feature/js-sdk/pull/704 where the wrong context was passed to namespaced providers.

### DIFF
--- a/packages/client/src/client/open-feature-client.ts
+++ b/packages/client/src/client/open-feature-client.ts
@@ -179,7 +179,7 @@ export class OpenFeatureClient implements Client {
     const allHooksReversed = [...allHooks].reverse();
 
     const context = {
-      ...OpenFeature.getContext(),
+      ...OpenFeature.getContext(this?.options?.name),
     };
 
     // this reference cannot change during the course of evaluation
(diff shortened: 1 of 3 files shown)


packages/client/src/client/open-feature-client.ts
```diff
@@ -179,7 +179,7 @@ export class OpenFeatureClient implements Client {
     const allHooksReversed = [...allHooks].reverse();
 
     const context = {
-      ...OpenFeature.getContext(),
+      ...OpenFeature.getContext(this?.options?.name),
```
toddbaert, line 182: We simply should have passed the context name here, if set, and forgot.

beeme1mr: We should probably add a test to make sure the correct context is being sent to hooks and to the resolve call. I missed that cause when adding client side context management feature.

lukas-reining (approved):
LGTM!

beeme1mr (approved):
Approving but I do think you should add an additional test. We're not sufficiently covering the evaluation process when using namespaced evaluation context.

toddbaert:
> Approving but I do think you should add an additional test. We're not sufficiently covering the evaluation process when using namespaced evaluation context.

Agreed I think it should be simple to add another.

beeme1mr (approved):

packages/client/src/client/open-feature-client.ts
toddbaert: Nice catch and done.

PR #753, opened 2024-01-11
fix: some handlers fail to run

toddbaert:
fixed an issue where some handlers would fail to run if the associated provider did not have a onContextChange method

### DIFF
--- a/packages/client/src/open-feature.ts
+++ b/packages/client/src/open-feature.ts
@@ -18,7 +18,7 @@ type OpenFeatureGlobal = {
   [GLOBAL_OPENFEATURE_API_KEY]?: OpenFeatureAPI;
 };
 type NameProviderTuple = {
-  namespace?: string;
+  name?: string;
   provider: Provider;
 }
 
@@ -89,18 +89,18 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider, Hook> impleme
       // collect all providers that are using the default context (not mapped to a name)
       const defaultContextNameProviders: NameProviderTuple[] = Array.from(this._clientProviders.entries())
         .filter(([name]) => !this._namedProviderContext.has(name))
-        .reduce<{ namespace: string; provider: Provider }[]>((acc, [name, provider]) => {
-          acc.push({ namespace: name, provider });
+        .reduce<NameProviderTuple[]>((acc, [name, provider]) => {
+          acc.push({ name: name, provider });
           return acc;
         }, []);
 
       const allProviders: NameProviderTuple[] = [
-        { namespace: undefined, provider: this._defaultProvider },
+        { name: undefined, provider: this._defaultProvider },
         ...defaultContextNameProviders,
       ];
       await Promise.all(
         allProviders.map((tuple) =>
-          this.runProviderContextChangeHandler(tuple.namespace, tuple.provider, oldContext, context),
+          this.runProviderContextChangeHandler(tuple.name, tuple.provider, oldContext, context),
         ),
       );
     }
--- a/packages/client/src/open-feature.ts
+++ b/packages/client/src/open-feature.ts
@@ -17,6 +17,11 @@ const GLOBAL_OPENFEATURE_API_KEY = Symbol.for('@openfeature/web-sdk/api');
 type OpenFeatureGlobal = {
   [GLOBAL_OPENFEATURE_API_KEY]?: OpenFeatureAPI;
 };
+type NameProviderTuple = {
+  namespace?: string;
+  provider: Provider;
+}
+
 const _globalThis = globalThis as OpenFeatureGlobal;
 
 export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider, Hook> implements ManageContext<Promise<void>> {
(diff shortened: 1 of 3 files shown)


packages/client/src/open-feature.ts
```diff
@@ -196,19 +207,25 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider, Hook> impleme
...
-        this._events?.emit(ProviderEvents.ContextChanged, { clientName, providerName });
-      }).catch((err) => {
-        this._logger?.error(`Error running ${provider.metadata.name}'s context change handler:`, err);
-        this.getAssociatedEventEmitters(clientName).forEach((emitter) => {
-          emitter?.emit(ProviderEvents.Error, { clientName, providerName, message: err?.message, });
-        });
-        this._events?.emit(ProviderEvents.Error, { clientName, providerName, message: err?.message, });
+    const providerName = provider.metadata.name;
+    try {
+      await provider.onContextChange?.(oldContext, newContext);
```
toddbaert, line 212: This was the bug - the previous implementation only emitted events if the provider had a `onContextChange` implementation, which is optional, because it was chained like this:

`provider.onContextChange?.(oldContext, newContext).then(...`

Now we await the `onContextChange` OR just run the handlers.

packages/client/test/events.spec.ts
```diff
@@ -43,6 +45,9 @@ class MockProvider implements Provider {
     this.enableEvents = options?.enableEvents ?? true;
     this.failOnInit = options?.failOnInit ?? false;
     this.failOnContextChange = options?.failOnContextChange ?? false;
+    if (!options?.noContextChanged) {
+      this.onContextChange = this.changeHandler;
+    }
```
toddbaert, line 50: supporting test option.

packages/client/test/events.spec.ts
```diff
@@ -730,5 +744,29 @@ describe('Events', () => {
 
       });
     });
+
+    describe('provider', () => {
+      describe('has no onContextChange handler', () => {
```
toddbaert, line 749: new test for the main bug.

packages/shared/src/open-feature.ts
```diff
@@ -108,6 +108,13 @@ export abstract class OpenFeatureCommonAPI<P extends CommonProvider = CommonProv
     this._events.removeHandler(eventType, handler);
   }
 
+  /**
+   * Removes all event handlers.
+   */
+  clearHandlers(): void {
+    this._events.removeAllHandlers();
+  }
+
```
toddbaert, line 117: This helps make testing simpler.

packages/client/src/open-feature.ts
```diff
@@ -17,6 +17,11 @@ const GLOBAL_OPENFEATURE_API_KEY = Symbol.for('@openfeature/web-sdk/api');
 type OpenFeatureGlobal = {
   [GLOBAL_OPENFEATURE_API_KEY]?: OpenFeatureAPI;
 };
+type NameProviderTuple = {
```
beeme1mr, line 20: We should come up with a better name for this.

packages/client/src/open-feature.ts
```diff
@@ -17,6 +17,11 @@ const GLOBAL_OPENFEATURE_API_KEY = Symbol.for('@openfeature/web-sdk/api');
...
+  namespace?: string;
```
beeme1mr, line 21: The namespace PR hasn't been approved yet.

packages/client/src/open-feature.ts
```diff
@@ -81,16 +86,22 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider, Hook> impleme
       const oldContext = this._context;
       this._context = context;
 
-      const providersWithoutContextOverride = Array.from(this._clientProviders.entries())
+      // collect all providers that are using the default context (not mapped to a name)
+      const defaultContextNameProviders: NameProviderTuple[] = Array.from(this._clientProviders.entries())
         .filter(([name]) => !this._namedProviderContext.has(name))
-        .reduce<Provider[]>((acc, [, provider]) => {
-          acc.push(provider);
+        .reduce<{ namespace: string; provider: Provider }[]>((acc, [name, provider]) => {
```
beeme1mr, line 92: Reuse the type above once we have a better name.

packages/client/src/open-feature.ts
```diff
@@ -81,16 +86,22 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<Provider, Hook> impleme
...
         }, []);
 
-      const allProviders = [this._defaultProvider, ...providersWithoutContextOverride];
+      const allProviders: NameProviderTuple[] = [
+        { namespace: undefined, provider: this._defaultProvider },
+        ...defaultContextNameProviders,
+      ];
       await Promise.all(
-        allProviders.map((provider) => this.runProviderContextChangeHandler(undefined, provider, oldContext, context)),
+        allProviders.map((tuple) =>
```
beeme1mr, line 102: Why did you call this a tuple?

toddbaert: Because it is?

packages/client/src/open-feature.ts
toddbaert: I think it will be and I just thought it would be an easy thing to miss and a rename, and not a problem to use it early here, but I can change it to `name`.

packages/client/src/open-feature.ts
toddbaert: I have changed "namespace" to "name" and in that case I think this is a good name for this type.

packages/client/src/open-feature.ts
toddbaert: Ya, that was why I created the name above.

packages/client/src/open-feature.ts
toddbaert: Renamed to record for clarity.

packages/client/src/open-feature.ts
toddbaert: renamed to -Record

beeme1mr (approved):

lukas-reining (approved):

PR #761, opened 2024-01-14
feat: add logger, event handlers and hooks to nest sdk

lukas-reining:
## This PR

Adds options for logger, event handlers and hooks to Nest SDK initialization.
This change makes the SDK completely configurable during the initialization without the need to explicitly call the OpenFeature API.

### Related Issues

Relates to #705

### DIFF
--- a/packages/nest/src/open-feature.module.ts
+++ b/packages/nest/src/open-feature.module.ts
@@ -7,7 +7,16 @@ import {
   Provider as NestProvider,
   ExecutionContext,
 } from '@nestjs/common';
-import { Client, OpenFeature, Provider, EvaluationContext } from '@openfeature/server-sdk';
+import {
+  Client,
+  Hook,
+  OpenFeature,
+  Provider,
+  EvaluationContext,
+  ServerProviderEvents,
+  EventHandler,
+  Logger,
+} from '@openfeature/server-sdk';
 import { ContextFactory, ContextFactoryToken } from './context-factory';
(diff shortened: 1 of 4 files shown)


packages/nest/src/open-feature.module.ts
```diff
@@ -17,7 +26,18 @@ import { EvaluationContextInterceptor } from './evaluation-context-interceptor';
 export class OpenFeatureModule {
   static forRoot({ useGlobalInterceptor = true, ...options }: OpenFeatureModuleOptions): DynamicModule {
     OpenFeature.setTransactionContextPropagator(new AsyncLocalStorageTransactionContext());
-    const providers: NestProvider[] = [];
+
+    if (options.logger) {
```
luizgribeiro, line 30: I'm wondering if there's a way to pass NetsJS' global logger and inject it instead of passing it explicitly as an option.

Do you think it is possible using the current structure?

luizgribeiro (approved):
LGTM!

beeme1mr (approved):

luizgribeiro:
@lukas-reining, I've opened #767 to this branch but I'm not 100% sure about how to do this.
I ended up with a hacky solution, but it might be enough to dispose OF Providers resources on App shutdown as expected.

packages/nest/src/open-feature.module.ts
lukas-reining: Had a look, it seems not to be possible to get it.
Maybe we could just add a facade for Nest.js loggers.

PR #766, opened 2024-01-16
feat!: remove generic hook, add specific type hooks

toddbaert:
This PR brings the react SDK's evaluation API in-line with other SDKS. It does this by:

- adding flag value hooks for each type (these all use common code, with only differing generic args)
- adding flag details hooks for each type (again using common code)
- adding optional generic constraints for each

I think this is important before a non-experimental release for 2 reasons:
- it's consistent with our other JS components and other SDKs
- it fixes a potential bug if uses accidentally pass the wrong default type

:warning: This is a breaking change.

### DIFF
--- a/packages/react/src/use-feature-flag.ts
+++ b/packages/react/src/use-feature-flag.ts
@@ -60,7 +60,7 @@ export function useBooleanFlagValue(flagKey: string, defaultValue: boolean, opti
 export function useBooleanFlagDetails(flagKey: string, defaultValue: boolean, options?: ReactFlagEvaluationOptions): EvaluationDetails<boolean> {
   return attachHandlersAndResolve(flagKey, defaultValue, (client) => {
     return client.getBooleanDetails;
-  },  options)
+  },  options);
 }
 
 /**
@@ -88,7 +88,7 @@ export function useStringFlagValue<T extends string = string>(flagKey: string, d
 export function useStringFlagDetails<T extends string = string>(flagKey: string, defaultValue: T, options?: ReactFlagEvaluationOptions): EvaluationDetails<T> {
   return attachHandlersAndResolve(flagKey, defaultValue, (client) => {
     return client.getStringDetails<T>;
-  },  options)
+  },  options);
 }
 
 /**
@@ -116,7 +116,7 @@ export function useNumberFlagValue<T extends number = number>(flagKey: string, d
 export function useNumberFlagDetails<T extends number = number>(flagKey: string, defaultValue: T, options?: ReactFlagEvaluationOptions): EvaluationDetails<T> {
   return attachHandlersAndResolve(flagKey, defaultValue, (client) => {
     return client.getNumberDetails<T>;
-  },  options)
+  },  options);
 }
 
 /**
@@ -144,7 +144,7 @@ export function useObjectFlagValue<T extends JsonValue = JsonValue>(flagKey: str
 export function useObjectFlagDetails<T extends JsonValue = JsonValue>(flagKey: string, defaultValue: T, options?: ReactFlagEvaluationOptions): EvaluationDetails<T> {
   return attachHandlersAndResolve(flagKey, defaultValue, (client) => {
     return client.getObjectDetails<T>;
-  },  options)
+  },  options);
 }
 
 function attachHandlersAndResolve<T extends FlagValue>(flagKey: string, defaultValue: T, resolver: (client: Client) => (flagKey: string, defaultValue: T) => EvaluationDetails<T>, options?: ReactFlagEvaluationOptions): EvaluationDetails<T> {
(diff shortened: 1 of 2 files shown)


packages/react/src/use-feature-flag.ts
```diff
@@ -80,19 +182,7 @@ export function useFeatureFlag<T extends FlagValue>(flagKey: string, defaultValu
...
-
-function getFlag<T extends FlagValue>(client: Client, flagKey: string, defaultValue: T): EvaluationDetails<T> {
-  if (typeof defaultValue === 'boolean') {
-    return client.getBooleanDetails(flagKey, defaultValue) as EvaluationDetails<T>;
-  } else if (typeof defaultValue === 'string') {
-    return client.getStringDetails(flagKey, defaultValue) as EvaluationDetails<T>;
-  } else if (typeof defaultValue === 'number') {
-    return client.getNumberDetails(flagKey, defaultValue) as EvaluationDetails<T>;
-  } else {
-    return client.getObjectDetails(flagKey, defaultValue) as EvaluationDetails<T>;
```
toddbaert, line 94: This function used `defaultValue` to call the correct resolver, which could result in some surprising behavior if the user passes a var of a different type than they are expecting. We decided on strongly typed methods to avoid this.

packages/react/src/use-feature-flag.ts
```diff
@@ -37,15 +37,117 @@ enum SuspendState {
...
+/**
+ * Evaluates a feature flag, returning a boolean.
+ * By default, components will re-render when the flag value changes.
+ * @param {string} flagKey the flag identifier
+ * @param {boolean} defaultValue the default value
+ * @param {ReactFlagEvaluationOptions} options options for this evaluation
+ * @returns { boolean} a EvaluationDetails object for this evaluation
+ */
+export function useBooleanFlagValue(flagKey: string, defaultValue: boolean, options?: ReactFlagEvaluationOptions): boolean {
+  return useBooleanFlagDetails(flagKey, defaultValue, options)?.value;
```
beeme1mr, line 49:
```suggestion
  return useBooleanFlagDetails(flagKey, defaultValue, options).value;
```

packages/react/src/use-feature-flag.ts
```diff
@@ -37,15 +37,117 @@ enum SuspendState {
...
+ * By default, components will re-render when the flag value changes.
+ * @param {string} flagKey the flag identifier
+ * @template {string} [T=string] A optional generic argument constraining the string
  * @param {T} defaultValue the default value
  * @param {ReactFlagEvaluationOptions} options options for this evaluation
- * @template T flag type
+ * @returns { boolean} a EvaluationDetails object for this evaluation
+ */
+export function useStringFlagValue<T extends string = string>(flagKey: string, defaultValue: T, options?: ReactFlagEvaluationOptions): T {
+  return useStringFlagDetails(flagKey, defaultValue, options)?.value;
```
beeme1mr, line 76:
```suggestion
  return useStringFlagDetails(flagKey, defaultValue, options).value;
```

packages/react/src/use-feature-flag.ts
```diff
@@ -37,15 +37,117 @@ enum SuspendState {
...
+ * Evaluates a feature flag, returning a number.
+ * By default, components will re-render when the flag value changes.
+ * @param {string} flagKey the flag identifier
+ * @template {number} [T=number] A optional generic argument constraining the number
+ * @param {T} defaultValue the default value
+ * @param {ReactFlagEvaluationOptions} options options for this evaluation
+ * @returns { boolean} a EvaluationDetails object for this evaluation
+ */
+export function useNumberFlagValue<T extends number = number>(flagKey: string, defaultValue: T, options?: ReactFlagEvaluationOptions): T {
+  return useNumberFlagDetails(flagKey, defaultValue, options)?.value;
```
beeme1mr, line 104:
```suggestion
  return useNumberFlagDetails(flagKey, defaultValue, options).value;
```

packages/react/src/use-feature-flag.ts
```diff
@@ -37,15 +37,117 @@ enum SuspendState {
...
+ * Evaluates a feature flag, returning an object.
+ * By default, components will re-render when the flag value changes.
+ * @param {string} flagKey the flag identifier
+ * @template {JsonValue} [T=JsonValue] A optional generic argument describing the structure
+ * @param {T} defaultValue the default value
+ * @param {ReactFlagEvaluationOptions} options options for this evaluation
+ * @returns { boolean} a EvaluationDetails object for this evaluation
+ */
+export function useObjectFlagValue<T extends JsonValue = JsonValue>(flagKey: string, defaultValue: T, options?: ReactFlagEvaluationOptions): T {
+  return useObjectFlagDetails<T>(flagKey, defaultValue, options)?.value;
```
beeme1mr, line 132:
```suggestion
  return useObjectFlagDetails<T>(flagKey, defaultValue, options).value;
```

beeme1mr (commented):
`value` should always be present on the `EvaluationDetails`.

luizgribeiro (approved):
The newer API made a lot o sense IMO.

LGTM!

luizgribeiro:
So, from what we've been exploring in the Nest SDK I think there might be a nicer way to deal with multiple providers at the same time. It's totally out of scope for this PR, but I wanted to write it down since it's the first time that I look at the React SDK.
I don't know if it is reasonable now, but could OF use one Context to store all providers resolutions and flag evaluation hooks provide the desired provider name  through `options`?

lukas-reining (approved):
I like the change and that it makes all SDKs and components be more consistent!

PR #767, opened 2024-01-18
feat: add shutdown service

luizgribeiro:
## This PR

- adds a way to gracefully shutdown providers in the Nest SDK

### DIFF
--- a/packages/nest/src/shutdown.service.ts
+++ b/packages/nest/src/shutdown.service.ts
@@ -1,0 +1,18 @@
+import { Injectable, OnApplicationShutdown } from '@nestjs/common';
+import { Provider } from '@openfeature/server-sdk';
+
+@Injectable()
+export class ShutdownService implements OnApplicationShutdown {
+  constructor(private readonly openFeatureProviders: Provider[]) {}
+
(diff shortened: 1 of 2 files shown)


beeme1mr (approved):

packages/nest/src/shutdown.service.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  async onApplicationShutdown() {
+    const closePromises: Array<Promise<void>> = [];
+    for (const ofProvider of this.openFeatureProviders) {
+      if (ofProvider.onClose) {
+        closePromises.push(ofProvider.onClose());
+      }
+    }
+
+    await Promise.all(closePromises);
```
lukas-reining, line 16:
```suggestion
    await OpenFeature.close()
```

I think we could just go for the close method of OpenFeature.
As the Application is shutdown anyways, we might just want to let OF manage the whole shutdown here.
This would also mean that we basically only have to create this service and not change anything about injecting the providers :)

lukas-reining (commented):
Good change! I left one comment that I would do before merging it.

PR #788, opened 2024-01-25
fix: event-handler leakage

toddbaert:
Fixes an issue where event handlers could be leaked if the same handler reference was added multiple times.

### DIFF
--- a/packages/shared/src/events/generic-event-emitter.ts
+++ b/packages/shared/src/events/generic-event-emitter.ts
@@ -1,6 +1,6 @@
 import { Logger, ManageLogger, SafeLogger } from '../logger';
 import { EventContext, EventDetails, EventHandler } from './eventing';
-import { AnyProviderEvent } from './events';
+import { AllProviderEvents, AnyProviderEvent } from './events';
 
 /**
  * The GenericEventEmitter should only be used within the SDK. It supports additional properties that can be included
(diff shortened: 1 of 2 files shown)


packages/shared/src/events/generic-event-emitter.ts
```diff
@@ -29,15 +34,16 @@ export abstract class GenericEventEmitter<E extends AnyProviderEvent, Additional
       await handler(details);    
     };
     // The async handler has to be written to the map, because we need to get the wrapper function when deleting a listener
-    this._handlers.set(handler, asyncHandler);
```
toddbaert, line 32: If the same handler (let's say 0x00000001) is added twice, the first time it's added to the weakmap by it's reference with a new wrapped async ref as a value (`0x00000001`:`0x00000002`). Then, it's added again (with a new value for it's new async wrapper since that's a local var (`0x00000001`:`0x00000003`). Since it's the same handler being added again, it has the same key... so we've overwritten `0x00000001`:`0x00000002` with `0x00000001`:`0x00000003` and `0x00000002` is lost forever and cannot be removed :sweat_smile:

packages/shared/test/events.spec.ts
```diff
@@ -130,11 +150,32 @@ describe('GenericEventEmitter', () => {
...
+      emitter.removeHandler(AllProviderEvents.ContextChanged, handler);
+
+      emitter.emit(AllProviderEvents.Stale);
+      emitter.emit(AllProviderEvents.ContextChanged);
+
+      await wait();
+
+      expect(handler).toHaveBeenCalledTimes(0);
+      expect(handler).toHaveBeenCalledTimes(0);
+    });
```
toddbaert, line 176: This test fails without the fix.

beeme1mr (approved):

packages/shared/test/events.spec.ts
```diff
@@ -130,11 +148,31 @@ describe('GenericEventEmitter', () => {
...
+      emitter.removeHandler(AllProviderEvents.Stale, handler);
+      emitter.removeHandler(AllProviderEvents.ContextChanged, handler);
+
+      emitter.emit(AllProviderEvents.Stale);
+      emitter.emit(AllProviderEvents.ContextChanged);
+
+      await wait();
+
+      expect(handler).toHaveBeenCalledTimes(0);
+    });
```
toddbaert, line 173: This test fails without the fix.

packages/shared/src/events/generic-event-emitter.ts
lukas-reining: Good catch🤦‍♂️

lukas-reining (approved):

toddbaert:
It occurred to me this fix is incomplete. It still possible to add the same handlers to the same event twice and lose the ref. I have another fix in mind. Marked as draft for now

toddbaert:
Hey @beeme1mr @lukas-reining please re-review. As mentioned [here](https://github.com/open-feature/js-sdk/pull/788#issuecomment-1911013687), we also need to support the addition and removal of duplicate handlers to the same event. This is done by making the weakmap values arrays, and appending/popping them.

I've added this functionality, and 2 relevant tests.

beeme1mr (approved):

lukas-reining (approved):

PR #794, opened 2024-01-27
fix: remove export of OpenFeatureClient

lukas-reining:
## This PR
As discussed here https://github.com/open-feature/js-sdk/pull/750#discussion_r1450896230, we should not export `OpenFeatureClient` from the server and web SDK. The type used outside the SDK should be `Client`, which is also used in the public APIs like `OpenFeatureApi`.

The question is, if we should mark this as breaking. 
Technically it will break code that imports `OpenFeatureClient` instead of `Client`, but as @toddbaert said code using it  could also be seen as "used wrong" while being technically fine.

I am still leaning towards marking it as breaking, to be sure we are not breaking something that is technically fine, just because it is unintended use. But I could also live with non-beaking.

### DIFF
--- a/packages/client/src/client/internal/open-feature-client.ts
+++ b/packages/client/src/client/internal/open-feature-client.ts
@@ -35,6 +35,11 @@ type OpenFeatureClientOptions = {
   version?: string;
 };
 
+/**
+ * This implementation of the {@link Client} is meant to only be instantiated by the SDK.
+ * It should not be used outside the SDK and so should not be exported.
+ * @internal
+ */
 export class OpenFeatureClient implements Client {
   private _hooks: Hook[] = [];
   private _clientLogger?: Logger;
--- a/packages/client/src/client/internal/open-feature-client.ts
+++ b/packages/client/src/client/internal/open-feature-client.ts
@@ -18,13 +18,13 @@ import {
   instantiateErrorByErrorCode,
   statusMatchesEvent,
 } from '@openfeature/core';
-import { FlagEvaluationOptions } from '../evaluation';
-import { ProviderEvents } from '../events';
-import { InternalEventEmitter } from '../events/internal/internal-event-emitter';
-import { Hook } from '../hooks';
-import { OpenFeature } from '../open-feature';
-import { Provider, ProviderStatus } from '../provider';
-import { Client } from './client';
+import { FlagEvaluationOptions } from '../../evaluation';
+import { ProviderEvents } from '../../events';
+import { InternalEventEmitter } from '../../events/internal/internal-event-emitter';
+import { Hook } from '../../hooks';
+import { OpenFeature } from '../../open-feature';
+import { Provider, ProviderStatus } from '../../provider';
+import { Client } from './../client';
 
 type OpenFeatureClientOptions = {
   /**
--- a/packages/client/src/open-feature.ts
+++ b/packages/client/src/open-feature.ts
@@ -9,10 +9,10 @@ import {
   stringOrUndefined,
 } from '@openfeature/core';
 import { Client } from './client';
+import { OpenFeatureClient } from './client/internal/open-feature-client';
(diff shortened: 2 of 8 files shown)


packages/client/src/client/internal/open-feature-client.ts
beeme1mr, line 1: Could you add a JSDoc to the top of the class or to the constructor that explains that this is meant to only be used internally?

beeme1mr (commented):
Perhaps we should enable the [stripInternal](https://www.typescriptlang.org/tsconfig#stripInternal) compiler option.  I haven't used it before but this may be a nice way to ensure we don't accidentally reexport it in the future.

toddbaert:
I think we can (debatabley) consider this non-breaking, since it's an export we didn't expect people to use... it really depends how strict we want to be.

In any case, I think it might be a good idea to discuss rolling together some upcoming breaking changes. I'd like to also include:

- making it impossible to return an error resolution from providers (force them to throw if something goes wrong)
- removing `setProviderAndWait` and just making `setProvider` async

What do you guys think?

cc @lukas-reining @beeme1mr @luizgribeiro @Kavindu-Dodan

Kavindu-Dodan:
> I think we can (debatabley) consider this non-breaking, since it's an export we didn't expect people to use... it really depends how strict we want to be.
> 
> In any case, I think it might be a good idea to discuss rolling together some upcoming breaking changes. I'd like to also include:
> 
> * making it impossible to return an error resolution from providers (force them to throw if something goes wrong)
> * removing `setProviderAndWait` and just making `setProvider` async
> 
> What do you guys think?
> 
> cc @lukas-reining @beeme1mr @luizgribeiro @Kavindu-Dodan

+1 for combining multiple breaking changes for the next release

luizgribeiro (commented):
I'd say it's a breaking change because if it's shipped as is since it would break minor upgrades in projects that are already using `OpenFeatureClient`.
I also think bundling more breaking changes in one new major is a good alternative.

Maybe another way that might not be a breaking change is to expose `OpenFeatureClient` as an alias type to `Client`. 
I'm not a big fan of using `Client` as a type. Many different packages have a `Client` exported type and it would end up with an alias anyway.

toddbaert:
> I'd say it's a breaking change because if it's shipped as is since it would break minor upgrades in projects that are already using `OpenFeatureClient`. I also think bundling more breaking changes in one new major is a good alternative.
> 
> Maybe another way that might not be a breaking change is to expose `OpenFeatureClient` as an alias type to `Client`. I'm not a big fan of using `Client` as a type. Many different packages have a `Client` exported type and it would end up with an alias anyway.

I think it's OK to require people to alias things, but I would be up for changing the name to something more specific in a 2.0. I definitely don't want to alias the `OpenFeatureClient` though since it's a class. It's a good idea to export and reference interfaces instead of classes where-ever possible, or you can run into nasty things [like this](https://cloud-native.slack.com/archives/C06E4DE6S07/p1706650042057369?thread_ts=1706562067.358389&cid=C06E4DE6S07) which can happen if people end up with multiple copies of a dep (we do our best to avoid that but it can happen).

toddbaert:
I've opened [this](https://github.com/open-feature/js-sdk/issues/799) to record (small) proposed breaking changes in 2.0

packages/client/src/client/internal/open-feature-client.ts
lukas-reining: did so, should we already add the `stripInternal` flag @beeme1mr @toddbaert ?

toddbaert:
@lukas-reining @luizgribeiro @beeme1mr I've rebased and resolved conflcits on this PR.

I also made the title non-breaking as we decided (though it's technically a break but not on an export we indented people to use).

toddbaert (approved):

beeme1mr (approved):

PR #796, opened 2024-01-30
feat: suspend on RECONCILING, mem provider fixes

toddbaert:
This PR adds "suspend on reconciling" to the React SDK, in a manner [consistent with this spec](https://openfeature.dev/specification/sections/events#event-handlers-and-context-reconciliation).

When the SDK emits a PROVIDER_RECONCILING event, loaders are displayed until the CONTEXT_CHANGED event is emitted.

### DIFF
--- a/packages/client/src/provider/in-memory-provider/in-memory-provider.ts
+++ b/packages/client/src/provider/in-memory-provider/in-memory-provider.ts
@@ -52,6 +52,7 @@ export class InMemoryProvider implements Provider {
       .map(([key]) => key);
 
     this._flagConfiguration = { ...flagConfiguration };
+
     try {
       await this.initialize(this._context);
       this.events.emit(ProviderEvents.ConfigurationChanged, { flagsChanged });
--- a/packages/react/src/use-feature-flag.ts
+++ b/packages/react/src/use-feature-flag.ts
@@ -5,10 +5,16 @@ import { useOpenFeatureClient } from './provider';
 type ReactFlagEvaluationOptions = {
   /**
    * Suspend flag evaluations while the provider is not ready.
-   * Set to false if you don't want to use React Suspense API.
+   * Set to false if you don't want to show suspense fallbacks util the provider is initialized.
    * Defaults to true.
    */
-  suspend?: boolean,
+  suspendUntilReady?: boolean,
+  /**
+   * Suspend flag evaluations while the provider's context is being reconciled.
+   * Set to true if you want to show suspense fallbacks while flags are re-evaluated after context changes.
+   * Defaults to false.
+   */
+  suspendWhileStale?: boolean,
   /**
    * Update the component if the provider emits a ConfigurationChanged event.
    * Set to false to prevent components from re-rendering when flag value changes
@@ -28,7 +34,8 @@ type ReactFlagEvaluationOptions = {
 const DEFAULT_OPTIONS: ReactFlagEvaluationOptions = {
   updateOnContextChanged: true,
   updateOnConfigurationChanged: true,
-  suspend: true,
+  suspendUntilReady: true,
+  suspendWhileStale: false,
 };
 
 enum SuspendState {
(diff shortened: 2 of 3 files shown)


packages/client/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -59,16 +59,11 @@ export class InMemoryProvider implements Provider {
...
-    this.events.emit(ProviderEvents.Stale);
-
     this._flagConfiguration = { ...flagConfiguration };
-    this.events.emit(ProviderEvents.ConfigurationChanged, { flagsChanged });
 
     try {
       await this.initialize(this._context);
-      // we need to emit our own events in this case, since it's not part of the init flow.
-      this.events.emit(ProviderEvents.Ready);
+      this.events.emit(ProviderEvents.ConfigurationChanged, { flagsChanged });
```
toddbaert, line 66: This should only be emitted after the change is fully processed, so I moved it later.

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -59,16 +59,11 @@ export class InMemoryProvider implements Provider {
       .filter(([key, value]) => this._flagConfiguration[key] !== value)
       .map(([key]) => key);
 
-    this.status = ProviderStatus.STALE;
```
toddbaert, line 62: No need for stale here - this is a configuration change (a change sourced from the provider itself) not a change from setting context, etc.

This corresponds to a change coming directly from a management system (no need for STALE/READY here).

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -59,16 +59,11 @@ export class InMemoryProvider implements Provider {
...
-    this.events.emit(ProviderEvents.Stale);
-
     this._flagConfiguration = { ...flagConfiguration };
-    this.events.emit(ProviderEvents.ConfigurationChanged, { flagsChanged });
 
     try {
       await this.initialize(this._context);
-      // we need to emit our own events in this case, since it's not part of the init flow.
-      this.events.emit(ProviderEvents.Ready);
```
beeme1mr, line 71: Why is this no longer relevant?

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
beeme1mr: Please double-check the server implementation to make sure it doesn't have the same issue.

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
toddbaert: See my [other comment](https://github.com/open-feature/js-sdk/pull/796#discussion_r1471597115).

There's no need for STALE/READY or anything here. This is a method completely unique to the in-memory provider that changes it's config. It should simply emit a config change event.

packages/client/src/provider/in-memory-provider/in-memory-provider.ts
toddbaert: I did.

beeme1mr (approved):

jonathannorris (approved):

lukas-reining (approved):
LGTM!

packages/react/src/use-feature-flag.ts
```diff
@@ -5,10 +5,16 @@ import { useOpenFeatureClient } from './provider';
 type ReactFlagEvaluationOptions = {
   /**
    * Suspend flag evaluations while the provider is not ready.
-   * Set to false if you don't want to use React Suspense API.
+   * Set to false if you don't want to show suspense fallbacks util the provider is initialized.
```
toddbaert, line 8:
```suggestion
   * Set to false if you don't want to show suspense fallbacks until the provider is initialized.
```

PR #800, opened 2024-01-31
fix: removed duped core types

toddbaert:
A while back I added:

```json
    "paths": {
      "@openfeature/core": [ "./packages/shared/src" ]
    },      
```

To the tsconfigs, which helped with local dev by always resolving `src/*` files from sibling packages instead of anything in their `dist/`.

It also had the nasty side-effect of causing our type bundling to include types from this package in it's output, causing duplicated artifacts and possible compiler issues. I've overridden this with a custom `tsconfig.rollup.json`.

Before this fix, the web/server SDK dists contained dupes of the stuff in shared:

```ts
import EventEmitter from 'events';

type FlagValueType = 'boolean' | 'string' | 'number' | 'object';
type PrimitiveValue = null | boolean | string | number;
type JsonObject = {
    [key: string]: JsonValue;
};
type JsonArray = JsonValue[];
/**
 * Represents a JSON node value.
 */
```

After this fix, they import them:

```ts
import { BaseHook, HookHints, EvaluationDetails, JsonValue, EvaluationLifeCycle, ManageLogger, Eventing, ClientMetadata, ProviderStatus, ClientProviderEvents, GenericEventEmitter, CommonEventDetails, FlagValue, CommonProvider, EvaluationContext, Logger, ResolutionDetails, EventHandler, OpenFeatureCommonAPI, ManageContext } from '@openfeature/core';
export * from '@openfeature/core';
```

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -1,7 +1,6 @@
 // this config rolls up all the types in the project to a single declaration (d.ts) file.
 // we do NOT use rollup to build (we use esbuild for that)
 
-import alias from '@rollup/plugin-alias';
 import dts from 'rollup-plugin-dts';
 
 export default {
@@ -18,6 +17,7 @@ export default {
     }
   ],
   plugins: [
-    dts({tsconfig: './tsconfig.json'}),
+    // use the rollup override tsconfig (applies equivalent in each sub-packages as well)
+    dts({tsconfig: './tsconfig.rollup.json'}),
   ],
 };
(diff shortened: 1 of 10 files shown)


rollup.config.mjs
```diff
@@ -1,7 +1,6 @@
 // this config rolls up all the types in the project to a single declaration (d.ts) file.
 // we do NOT use rollup to build (we use esbuild for that)
 
-import alias from '@rollup/plugin-alias';
```
beeme1mr, line 4: Is this no longer being used? Can we remove it as a dependency?

toddbaert: yep

beeme1mr (approved):

lukas-reining (approved):
Ah this makes sense. I also saw this when looking at the shipped types lately but did not realize that this would clash.
The change looks good!

PR #805, opened 2024-01-31
feat: add support for domains

### LINKED ISSUE #820: [BUG] named context not working for web provider
### Observed behavior

Following is my code:

```ts
  const clientName = 'my-client';
  await OpenFeature.setContext(clientName, { user: "mike" });
  OpenFeature.setProvider(
    clientName,
    new FlagdWebProvider({
      host: "localhost",
      port: 8013,
      tls: false,
    })
  );

  const client = OpenFeature.getClient(clientName);

    const flag = client.getBooleanValue("my-flag", false);

    const flagElm = document.querySelector('#flag-value');
    if (flagElm) {
      flagElm.innerHTML = flag.toString();
    }
  });
```

and `flags.json`

```json
{
  "$schema": "https://flagd.dev/schema/v0/flags.json",
  "flags": {
    "my-flag": {
      "state": "ENABLED",
      "variants": {
        "on": true,
        "off": false
      },
      "defaultVariant": "off",
      "targeting": {
        "if": [
          {
            "===": [
              {
                "var": "user"
              },
              "mike"
            ]
          },
          "on",
          "off"
        ]
      }
    }
  }
}
```

### Expected Behavior

I have set the named context and provider, and created a named client, if the context is working, I should get the `on` variant, but I got `off`, after open DevTools, the context didn't pass to flagd service.

image

When I use the empty name for context, provider, and client, the context is passed to flagd service as my expectation.

image

So I guess the named context is currently not working.
 

### Steps to reproduce

I have a reproduced repo

https://github.com/wellwind/flagd-context-demo

in `src/main.ts`，if I set `clientName` to any non empty name, the flag result will be `false`, when I set `clientName` to `''`, I can get the `true` result.

beeme1mr:
## This PR

- adds domain as a concept to the server and web SDK
- adds a deprecation warning anywhere client name was exposed to users.
- fixes an issue in the web SDK where context set on a domain before a provider is registered was not used.

## Addresses

fixes #820 https://github.com/open-feature/js-sdk/pull/805/commits/4aa965721ff189a4b2a659ff92ae8114213930b0

### Notes

This change is based on [this spec](https://github.com/open-feature/spec/pull/229) change. I tried to make it a non-breaking change but I may have missed an untested condition. Please carefully review to make sure I didn't miss anything.

### Follow-up Tasks

- Update the doc readme parser to support "domain".
- Update the NestJS and React SDKS. We should consider making those a breaking change since they're sub 1.0.


toddbaert:
Should we actually consider this a `feat`, considering it doesn't REALLY add new features? Maybe it's more of a chore? I guess you could argue enhanced clarity is some kind of feature?

beeme1mr:
I originally opened this as a refactor but we do add domain to event payloads. That's why I changed it to a feature. I'm good either way though. Whatever would be the most clear to users reading the release notes.

packages/client/src/client/open-feature-client.ts
```diff
@@ -44,7 +48,9 @@ export class OpenFeatureClient implements Client {
 
   get metadata(): ClientMetadata {
     return {
-      name: this.options.name,
+      // Use domain if name is not provided
+      name: this.options.domain ?? this.options.name,
+      domain: this.options.domain ?? this.options.name,
```
lukas-reining, line 53: Maybe we should throw in the constructor if name and domain are supplied?
I am not sure but it could be confusing if someone outs in a name and a domain and the domain is set as name.
This would not break existing implementations but also keep people from using both.

lukas-reining (approved):
Looks good, left one thought.

packages/client/src/client/open-feature-client.ts
toddbaert: I could be missing something, but people shouldn't be constructing `OpenFeatureClient` anyway, and with your recent change they won't be able to, so I don't think this matters so much. This is basically internal, used by `OpenFeature.getClient()`, which just takes `domain` as a string now instead of name.

packages/shared/src/events/eventing.ts
```diff
@@ -6,7 +6,11 @@ export type EventMetadata = {
 
 export type CommonEventDetails = {
   providerName: string;
+  /**
+   * @deprecated Use `domain` instead
+   */
```
toddbaert, line 11:
```suggestion
  /**
   * @deprecated alias of "domain", use domain instead
   */
```
I think this implies we'll set both so migration is safe.

packages/shared/src/client/client.ts
```diff
@@ -2,7 +2,11 @@ import { Metadata } from '../types';
 import { ProviderMetadata } from '../provider/provider';
 
 export interface ClientMetadata extends Metadata {
-  readonly version?: string;
+  /**
+   * @deprecated Use `domain` instead.
```
toddbaert, line 6:
```suggestion
   * @deprecated alias of "domain", use domain instead
```

packages/shared/src/events/eventing.ts
```diff
@@ -6,7 +6,11 @@ export type EventMetadata = {
...
   clientName?: string;
+  domain?: string;
```
toddbaert, line 13:
```suggestion
  readonly domain?: string;
```

All these props should be `readonly` I think (build/test passes fine when I do that), but we'll probably have to wait for 2.0 to fix that. Might as well fix this one now though I guess?

toddbaert:
@beeme1mr I found a few missing test assertions of `domain` where we were already asserting `clientName`, so I added them, but I can't see anything else. I did leave a couple minor suggestions above.

toddbaert (approved):

packages/client/README.md
```diff
@@ -89,16 +89,16 @@ See [here](https://open-feature.github.io/js-sdk/modules/_openfeature_web_sdk.ht
...
+| ------ | ----------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| ✅      | [Providers](#providers)             | Integrate with a commercial, open source, or in-house feature management tool.                                                     |
+| ✅      | [Targeting](#targeting-and-context) | Contextually-aware flag evaluation using [evaluation context](https://openfeature.dev/docs/reference/concepts/evaluation-context). |
+| ✅      | [Hooks](#hooks)                     | Add functionality to various stages of the flag evaluation life-cycle.                                                             |
+| ✅      | [Logging](#logging)                 | Integrate with popular logging packages.                                                                                           |
+| ✅      | [Domains](#domains)                 | Logically bind clients with providers  application.                                                                                |
```
federicobond, line 98: This reads a little awkward with the application at the end. Did you mean: *Logically bind clients with providers.*?

packages/client/src/client/open-feature-client.ts
toddbaert: I think you're probably right about that, ya.

packages/client/README.md
toddbaert: I feel like this must be a typo.

beeme1mr: Yep, fixed. I just dropped "application" from the end.

packages/client/src/client/open-feature-client.ts
beeme1mr: Yeah, I considered that but it could technically be viewed as a breaking change if someone was using the class. I'm happy to remove it if we think that's okay.

beeme1mr: What do you think @lukas-reining @toddbaert?

toddbaert: If you want to merge this now, you should keep it fully non-beaking I think.

lukas-reining: Yes then lets prefer to release it and not remove the name from the options.
We can remove it later then when we do the breaking 2.0 release. Especially as the `OpenFeatureClient` class will then be removed from the exports anyways.
As we discussed already both changes should not be breaking conceptually but we can not guarantee technically.

packages/client/README.md
```diff
@@ -205,26 +205,29 @@ const client = OpenFeature.getClient();
...
-import { OpenFeature } from "@openfeature/web-sdk";
+import { OpenFeature, InMemoryProvider } from "@openfeature/web-sdk";
 
 // Registering the default provider
-OpenFeature.setProvider(NewLocalProvider());
-// Registering a named provider
-OpenFeature.setProvider("clientForCache", new NewCachedProvider());
+OpenFeature.setProvider(InMemoryProvider(myFlags));
+// Registering a provider to a domain
+OpenFeature.setProvider("in-memory", new InMemoryProvider(someOtherFlags));
```
federicobond, line 220: The domain in this example suggests using the provider name. I would try to lead users away from this practice towards naming scopes. See https://github.com/open-feature/spec/pull/229#issuecomment-1891083384

packages/client/README.md
```diff
@@ -205,26 +205,29 @@ const client = OpenFeature.getClient();
...
 
-### Named clients
+### Domains
 
-Clients can be given a name.
-A name is a logical identifier that can be used to associate clients with a particular provider.
-If a name has no associated provider, the global provider is used.
+Clients can be assigned to a domain.
+A domain is a logical identifier which can be used to associate clients with a particular provider.
+If a domain has no associated provider, the global provider is used.
```
federicobond, line 212: Should we use "default provider" instead of "global provider" here?

packages/client/README.md
beeme1mr: I changed it but kept the term generic.

beeme1mr:
I've addressed all the feedback and included a bug fix. I'll merge this tomorrow unless I hear any objections. Thanks!

PR #816, opened 2024-02-14
fix: improve event handler types

toddbaert:
Before this change the `flagsChanged` prop on `EventDetails` was of type unknown for all event types, and a cast was needed.

After this change, `flagsChanged` is ONLY defined on `ProviderEvents.ConfigurationChanged`, where it should be, as (`string[] | undefined`) . In other words:

```typescript
client.addHandler(ProviderEvents.Ready, (details) => {
  expect(details?.flagsChanged?.length).toEqual(1);  // does not compile
});
```

```typescript
client.addHandler(ProviderEvents.ConfigurationChanged, (details) => {
  expect(details?.flagsChanged?.length).toEqual(1);  // DOES compile
});
```

Additionally, `flagsChanged` does not autocomplete for events handlers other than `ProviderEvents.ConfigurationChanged`.

### DIFF
--- a/packages/shared/src/events/eventing.ts
+++ b/packages/shared/src/events/eventing.ts
@@ -1,5 +1,5 @@
 import { ErrorCode } from '../evaluation';
-import { AllProviderEvents, AnyProviderEvent } from './events';
+import { ClientProviderEvents, ServerProviderEvents, AnyProviderEvent } from './events';
 
 export type EventMetadata = {
   [key: string]: string | boolean | number;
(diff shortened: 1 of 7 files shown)


packages/shared/src/events/eventing.ts
```diff
@@ -19,40 +19,62 @@ export type ErrorEvent = CommonEventProps;
...
-export type EventHandler = (eventDetails?: EventDetails) => Promise<unknown> | unknown;
+export type EventDetails<
+  T extends ServerProviderEvents | ClientProviderEvents = ServerProviderEvents | ClientProviderEvents,
+> = EventContext<Record<string, unknown>, T> & CommonEventDetails;
+export type EventHandler<
+  T extends ServerProviderEvents | ClientProviderEvents = ServerProviderEvents | ClientProviderEvents,
+> = (eventDetails?: EventDetails<T>) => Promise<unknown> | unknown;
 
-export interface Eventing {
+export interface Eventing<T extends ServerProviderEvents | ClientProviderEvents> {
```
toddbaert, line 54: Made this generic to support either client OR server events in the implementation.

packages/shared/src/events/eventing.ts
```diff
@@ -19,40 +19,62 @@ export type ErrorEvent = CommonEventProps;
...
+
+export type NotServerChangeEvents = ServerProviderEvents.Ready | ServerProviderEvents.Error | ServerProviderEvents.Stale;
+export type NotClientChangeEvents = ClientProviderEvents.Ready | ClientProviderEvents.Error | ClientProviderEvents.Stale | ClientProviderEvents.ContextChanged;
+export type NotChangeEvents = NotServerChangeEvents | NotClientChangeEvents;
+
+export type EventContext<
+  U extends Record<string, unknown> = Record<string, unknown>,
+  T extends ServerProviderEvents | ClientProviderEvents = ServerProviderEvents | ClientProviderEvents,
+> = (T extends ClientProviderEvents ? ClientEventMap[T] : T extends ServerProviderEvents ? ServerEventMap[T] : never) &
+  U;
```
toddbaert, line 45: I think this is the first time I've used a conditional type.

packages/shared/src/events/eventing.ts
```diff
@@ -19,40 +19,62 @@ export type ErrorEvent = CommonEventProps;
...
    * The handlers are called in the order they have been added.
-   * @param {AnyProviderEvent} eventType The provider event type to listen to
+   * @param eventType The provider event type to listen to
    * @param {EventHandler} handler The handler to run on occurrence of the event type
    */
-  addHandler(eventType: AnyProviderEvent, handler: EventHandler): void;
+  addHandler(
+    eventType: ClientProviderEvents.ConfigurationChanged | ServerProviderEvents.ConfigurationChanged,
+    handler: EventHandler<ClientProviderEvents.ConfigurationChanged | ServerProviderEvents.ConfigurationChanged>,
+  ): void;
```
lukas-reining, line 64: This means that eventType will never be more specific than  `ClientProviderEvents.ConfigurationChanged | ServerProviderEvents.ConfigurationChanged` which can be seen in the test: 

Bildschirmfoto 2024-02-14 um 21 31 49

Maybe we could go for a conditional type for the generic `T` of the eventing? It is a bit confusing to see the client types in the server.

This could look like this (I just had time to sketch this, we should maybe create explicit types for these inline conditionals): 
```typescript
export interface Eventing<T extends ServerProviderEvents | ClientProviderEvents> {
  /**
   * Adds a handler for the given provider event type.
   * The handlers are called in the order they have been added.
   * @param eventType The provider event type to listen to
   * @param {EventHandler} handler The handler to run on occurrence of the event type
   */
  addHandler(
    eventType: T extends ServerProviderEvents
      ? ServerProviderEvents.ConfigurationChanged
      : ClientProviderEvents.ConfigurationChanged,
    handler: EventHandler<
      T extends ServerProviderEvents
        ? ServerProviderEvents.ConfigurationChanged
        : ClientProviderEvents.ConfigurationChanged
    >,
  ): void;

  addHandler(
    eventType: T extends ServerProviderEvents ? NotServerChangeEvents : NotClientChangeEvents,
    handler: EventHandler<T extends ServerProviderEvents ? NotServerChangeEvents : NotClientChangeEvents>,
  ): void;

  /**
   * Removes a handler for the given provider event type.
   * @param {AnyProviderEvent} eventType The provider event type to remove the listener for
   * @param {EventHandler} handler The handler to remove for the provider event type
   */
  removeHandler(eventType: T, handler: EventHandler<T>): void;

  /**
   * Gets the current handlers for the given provider event type.
   * @param {AnyProviderEvent} eventType The provider event type to get the current handlers for
   * @returns {EventHandler[]} The handlers currently attached to the given provider event type
   */
  getHandlers(eventType: T): EventHandler<T>[];
}
```
Bildschirmfoto 2024-02-14 um 21 31 49

lukas-reining (approved):
Looks good, left one comment :)

packages/shared/src/events/eventing.ts
toddbaert: Oh that makes sense. I like it. I will give that a shot, maybe with aliases as you say.

beeme1mr (approved):

toddbaert:
I'm going to rebase this on https://github.com/open-feature/js-sdk/pull/795 when it's merged.

packages/shared/src/events/eventing.ts
toddbaert: I fixed this, and it doesn't even need to be generic here. I did so by putting a ternary like the one you show here in the `Eventing` interface itself, and then this implementation simply can restrict it further to `ProviderEvents` (which is the package-specific enum for server/client).

toddbaert: @lukas-reining unfortunately GitHub is having [some issues](https://www.githubstatus.com/), so the latest changes aren't showing up on this PR yet, even though they are on the branch :exploding_head:

toddbaert: It's working now.

lukas-reining (approved):
Looks good to me!

beeme1mr (approved):

PR #837, opened 2024-02-27
feat: context propagation

lukas-reining:
## This PR

- removes experimental warning from context propagation js-docs
- improves typing of `setTransactionContext`
- adds `AsyncLocalStorageTransactionContextProvider` to server SDK
  - @beeme1mr @toddbaert I am not 100% sure on this one. 
     To me it makes much sense to add this to the server SDK as it uses the Node default way `async_hooks`/`async_local_storage` which is part of Node since Node 16.x. I expect almost every project using the feature, to build exactly this so I wanted to include it in the SDK.
     As we are using Node types anyways I do not see a problem here, but still we could leave this out as it couples the implementation closer to Node.

Before merging I will have to change the README.


packages/server/src/transaction-context/async-local-storage-transaction-context-provider.ts
```diff
@@ -0,0 +1,19 @@
+import { EvaluationContext } from '@openfeature/core';
+import { TransactionContext, TransactionContextPropagator } from './transaction-context';
+import { AsyncLocalStorage } from 'async_hooks';
```
beeme1mr, line 3: It looks like this is available as far back as Node v12.17.0 so I think it's fine to use. Originally, I thought we may need to lazy load when the class is instantiated.

packages/nest/src/open-feature.module.ts
beeme1mr, line 1: We'll just need to remember to release the service SDK before releasing this one.

beeme1mr (commented):
Looks good so far. I'll rereview once it's marked as ready.

packages/server/src/transaction-context/transaction-context.ts
```diff
@@ -25,30 +21,41 @@ export interface ManageTransactionContextPropagator<T> extends TransactionContex
...
   /**
-   * EXPERIMENTAL: Transaction context propagation is experimental and subject to change.
-   * The OpenFeature Enhancement Proposal regarding transaction context can be found [here](https://github.com/open-feature/ofep/pull/32).
-   *
    * Sets the transaction context using the registered transaction context propagator.
-   * @experimental
+   * Runs the {@link callback} function, in which the {@link transactionContext} will be available by calling
+   * {@link this#getTransactionContext}.
+   *
+   * The {@link TransactionContextPropagator} must persist the {@link transactionContext} and make it available
```
luizgribeiro, line 35: Should we state context precedence in this doct?
I discovered from source code that flag evaluation passed context shadows keys set from other contexts

lukas-reining: Mh good question.
I think we could, on the other hand we are not doing it for all the other context variants right?
Do you want to propose something?

beeme1mr: We could include a link to https://openfeature.dev/specification/sections/evaluation-context#requirement-323

beeme1mr (approved):

packages/server/src/transaction-context/transaction-context.ts
lukas-reining: Ya that's a good option @beeme1mr 🙂

luizgribeiro (approved):

Issue #845, opened 2024-03-05
[BUG] Removing skipLibCheck throws an error


gogbog:
### Observed behavior

We have a project where we want to use web-sdk and we already have a provider. When we try to use sdk in the project without the `--skipLibCheck` it doesn't compile.

It throws the following error:

```
Could not find a declaration file for module 'events'. 'relative-path/angular-first/node_modules/events/events.js' implicitly has an 'any' type.

Try `npm i --save-dev @types/events` if it exists or add a new declaration (.d.ts) file containing `declare module 'events';` [plugin angular-compiler]
```

By installing `npm i --save-dev @types/events`, I'm getting the following error:

```
TS2709: Cannot use namespace 'EventEmitter' as a type. [plugin angular-compiler]
node_modules/@openfeature/web-sdk/dist/types.d.ts:110:37:
protected readonly eventEmitter: EventEmitter;
```

### Expected Behavior

There is no compile error, even tho the skipLibCheck is false.

### Steps to reproduce

1. Create a new angular project - `ng new angular-test`
2. Install open-feature SDK - `npm i @openfeature/web-sdk`
3. Import Open Feature SDK in some TS file - `import { OpenFeature } from '@openfeature/web-sdk'`
4. Switch SkipLibCheck in tsconfig.json to **FALSE**

lukas-reining:
Hey @gogbog, thanks for submitting the issue.
I will have a look at this later.

toddbaert:
@gogbog version `0.4.15` is released and @lukas-reining and I believe it should fix your issue, but please confirm and close if it does!

This new version also has the benefit for totaling 220kB instead of 250kB

gogbog:
@toddbaert @lukas-reining Thanks for the fast response. Everything worked as expected. I'm closing the issue.

PR #846, opened 2024-03-05
fix: correct rollup to bundle all but core

toddbaert:
My rollup config was invalid, but clumsily working until a recent change in how we defined the event emitter type. The impact was the the bundled types for the web imported `from 'events'` instead of bundling it.

This PR fixes the mis-configuration (see comments).

Here is a screenshot showing the bundled types in the dist now (notice the import is gone and the emitter is in-lined, no other changes are present): 

image

In server, this type is available from node, so we DON'T need to bundle it, and in fact, SHOULD import it. This is easily done by importing it like `from 'node:events'`. I don't think it would be a problem if we bundled it anyway, but this is more correct.

Fixes: https://github.com/open-feature/js-sdk/issues/845

### DIFF
--- a/packages/server/src/events/open-feature-event-emitter.ts
+++ b/packages/server/src/events/open-feature-event-emitter.ts
@@ -1,5 +1,5 @@
 import { GenericEventEmitter } from '@openfeature/core';
-import EventEmitter from 'events';
+import { EventEmitter } from 'node:events';
 import { ProviderEvents } from './events';
 
 /**
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -7,17 +7,15 @@ export default {
   input: "./src/index.ts",
   output: {
     file: './dist/types.d.ts',
-    format: 'es', // module format doesn't really matter here since output i
+    format: 'es', // module format doesn't really matter here since output is types
   },
-  external: [
-    // function indicating which deps should be considered external: non-external deps will have their types bundled
-    (id) => {
-      // bundle 'events' types
-      return id !== 'events';
-    }
-  ],
+  // function indicating which deps should be considered external: external deps will NOT have their types bundled
+  external: (id) => {
+    // bundle everything but '@openfeature/core', which is a peer
+    return id === '@openfeature/core';
+  },
   plugins: [
     // use the rollup override tsconfig (applies equivalent in each sub-packages as well)
-    dts({tsconfig: './tsconfig.rollup.json'}),
+    dts({tsconfig: './tsconfig.rollup.json', respectExternal: true }),
   ],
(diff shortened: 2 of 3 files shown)


packages/server/src/events/open-feature-event-emitter.ts
```diff
@@ -1,5 +1,5 @@
 import { GenericEventEmitter } from '@openfeature/core';
-import EventEmitter from 'events';
+import { EventEmitter } from 'node:events';
```
toddbaert, line 2: The `node:` specifier here prevents this from being bundled in server.

rollup.config.mjs
```diff
@@ -7,17 +7,15 @@ export default {
...
     file: './dist/types.d.ts',
-    format: 'es', // module format doesn't really matter here since output i
+    format: 'es', // module format doesn't really matter here since output is types
+  },
+  // function indicating which deps should be considered external: external deps will NOT have their types bundled
+  external: (id) => {
+    // bundle everything but '@openfeature/core', which is a peer
+    return id === '@openfeature/core';
   },
-  external: [
```
toddbaert, line 12: Note this was an array containing a function before... that was wrong.

toddbaert:
@lukas-reining I think this should work... please verify however you want. Thanks!

rollup.config.mjs
```diff
@@ -7,17 +7,15 @@ export default {
...
-    // function indicating which deps should be considered external: non-external deps will have their types bundled
-    (id) => {
-      // bundle 'events' types
-      return id !== 'events';
-    }
-  ],
   plugins: [
     // use the rollup override tsconfig (applies equivalent in each sub-packages as well)
-    dts({tsconfig: './tsconfig.rollup.json'}),
+    dts({tsconfig: './tsconfig.rollup.json', respectExternal: true }),
```
beeme1mr, line 19: The docs mention that `respectExternal` is not recommended. Can you explain why this is required?

https://github.com/Swatinem/rollup-plugin-dts?tab=readme-ov-file#what-to-expect

toddbaert: It's what allows the `external` classifier above to work at all now.

If we don't `respectExternal`, it doesn't matter what we put in the function above - everything is considered external unless it's a relative import. This would result in the vents being bundled, as well as `core` being bundled, which we do not want.

beeme1mr (approved):

rollup.config.mjs
lukas-reining: Makes sense!

lukas-reining (approved):
Change looks good and works with a local angular test app for me.

PR #852, opened 2024-03-06
fix(types): conflicts with peer types

toddbaert:
With our recent changes in bundling, there's a new problem with the framework-specific SDKs (react + nestjs). These have peers, and the external function was bundling those peers as well. This change prevents us from bundling any peers (which is what we want, so we avoid conflicts with the consumers installed peers).

This has no runtime impact, and there was not changes to core, web, or server SDK dists. See image below for the difference in the react SDK:

image

### DIFF
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -61,6 +61,19 @@ Capabilities include:
 npm install --save @openfeature/nestjs-sdk
 ```
 
+#### npm
+
+```sh
+npm install --save @openfeature/nestjs-sdk
+```
+
+#### yarn
+
+```sh
+# yarn requires manual installation of the peer dependencies (see below)
+yarn add @openfeature/nestjs-sdk @openfeature/server-sdk @openfeature/core
+```
+
 #### Required peer dependencies
 
 The following list contains the peer dependencies of `@openfeature/nestjs-sdk` with its expected and compatible versions:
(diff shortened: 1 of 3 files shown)


packages/nest/README.md
```diff
@@ -61,6 +61,19 @@ Capabilities include:
 npm install --save @openfeature/nestjs-sdk
 ```
 
+#### npm
```
toddbaert, line 64: README change just to prompt a new release, but also we were missing the yarn section.

packages/react/README.md
```diff
@@ -62,11 +71,18 @@ In addition to the feature provided by the [web sdk](https://openfeature.dev/doc
 npm install --save @openfeature/react-sdk
 ```
 
+#### yarn
```
toddbaert, line 74: README change just to prompt a new release, but also we were missing the yarn section.

toddbaert:
@luizgribeiro @lukas-reining the bundled types for the nest SDK now are smaller, and the imports look like this:

```ts
import * as _nestjs_common from '@nestjs/common';
import { ExecutionContext, DynamicModule, NestInterceptor, CallHandler } from '@nestjs/common';
import { Provider, Logger, EvaluationContext as EvaluationContext$1, Hook, ServerProviderEvents, EventHandler, JsonValue, FlagValue } from '@openfeature/server-sdk';
export * from '@openfeature/server-sdk';
import { EvaluationContext } from '@openfeature/core';
import { Observable } from 'rxjs';
```

Which seems right to me. The file contents are now just types generated from OUR actual source code.

beeme1mr (approved):

lukas-reining (approved):
Looks good and does what it always should :)

PR #875, opened 2024-03-14
feat: remove node 16

lukas-reining:
## This PR

Removes Node 16 as it is deprecated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "homepage": "https://github.com/open-feature/js-sdk#readme",
   "engines": {
-    "node": ">=16"
+    "node": ">=18"
   },
   "devDependencies": {
     "@rollup/plugin-typescript": "^11.1.6",
--- a/.github/workflows/audit-pending-releases.yml
+++ b/.github/workflows/audit-pending-releases.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
-          node-version: 16
+          node-version: 18
           registry-url: "https://registry.npmjs.org"
           cache: 'npm'
 
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           registry-url: 'https://registry.npmjs.org'
-          node-version: 16
+          node-version: 18
           cache: 'npm'
 
       - name: Install
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -16,7 +16,6 @@ jobs:
     strategy:
       matrix:
         node-version:
-          - 16.x
           - 18.x
           - 20.x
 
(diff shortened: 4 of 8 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
   },
   "homepage": "https://github.com/open-feature/js-sdk#readme",
   "engines": {
-    "node": ">=16"
+    "node": ">=18"
```
beeme1mr, line 36: This arguably makes this a breaking change. What do you think about leaving it at 16 but taking the officially stance that we only support actively maintained versions of node?

lukas-reining: Mh thats a good question.
The problem is that we would have to wait for removing node 16 until the next release of a major would happen then. I guess this could be a pretty long time. On the other hand just breaking the SDK for this now also does not feel good.

Technically we could maybe separate provider Node version from SDK version to avoid this I guess.
In this case people using EOL Node 16 could still use the SDK but might not be able to use every provider.
This could be better than having a breaking change in the SDK.
On the other hand there will be a moment we will have to remove Node 16 but I am really not sure when this would be a good moment.

lukas-reining: So we could maybe change the Node version for contribs but keep this, what do you think @beeme1mr?
If we go for this I would still wonder when we should remove Node 16 from here.

beeme1mr: I'm mainly concerned about the sdk. Providers can specify their own minimum node version.

lukas-reining: So it might be an option to close this for now but merge https://github.com/open-feature/js-sdk-contrib/pull/797.
Still at some point we would then have to remove Node 16 support as it is EOL.

lukas-reining: What if we generally do it like OTEL here @beeme1mr? This way we would not have to have a new major if we remove a node version that is not supported anymore.

https://github.com/open-telemetry/opentelemetry-js?tab=readme-ov-file#supported-runtimes

toddbaert: I don't think we should consider it breaking if we remove EOL runtime versions. We can put that somewhere in the spec or something.

toddbaert (approved):

package.json
lukas-reining: > I don't think we should consider it breaking if we remove EOL runtime versions. We can put that somewhere in the spec or something.
> 
> 
> 
> Removing actively supported versions is a different matter, and I think there's a stronger argument for considering that breaking, but it doesn't apply here.

Yes I think this is fine.
@toddbaert should we add a section like the OTEL docs I linked?

toddbaert (changes requested):
Blocking this as per discussion with @lukas-reining @beeme1mr . We will have to decide on stability guarantees with EOL versions.

toddbaert:
I'll be merging this today, consistent with our guidelines [here](https://github.com/open-feature/community/blob/main/technical-guidelines.md#platform-support). The TC met about this issue and concluded that since this is a relatively new library and since Node 16 is well beyond its EOL, we will merge this and consider it a non-breaking change. I will add some copy in the release notes about it.

toddbaert (approved):

Issue #834, opened 2024-02-26
[BUG] Issue with setContext when updating @openfeature/core from 0.0.25 to 0.0.26


nguyennick197:
EDIT: just did some testing and this was working on version 0.0.25, but still breaks on 0.0.26

### Observed behavior

`setContext` is throwing an error after updating from 0.0.23 to 0.026. 

This is the error thrown: 
```
TypeError
Cannot read properties of undefined (reading 'entries')
```

Error happens when running this code:
```typescript
export const setOpenFeatureTargetingKey = async (uuid: string) =>
  await OpenFeature.setContext({
    targetingKey: uuid,
  })
```  

Error is pointing to this section of the code:
```typescript
const oldContext = this._context;
this._context = context;
const defaultContextNameProviders = Array.from(this._clientProviders.entries()).filter(([name]) => !this._namedProviderContext.has(name)).reduce((acc, [name, provider]) => {
   acc.push({ name, provider });
   return acc;
}, []);
```

### Expected Behavior

No error is thrown when downgrading back to 0.0.23 and context is set.

### Steps to reproduce

```typescript
import { LaunchDarklyClientProvider } from '@openfeature/launchdarkly-client-provider'
import { OpenFeature, InMemoryProvider } from '@openfeature/react-sdk'

export const initOpenFeature = () => {
  const ldClientSideId = process.env.VITE_LAUNCH_DARKLY_CLIENT_SIDE_ID

  if (ldClientSideId) {
    const ldOptions = {
      streaming: true,
    }
    const ldOpenFeatureProvider = new LaunchDarklyClientProvider(
      ldClientSideId,
      ldOptions
    )
    // @ts-expect-error Silences "'LaunchDarklyClientProvider' is not assignable to parameter of type 'Provider'"
    OpenFeature.setProvider('openFeatureClient', ldOpenFeatureProvider)
  } else {
    OpenFeature.setProvider(new InMemoryProvider())
  }
}

export const setOpenFeatureTargetingKey = async (uuid: string) =>
  await OpenFeature.setContext({
    targetingKey: uuid,
  })

export const clearOpenFeatureTargetingKey = async () =>
  await OpenFeature.clearContext()
```

nguyennick197:
Thanks for the response @toddbaert! I really appreciate your help with this.

I'm using npm and didn't receive warnings warnings about mismatched versions during the install. 

This is my package.json:
```javascript
{
  "name": “app-name”,
  "version": "0.1.0",
[...]

PR #876, opened 2024-03-14
docs: add peer dep explainer

toddbaert:
Adds a blurb out peer deps and lib development, installation.

@beeme1mr I also added a bit about `--legacy-peer-deps` after the closure of https://github.com/open-feature/js-sdk/issues/834

### DIFF
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -36,7 +36,7 @@
 </p>
 <!-- x-hide-in-docs-start -->
 
-[OpenFeature](https://openfeature.dev) is an open specification that provides a vendor-agnostic, community-driven API for feature flagging that works with your favorite feature flag management tool.
+[OpenFeature](https://openfeature.dev) is an open specification that provides a vendor-agnostic, community-driven API for feature flagging that works with your favorite feature flag management tool or in-house solution.
 
 <!-- x-hide-in-docs-end -->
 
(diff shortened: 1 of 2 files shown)


packages/client/README.md
```diff
@@ -261,6 +266,14 @@ await OpenFeature.close()
 
 ## Extending
 
+### Considerations
```
beeme1mr, line 269: Could you please move this to the bottom on the the extending section?

toddbaert: Sure, if you think it's better there.

juanparadox (approved):

packages/client/README.md
```diff
@@ -261,6 +266,14 @@ await OpenFeature.close()
...
+
+When developing a library based on OpenFeature components, it's important to list the `@openfeature/web-sdk` as a `peerDependency` of your package.
+This is a general best-practice when developing JavaScript libraries that have dependencies in common with their consuming application.
+Failing to do this can result in multiple copies of the OpenFeature SDK in the consumer, which can lead to type errors, and broken singleton behavior.
+The `@openfeature/core` package itself follows this pattern: the `@openfeature/web-sdk` has a peer dependency on `@openfeature/core`, and uses whatever copy of that module the consumer has installed (note that NPM installs peers automatically unless `--legacy-peer-deps` is set, while yarn does not, and PNPM does so based on it's configuration).
```
beeme1mr, line 274:
```suggestion
The `@openfeature/core` package itself follows this pattern: the `@openfeature/web-sdk` has a peer dependency on `@openfeature/core`, and uses whatever copy of that module the consumer has installed (note that NPM installs peers automatically unless `--legacy-peer-deps` is set, while yarn does not, and PNPM does so based on its configuration).
```

toddbaert: Not accepting because it'll conflict with the move, but I'll make this change.

beeme1mr (approved):

lukas-reining (approved):

PR #883, opened 2024-03-24
fix: return metadata for the bound provider in hookContext

federicobond:
Clients were incorrectly populating the hook context provider metadata field with metadata from the default provider instead of the one bound to the client.

### DIFF
--- a/packages/client/src/client/open-feature-client.ts
+++ b/packages/client/src/client/open-feature-client.ts
@@ -201,7 +201,7 @@ export class OpenFeatureClient implements Client {
       defaultValue,
       flagValueType: flagType,
       clientMetadata: this.metadata,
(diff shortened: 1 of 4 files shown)


beeme1mr:
Good catch! Could you please add some tests so that we can ensure we don't accidentally introduce a regression?

toddbaert:
> Good catch! Could you please add some tests so that we can ensure we don't accidentally introduce a regression?

Agreed on this one, but if you have any trouble let me know and I'll happily add them myself; I know you're usually our "Python guy" so feel free to let me know if adding those tests is too tedious!

federicobond:
Would something like this work? I wasn't sure where to put the mock provider definition.

beeme1mr (approved):
Thanks

toddbaert (approved):

PR #896, opened 2024-04-01
feat: allow constraint of booleans

toddbaert:
Support constraining boolean evaluations. This probably seems weird, but it helps avoid a bunch of ugly and dangerous type assertions with some upcoming react SDK features, and it can't hurt anything.

Fully backwards compatible, new tests added.

We CAN avoid this if we want, but it makes things consistent and again, avoids some risky type assertions going forward.

### DIFF
--- a/packages/client/src/client/open-feature-client.ts
+++ b/packages/client/src/client/open-feature-client.ts
@@ -107,16 +107,27 @@ export class OpenFeatureClient implements Client {
     return this;
   }
 
-  getBooleanValue(flagKey: string, defaultValue: boolean, options?: FlagEvaluationOptions): boolean {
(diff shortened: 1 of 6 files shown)


beeme1mr (approved):

lukas-reining:
Hey @toddbaert, do you have an example where a value would be guaranteed to be true or false? What value would a flag have then?
Or am I missing another type for boolean that T could have? For boolean it seems to me that the type param is guaranteeing the value which seems counter intuitive to me to have.

toddbaert:
Pausing this for a bit because I'm not 100% sure it's helpful after all. We can still add it if we feel like it but I don't think it's going to help the way I thought it might.

toddbaert:
yep, closing.

PR #897, opened 2024-04-02
feat: query-style, generic useFlag hook

toddbaert:
This PR adds a new `useFlag` hook. This hook:

- returns an object supporting a react "query-style" API with relevant properties: `const { value: newMessage, isError, reason, errorCode, isAuthoritative, type } = useFlag('new-message', true);`
- supports ergonomic generics: 
```
const { value: boolVal }: FlagQuery<boolean> = useFlag('bool-flag', true);
const { value: stringVal }: FlagQuery<string> = useFlag('string-flag', 'string');
const { value: objVal }: FlagQuery<{greeting: string}> = useFlag('obj-flag', { greeting: 'hi' });
```
- supports all the same options and features as other hooks (suspense, etc)

This PR fixes some type bugs as well, and adds to the readme.

DEPENDS ON: https://github.com/open-feature/js-sdk/pull/898

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -126,11 +126,11 @@ function App() {
 
 function Page() {
   // Use the "query-style" flag evaluation hook.
-  const { value: newMessage } = useFlag('new-message', true);
+  const { value: newMessageFlagValue } = useFlag('new-message', true);
   return (
     <div className="App">
       <header className="App-header">
-        {newMessage ? <p>Welcome to this OpenFeature-enabled React app!</p> : <p>Welcome to this React app.</p>}
+        {newMessageFlagValue ? <p>Welcome to this OpenFeature-enabled React app!</p> : <p>Welcome to this React app.</p>}
       </header>
     </div>
   )
@@ -191,7 +191,7 @@ You can disable this feature in the hook options:
 
 ```tsx
 function Page() {
-  const newMessage = useBooleanFlagValue('new-message', false, { updateOnContextChanged: false });
+  const newMessageFlagValue = useBooleanFlagValue('new-message', false, { updateOnContextChanged: false });
   return (
     <MyComponents></MyComponents>
   )
@@ -208,7 +208,7 @@ You can disable this feature in the hook options:
 
 ```tsx
 function Page() {
-  const newMessage = useBooleanFlagValue('new-message', false, { updateOnConfigurationChanged: false });
+  const newMessageFlagValue = useBooleanFlagValue('new-message', false, { updateOnConfigurationChanged: false });
   return (
     <MyComponents></MyComponents>
   )
@@ -235,11 +235,11 @@ function Content() {
 
 function Message() {
   // component to render after READY.
-  const newMessage = useBooleanFlagValue('new-message', false);
+  const newMessageFlagValue = useBooleanFlagValue('new-message', false);
 
   return (
     <>
-      {newMessage ? (
+      {newMessageFlagValue ? (
         <p>Welcome to this OpenFeature-enabled React app!</p>
       ) : (
         <p>Welcome to this plain old React app!</p>
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -138,7 +138,7 @@ function Page() {
 
 export default App;
 ```
-You use the strongly-typed flag value and flag evaluation details hooks as well, if you prefer.
+You can use the strongly-typed flag value and flag evaluation detail hooks as well, if you prefer.
 
 ```tsx
 import { useBooleanFlagValue } from '@openfeature/react-sdk';
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -87,10 +87,12 @@ The following list contains the peer dependencies of `@openfeature/react-sdk` wi
 
 ### Usage
 
+The `OpenFeatureProvider` represents a scope for feature flag evaluations within a react application.
+It binds an OpenFeature client to all evaluations within child components, and allows the use of evaluation hooks.
 The example below shows how to use the `OpenFeatureProvider` with OpenFeature's `InMemoryProvider`.
 
 ```tsx
-import { EvaluationContext, OpenFeatureProvider, useBooleanFlagValue, useBooleanFlagDetails, OpenFeature, InMemoryProvider } from '@openfeature/react-sdk';
+import { EvaluationContext, OpenFeatureProvider, useFlag, OpenFeature, InMemoryProvider } from '@openfeature/react-sdk';
 
 const flagConfig = {
   'new-message': {
@@ -109,8 +111,11 @@ const flagConfig = {
   },
 };
 
+// Instantiate and set our provider (be sure this only happens once)!
+// Note: there's no need to await it's initialization, the React SDK handles re-rendering and suspense for you!
 OpenFeature.setProvider(new InMemoryProvider(flagConfig));
 
+// Enclose your content in the configured provider
 function App() {
   return (
     <OpenFeatureProvider>
@@ -120,7 +125,8 @@ function App() {
 }
 
 function Page() {
-  const newMessage = useBooleanFlagValue('new-message', false);
+  // Use the "query-style" flag evaluation hook.
+  const { value: newMessage } = useFlag('new-message', true);
   return (
     <div className="App">
       <header className="App-header">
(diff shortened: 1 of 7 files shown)


packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -52,6 +53,57 @@ enum SuspendState {
...
+  T extends boolean
+    ? boolean
+    : T extends number
+      ? number
+      : T extends string
+        ? string
+        : T extends JsonValue
+          ? T
+          : JsonValue
+>;
```
toddbaert, line 67: This type is a bit wild-looking, we need it. We have to use the conditional, because otherwise `useFlag('key', false)` would return `false`, not `boolean` (too constrained).

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -52,6 +53,57 @@ enum SuspendState {
...
+): FlagQuery<
+T extends boolean
+  ? boolean
+  : T extends number
+    ? number
+    : T extends string
+      ? string
+      : T extends JsonValue
+        ? T
+        : JsonValue
```
toddbaert, line 92: I intentionally didn't use the type alias above (`ConstrainedFlagQuery`) here because this naming is better and TS users see it.

packages/react/src/query/query.ts
```diff
@@ -0,0 +1,58 @@
+import { ErrorCode, EvaluationDetails, FlagMetadata, FlagValue, StandardResolutionReasons } from '@openfeature/core';
+
+export interface FlagQuery<T extends FlagValue = FlagValue> {
```
toddbaert, line 3: I tried to strike a balance here between adding too much and also being useful.

It mostly adds everything on `EvaluationDetails`, plus some stuff more useful to React usecases and this hook (type, since hook is generic).

packages/react/src/query/query.ts
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * A message associated with the error.
+   */
+  readonly errorMessage: string | undefined;
+
+  /**
+   * Indicates this flag is up-to-date and in sync with the source of truth.
+   * Specifically, indicates the evaluation did not default due to error, and the reason is neither {@link StandardResolutionReasons.CACHED} or {@link StandardResolutionReasons.DISABLED}.
+   */
+  readonly isAuthoritative: boolean;
```
toddbaert, line 52: @moredip

packages/react/README.md
```diff
@@ -120,7 +125,8 @@ function App() {
 }
 
 function Page() {
-  const newMessage = useBooleanFlagValue('new-message', false);
+  // Use the "query-style" flag evaluation hook.
+  const { value: newMessage } = useFlag('new-message', true);
```
toddbaert, line 129: This README promotes the new hook first, which I think is good, LMK what you think.

packages/react/src/evaluation/use-feature-flag.ts
beeme1mr: We need [NoInfer](https://devblogs.microsoft.com/typescript/announcing-typescript-5-4/#the-noinfer-utility-type) but that would require TypeScript 5.4. Maybe we could use a polyfill for now.

packages/react/src/query/query.ts
beeme1mr: Why would `CACHED` be non-authoritative? What about `STALE`?

packages/react/src/evaluation/use-feature-flag.ts
toddbaert: Maybe, but if you look closely, we actually DO want to infer that last condition, so that specific object generic args work nicely:

```
const { value: objVal }: FlagQuery<{greeting: string}> = useFlag('obj-flag', { greeting: 'hi' }); // we don't want a `JsonValue` here, we want a `T` (in this case a `greeting: string}`.
```

packages/react/README.md
```diff
@@ -87,10 +87,12 @@ The following list contains the peer dependencies of `@openfeature/react-sdk` wi
 
 ### Usage
 
+The `OpenFeatureProvider` represents a scope for feature flag evaluations within a react application.
```
juanparadox, line 90:
```suggestion
The `OpenFeatureProvider` represents a scope for feature flag evaluations within a React application.
```

packages/react/README.md
```diff
@@ -109,8 +111,11 @@ const flagConfig = {
   },
 };
 
+// Instantiate and set our provider (be sure this only happens once)!
+// Note: there's no need to await it's initialization, the React SDK handles re-rendering and suspense for you!
```
juanparadox, line 115: Praise: Great call out!

packages/react/README.md
juanparadox: Question: Should we add why you might want to use this hook over others? Does this new hook render the others as deprecated or to-be deprecated?

juanparadox (commented):
Just some comments 😄

packages/react/src/query/query.ts
toddbaert: Oh darn, this is my bad. I actually realized during this PR that we don't have the `STALE` reason pre-enumerated in the JS-SDK. I just put cached here as a placeholder. I will open another PR for that.

toddbaert: https://github.com/open-feature/js-sdk/pull/898

packages/react/README.md
toddbaert: > Question: Should we add why you might want to use this hook over others? Does this new hook render the others as deprecated or to-be deprecated?

The reason is just for the react-style return object that enabled destructuring-assignment of a bunch of useful props, and the fact that it's totally generic. It's really just react-flavored sugar.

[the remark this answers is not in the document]
toddbaert: This is a destructuring assignment; `value` is the name of the field on the `FlagQuery` that stores that flag value, while `newMessage` is the var we are declaring - it's reverse from usual.

juanparadox: That makes sense, I just worry about devs not knowing if this hook is recommended over all of the others, when to use this hook, or if the other hooks are going away.

toddbaert: I've tried to outline this in the README, bit I could certain add more inline doc that makes it clear both APIs are fine to use. How does that sound?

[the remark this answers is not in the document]
toddbaert: Oh, I see, yes that makes sense I'll do that. I suppose it's good to be gratuitously obvious with sample code.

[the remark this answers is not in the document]
toddbaert: Done, see: https://github.com/open-feature/js-sdk/pull/897/commits/6938f9bdb51e5023503d2fbe4d9fa9fc89d8fd17

toddbaert: Done, see: https://github.com/open-feature/js-sdk/pull/897/commits/6938f9bdb51e5023503d2fbe4d9fa9fc89d8fd17

juanparadox (approved):

packages/react/package.json
```diff
@@ -46,7 +46,7 @@
   },
   "homepage": "https://github.com/open-feature/js-sdk#readme",
   "peerDependencies": {
-    "@openfeature/web-sdk": ">=1.0.0",
+    "@openfeature/web-sdk": "^1.0.2",
```
toddbaert, line 49: Need this because we use STALE.

packages/react/src/query/query.ts
toddbaert: This is STALE now.

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -336,3 +398,52 @@ function suspenseWrapper<T>(promise: Promise<T>) {
...
+    return this._details?.errorMessage;
+  }
+
+  get isAuthoritative() {
+    return (
+      !this.isError &&
+      this._details.reason != StandardResolutionReasons.STALE &&
+      this._details.reason != StandardResolutionReasons.DISABLED
+    );
+  }
```
toddbaert, line 444: @moredip this naming makes sense to me, since we specifically say in the spec, that STALE means:

> The resolved value is **_non-authoritative_** or possibly out of date

(emphasis mine).

beeme1mr (approved):

Issue #893, opened 2024-03-29
[FEATURE] Web-sdk: Access to provider status inside the provider


thomaspoignant:
### Requirements

With the web-sdk `v1.0.0`, the provider status is now managed directly by the SDK which is great, but there are some usecases when it will be interesting to have access to the provider status inside the SDK.

One of the use cases I have is being able to send a `CACHED` reason when the provider is `STALE`.

weyert:
Totally agree with this, probably worth doing a spec update for this?

thomaspoignant:
> Totally agree with this, probably worth doing a spec update for this?

Not sure what the others think but it could be a great idea to spec it.

thomaspoignant:
> You mean read-only access, I guess based on your use-case? I think I could be convinced of that. 🤔

Yes read-only, I don't need to change the status but I need to have access to it.

beeme1mr:
It might be slightly complicated to get the status due to `domains`. Perhaps we should quickly prototype it to see how it could be implemented and then used in a provider.

PR #898, opened 2024-04-02
feat: add STALE pre-defined reason

toddbaert:
Adds STALE pre-defined reason, re-orders reasons in source to match spec.

### DIFF
--- a/packages/shared/src/evaluation/evaluation.ts
+++ b/packages/shared/src/evaluation/evaluation.ts
@@ -39,6 +39,16 @@ export type EvaluationDetails<T extends FlagValue> = {
 
 export const StandardResolutionReasons = {
   /**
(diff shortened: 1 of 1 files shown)


packages/shared/src/evaluation/evaluation.ts
```diff
@@ -38,6 +38,16 @@ export type EvaluationDetails<T extends FlagValue> = {
 } & ResolutionDetails<T>;
 
 export const StandardResolutionReasons = {
```
toddbaert, line 40: I re-ordered these to be consistent with their listing in the spec.

beeme1mr (approved):

beeme1mr:
@toddbaert loosely related to https://github.com/open-feature/js-sdk/issues/893

thomaspoignant (approved):

PR #900, opened 2024-04-05
feat!: options inheritance, useWhenProviderReady, suspend by default

toddbaert:
* suspends while reconciling by default
* adds the ability to configure options at the context-provider level
* adds new `useWhenProviderReady` hook which force-suspend components until the provider is ready (especially good for React 17 shortcomings
* updates README with FAQ/troubleshooting and more


packages/react/src/common/options.ts
```diff
@@ -0,0 +1,39 @@
...
+   */
+  updateOnContextChanged?: boolean;
+} & FlagEvaluationOptions;
+
+export const DEFAULT_OPTIONS: ReactFlagEvaluationOptions = {
+  updateOnContextChanged: true,
+  updateOnConfigurationChanged: true,
+  suspendUntilReady: true,
+  suspendWhileReconciling: true,
+};
```
toddbaert, line 37: I changed the defaults here! Now suspense it automatic for both reconciliation and readiness. I think this is the best choice, but I could also be convinced we could set both to false.

Either way, I think both suspense settings should match.

packages/react/src/common/suspend.ts
```diff
@@ -0,0 +1,73 @@
+import { Client, ProviderEvents } from '@openfeature/web-sdk';
```
toddbaert, line 1: No changes to these functions, I just extracted them so we could use them in multiple hooks.

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -1,59 +1,19 @@
...
-import { Dispatch, SetStateAction, useEffect, useState } from 'react';
-import { useOpenFeatureClient } from '../provider';
+import { useEffect, useState } from 'react';
+import { ReactFlagEvaluationOptions, getDefaultedOptions } from '../common/options';
+import { suspend } from '../common/suspend';
+import { useProviderOptions } from '../provider/context';
+import { useOpenFeatureClient } from '../provider/use-open-feature-client';
 import { FlagQuery } from '../query';
 
-type ReactFlagEvaluationOptions = {
```
toddbaert, line 15: extracted into `common/`

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -338,69 +299,6 @@ function attachHandlersAndResolve<T extends FlagValue>(
   return resolver(client).call(client, flagKey, defaultValue);
 }
 
-/**
```
toddbaert, line 341: extracted into `common/`

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -284,7 +244,8 @@ function attachHandlersAndResolve<T extends FlagValue>(
   resolver: (client: Client) => (flagKey: string, defaultValue: T) => EvaluationDetails<T>,
   options?: ReactFlagEvaluationOptions,
 ): EvaluationDetails<T> {
-  const defaultedOptions = { ...DEFAULT_OPTIONS, ...options };
+  // highest priority > evaluation hook options > provider options > default options > lowest priority
```
toddbaert, line 247: I hope this is clear enough.

packages/react/src/provider/context.ts
```diff
@@ -0,0 +1,16 @@
...
+
+// DO NOT EXPORT PUBLICLY
+export const Context = React.createContext<{ client: Client; options: ReactFlagEvaluationOptions } | undefined>(undefined);
+
+/**
+ * Get a copy of the effective options used for this OpenFeatureProvider.
+ * DO NOT EXPORT PUBLICLY
+ * @returns {ReactFlagEvaluationOptions} options the defaulted options
+ */
+export function useProviderOptions() {
```
toddbaert, line 13: I didn't think this was a good one to export.

packages/react/src/provider/use-open-feature-client.ts
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { Context } from './context';
+
+/**
+ * Get the {@link Client} instance for this OpenFeatureProvider context.
+ * Note that the provider to which this is bound is determined by the OpenFeatureProvider's domain.
+ * @returns Client
+ */
+export function useOpenFeatureClient() {
```
toddbaert, line 9: Extracted.

@lukas-reining you originally wrote this, and it certainly makes sense internally, but I also think it's useful to consumers. Do you think we should continue to expose it? I think it could be useful.

packages/react/src/provider/use-when-provider-ready.ts
```diff
@@ -0,0 +1,20 @@
...
+import { useEffect, useState } from 'react';
+import { suspend } from '../common/suspend';
+import { useOpenFeatureClient } from './use-open-feature-client';
+
+/**
+ * Utility hook that triggers suspense until the provider is ready, without evaluating any flags.
+ * Especially useful for React v16/17 "Legacy Suspense", in which siblings to suspending components are
+ * initially mounted and then hidden (see: https://github.com/reactwg/react-18/discussions/7).
+ */
+export function useWhenProviderReady(): void {
```
toddbaert, line 11: New hook for "forcing" suspense. Mentioned in the FAQ.

packages/react/README.md
```diff
@@ -252,6 +261,37 @@ function Fallback() {
...
+In React 16/17's "Legacy Suspense", when a component suspends, it's sibling components initially mount and then are hidden.
+This can cause surprising effects and inconsistencies if they are render while the provider is still getting ready.
+To fix this, you can upgrade to React 18, which uses "Concurrent Suspense", in which siblings are not mounted until their suspended sibling resolves.
+Alternatively, if you cannot upgrade to React 18, you can use the `useWhenProviderReady` utility hook in any sibling components to force them from mounting until the provider is ready.
+
+> I am using multiple `OpenFeatureProvider` contexts, but they are sharing the same provider or evaluation context. Why?
+
+The `OpenFeatureProvider` binds a `client` to all child components, but the provider associated with that client is controlled by the `domain` parameter.
+This is consistent with all OpenFeature SDKs.
+To scope a provider to a particular client (or set of clients) set the `domain` parameter on your `OpenFeatureProvider`:
```
toddbaert, line 289:
```suggestion
To scope a provider to a particular provider/context set the `domain` parameter on your `OpenFeatureProvider`:
```

packages/react/README.md
```diff
@@ -252,6 +261,37 @@ function Fallback() {
...
+> I get odd rendering issues, or errors when components mount, if I use the suspense features.
+
+In React 16/17's "Legacy Suspense", when a component suspends, it's sibling components initially mount and then are hidden.
+This can cause surprising effects and inconsistencies if they are render while the provider is still getting ready.
+To fix this, you can upgrade to React 18, which uses "Concurrent Suspense", in which siblings are not mounted until their suspended sibling resolves.
+Alternatively, if you cannot upgrade to React 18, you can use the `useWhenProviderReady` utility hook in any sibling components to force them from mounting until the provider is ready.
+
+> I am using multiple `OpenFeatureProvider` contexts, but they are sharing the same provider or evaluation context. Why?
+
+The `OpenFeatureProvider` binds a `client` to all child components, but the provider associated with that client is controlled by the `domain` parameter.
```
toddbaert, line 287:
```suggestion
The `OpenFeatureProvider` binds a `client` to all child components, but the provider and context associated with that client is controlled by the `domain` parameter.
```

packages/react/README.md
```diff
@@ -252,6 +261,37 @@ function Fallback() {
...
+> I get an error that says something like: `A React component suspended while rendering, but no fallback UI was specified.`
+
+The OpenFeature React SDK feature built-in [suspense support](#suspense-support).
+This means that it will render your loading fallback automatically while the your provider starts up, and during context reconciliation for any of your components using feature flags!
+However, you will see this error if you neglect to create a suspense boundary around any components using feature flags; add a suspense boundary to resolve this issue.
+Alternatively, you can disable this feature by setting `suspendWhileReconciling=true` and `suspendUntilReady=true` in the [evaluation hooks](#evaluation-hooks) or the [OpenFeatureProvider](#openfeatureprovider-context-provider) (which applies to all evaluation hooks in child components).
+
+> I get odd rendering issues, or errors when components mount, if I use the suspense features.
+
+In React 16/17's "Legacy Suspense", when a component suspends, it's sibling components initially mount and then are hidden.
```
toddbaert, line 280:
```suggestion
In React 16/17's "Legacy Suspense", when a component suspends, its sibling components initially mount and then are hidden.
```

packages/react/README.md
```diff
@@ -252,6 +261,37 @@ function Fallback() {
...
+
+The OpenFeature React SDK feature built-in [suspense support](#suspense-support).
+This means that it will render your loading fallback automatically while the your provider starts up, and during context reconciliation for any of your components using feature flags!
+However, you will see this error if you neglect to create a suspense boundary around any components using feature flags; add a suspense boundary to resolve this issue.
+Alternatively, you can disable this feature by setting `suspendWhileReconciling=true` and `suspendUntilReady=true` in the [evaluation hooks](#evaluation-hooks) or the [OpenFeatureProvider](#openfeatureprovider-context-provider) (which applies to all evaluation hooks in child components).
+
+> I get odd rendering issues, or errors when components mount, if I use the suspense features.
+
+In React 16/17's "Legacy Suspense", when a component suspends, its sibling components initially mount and then are hidden.
+This can cause surprising effects and inconsistencies if they are render while the provider is still getting ready.
```
toddbaert, line 281:
```suggestion
This can cause surprising effects and inconsistencies if sibling components are rendered while the provider is still getting ready.
```

packages/react/src/provider/context.ts
toddbaert: I think there are some packages and rollup plugins etc that do this as a post compilation step, but in my experience they are tough to maintain.

packages/react/README.md
```diff
@@ -123,9 +128,15 @@ function App() {
...
+```
+
+#### Evaluation hooks
 
+Within the provider, you can use the various evaluation hooks to evaluate flags. 
+
+```tsx
 function Page() {
-  // Use the "query-style" flag evaluation hook.
+  // Use the "query-style" flag evaluation hook, specifying a flag-key and a default value.
```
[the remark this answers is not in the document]
toddbaert: We say that in the JS doc, I'm hesitant to make this too wordy.

packages/react/src/common/suspend.ts
```diff
@@ -0,0 +1,73 @@
...
+
+enum SuspendState {
+  Pending,
+  Success,
+  Error,
+}
+
+/**
+ * Suspend function. If this runs, components using the calling hook will be suspended.
+ * DO NOT EXPORT PUBLICLY
```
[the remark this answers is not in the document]
toddbaert: I was actually going to, but `@internal` doesn't autocomplete in my vscode, so I hesitated.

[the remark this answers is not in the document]
toddbaert: Apparently this is because it's official in TS, but not JSDoc. I will add it.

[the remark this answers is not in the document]
toddbaert: Added.

packages/react/src/common/options.ts
moredip: I don't know much about suspense - I haven't personally worked on a codebase that uses it yet.

What effect would this have for codebases that haven't opted into Suspense at all?

lukas-reining: > What effect would this have for codebases that haven't opted into Suspense at all?

This would result in a runtime error @moredip. If you do not have a suspense boundary `<Suspense ...>` above your use of a suspense enabled data source it would simply let the "thrown" error trickle up.

packages/react/src/provider/use-open-feature-client.ts
lukas-reining: > Do you think we should continue to expose it? I think it could be useful.

Yes I think noting speaks against it.
There might be a reason to just get the client directly, e.g. you do not have a suspense boundary or want to use a flag value in a callback that should evaluate the value on call (sure this can be done differently but it might be convenient to just pass the client.
And I guess it should also not be a problem regarding encapsulation or so. We also have this in Nest.js SDK.

packages/react/README.md
```diff
@@ -252,6 +261,37 @@ function Fallback() {
...
+This can be disabled in the hook options (or in the [OpenFeatureProvider](#openfeatureprovider-context-provider)).
+
+## FAQ and troubleshooting
+
+> I get an error that says something like: `A React component suspended while rendering, but no fallback UI was specified.`
+
+The OpenFeature React SDK features built-in [suspense support](#suspense-support).
+This means that it will render your loading fallback automatically while the your provider starts up, and during context reconciliation for any of your components using feature flags!
+However, you will see this error if you neglect to create a suspense boundary around any components using feature flags; add a suspense boundary to resolve this issue.
+Alternatively, you can disable this feature by setting `suspendWhileReconciling=true` and `suspendUntilReady=true` in the [evaluation hooks](#evaluation-hooks) or the [OpenFeatureProvider](#openfeatureprovider-context-provider) (which applies to all evaluation hooks in child components).
```
lukas-reining, line 276:
```suggestion
Alternatively, you can disable this feature by setting `suspendWhileReconciling=false` and `suspendUntilReady=false` in the [evaluation hooks](#evaluation-hooks) or the [OpenFeatureProvider](#openfeatureprovider-context-provider) (which applies to all evaluation hooks in child components).
```

If I am not totally misreading this, the options should be set to false right?

packages/react/src/common/options.ts
lukas-reining: I think these are the right defaults @toddbaert.
But I think I would like an option like `enableSuspense` that overwrites the fine grained controls.
An application that does not use suspense could then just set this prop of the provider and would not have to set two options.

packages/react/src/provider/provider.tsx
```diff
@@ -20,26 +22,18 @@ type ClientOrDomain =
...
-
-  if (!client) {
-    throw new Error(
-      'No OpenFeature client available - components using OpenFeature must be wrapped with an <OpenFeatureProvider>'
-    );
-  }
-
-  return client;
-};
+  return <Context.Provider value={{ client, options }}>{children}</Context.Provider>;
```
lukas-reining, line 38: I was wondering about having a default suspense boundary in the `OpenFeatureProvider`.
Thinking about it I think it is no good idea as the following

```ts
return (
  <App>
   <Suspense ...>
      <OpenFeatureProvider ... >
        <MyFeatureFlagConsumingComponent>
      </OpenFeatureProvider>
   </Suspense>
  </App>
```
 results in a higher level suspense not being triggered without e.g. disabling the default suspense from `OpenFeatureProvider`.
On the other hand we could add a suspense if a suspense component is given like:

```ts
return (
  <App>
      <OpenFeatureProvider ... onSuspense={<div>Loading...</div>} >
        <MyFeatureFlagConsumingComponent>
      </OpenFeatureProvider>
  </App>
```

In the end I think this only adds complexity to avoid folks using `Suspense` directly which is a React primitive.
Just thought a bit about it and would like to get others opinion @toddbaert.

lukas-reining (approved):
Looks great to me!
Left some thought about suspense.

packages/react/README.md
toddbaert: Absolutely! Nice catch.

packages/react/src/common/options.ts
toddbaert: Ya I can add that, as well as some typings that prevent nonsense configuration. Good idea.

packages/react/src/provider/provider.tsx
toddbaert: 🤔 

I'll think about this some more.

toddbaert: These are both good things to consider, but I think they take control aware from devs. I'm glad you brought them up though because they are possible alternatives to the existing API.

I think a built is suspense is not good for the reason you say: 

> higher level suspense not being triggered without e.g. disabling the default suspense from OpenFeatureProvider.

> `<OpenFeatureProvider ... onSuspense={<div>Loading...</div>} >`

This is a more compelling idea, but I think it also takes too much freedom away from users. Particularly, this would result in suspending everything inside the Context Provider, which will work for simple cases like the [react-sdk-demo](https://github.com/open-feature/react-test-app). It's not great however, when you have multiple components in a UI (for instance, multiple cards) and you want them to load independently. Some of our internal UI's using the React SDK work like this (1 or 2 smaller views in the large UI use flags and have their own suspense boundary); they load independently of other components when the provider becomes ready. I think doing "built-in" suspense as proposed above will discourage or confuse the pattern I describe, which I don't want.

packages/react/src/common/options.ts
toddbaert: I did this.

PR #901, opened 2024-04-05
fix: re-render w/ useWhenProviderReady, add tests

toddbaert:
This PR:

- brings react-sdk test coverage from 0% to ~95%
  - adds DOM-based testing (tests based on asserting DOM entity states)
    - tests for query-style, basic, and detailed evaluation APIs
    - tests for suspense functionality and re-rendering on context change
  - tests for some util functions and hooks
  - `renderHook` tests for non suspending hooks
- fixes a bug where `useWhenProviderReady` didn't cause re-render after the provider is ready if suspense wasn't used (leading to an out-of-date return value for provider readiness) 

image

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -171,6 +171,24 @@ export default {
         ],
       },
     },
+    {
+      displayName: 'react',
+      testEnvironment: 'jsdom',
+      preset: 'ts-jest',
+      testMatch: ['<rootDir>/packages/react/test/**/*.spec.ts*'],
+      moduleNameMapper: {
+        '@openfeature/core': '<rootDir>/packages/shared/src',
+        '@openfeature/web-sdk': '<rootDir>/packages/client/src',
+      },
+      transform: {
+        '^.+\\.tsx$': [
+          'ts-jest',
+          {
+            tsconfig: '<rootDir>/packages/react/test/tsconfig.json',
+          },
+        ],
+      },
+    },
   ],
 
   // Use this configuration option to add custom reporters to Jest
--- a/packages/react/src/provider/use-when-provider-ready.ts
+++ b/packages/react/src/provider/use-when-provider-ready.ts
@@ -18,12 +18,24 @@ export function useWhenProviderReady(options?: Options): boolean {
   const [, updateState] = useState<object | undefined>();
   const client = useOpenFeatureClient();
   // highest priority > evaluation hook options > provider options > default options > lowest priority
-  const defaultedOptions = { ...DEFAULT_OPTIONS, ...useProviderOptions(), ...normalizeOptions(options)};
+  const defaultedOptions = { ...DEFAULT_OPTIONS, ...useProviderOptions(), ...normalizeOptions(options) };
+  const updateStateRef = () => {
+    updateState({});
+  };
 
   useEffect(() => {
-    if (defaultedOptions.suspendUntilReady && client.providerStatus === ProviderStatus.NOT_READY) {
-      suspend(client, updateState, ProviderEvents.Ready);
+    if (client.providerStatus === ProviderStatus.NOT_READY) {
+      // re-render when provider is ready
+      client.addHandler(ProviderEvents.Ready, updateStateRef);
+      if (defaultedOptions.suspendUntilReady) {
+        // suspend and update when the provider is ready
+        suspend(client, updateState, ProviderEvents.Ready);
+      }
     }
+    return () => {
+      // cleanup the handler
+      client.removeHandler(ProviderEvents.Ready, updateStateRef);
+    };
   }, []);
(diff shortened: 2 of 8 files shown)


packages/react/test/evaluation.spec.tsx
```diff
@@ -0,0 +1,384 @@
...
+      });
+
+      // expect to see static value until we reconcile
+      await waitFor(() => expect(screen.queryByText(STATIC_FLAG_VALUE)).toBeInTheDocument(), { timeout: DELAY / 2 });
+
+      // make sure we updated after reconciling
+      await waitFor(() => expect(screen.queryByText(TARGETED_FLAG_VALUE)).toBeInTheDocument(), { timeout: DELAY * 2 });
+    });
+  });
+});
```
therynamo, line 384: **Suggestion**: It feels like a lot of these tests are creating `DOM` elements for the sake of verifying that the values expected to indeed evaluate from the `hook`s appropriately. Am I correct in that assumption? 

If so, I'm wondering if using the [`renderHook`](https://testing-library.com/docs/solid-testing-library/api/#renderhook) pattern to test the hooks explicitly may be a more straight forward and clear way to test these. 

The provider would still be valuable as you'd need to send that to `renderHook` as a `wrapper` component, but ultimately you could avoid having to build `TestComponent`, and variations of, in every component. 

In theory this would speed up tests over time as they'd be less concerned with rendering setup in headless `DOM`. 

Thoughts?

therynamo (commented):
This is awesome to see coverage getting added to the `js-sdk`. Super needed, especially as you said in pushing this past `experimental`.  

I had a few comments, nothing blocking, but overall this is 👍 x 💯 . Thanks for writing these! I always come into a library and look at tests and how the functions/components/etc. are used in order to get a feel for how the maintainer expects me to use their package. These types of efforts are non-trivial as well as helpful for all end-users. Thank you!

packages/react/test/evaluation.spec.tsx
```diff
@@ -0,0 +1,384 @@
...
+  async initialize(context?: EvaluationContext | undefined): Promise<void> {
+    await new Promise((resolve) => setTimeout(resolve, this.delay));
+    return super.initialize(context);
+  }
+
+  // artificially delay context changes
+  async onContextChange(): Promise<void> {
+    await new Promise((resolve) => setTimeout(resolve, this.delay));
+  }
+}
```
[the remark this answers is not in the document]
toddbaert: I've done this. Thanks!

packages/react/test/evaluation.spec.tsx
toddbaert: Hey @therynamo , I used this for the `useOpenFeatureClient` hook tests. It's a nice pattern!

packages/react/src/provider/use-when-provider-ready.ts
```diff
@@ -18,12 +18,24 @@ export function useWhenProviderReady(options?: Options): boolean {
...
-  const defaultedOptions = { ...DEFAULT_OPTIONS, ...useProviderOptions(), ...normalizeOptions(options)};
+  const defaultedOptions = { ...DEFAULT_OPTIONS, ...useProviderOptions(), ...normalizeOptions(options) };
+  const updateStateRef = () => {
+    updateState({});
+  };
 
   useEffect(() => {
-    if (defaultedOptions.suspendUntilReady && client.providerStatus === ProviderStatus.NOT_READY) {
-      suspend(client, updateState, ProviderEvents.Ready);
+    if (client.providerStatus === ProviderStatus.NOT_READY) {
```
toddbaert, line 27: Lacking

packages/react/src/provider/use-when-provider-ready.ts
```diff
@@ -18,12 +18,24 @@ export function useWhenProviderReady(options?: Options): boolean {
...
+      // re-render when provider is ready
+      client.addHandler(ProviderEvents.Ready, updateStateRef);
```
toddbaert, line 29: The lack of this handler caused the hook not to re-render if suspense wasn't used, which means a stale value for provider readiness.

lukas-reining (approved):
Looks great to me!

jest.config.ts
```diff
@@ -171,6 +171,24 @@ export default {
...
+      },
+      transform: {
+        '^.+\\.tsx$': [
+          'ts-jest',
+          {
+            tsconfig: '<rootDir>/packages/react/test/tsconfig.json',
+          },
+        ],
+      },
+    },
```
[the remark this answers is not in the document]
weyert: Vitest is pretty slick

weyert (approved):
Looks great! Thank you 🤩

PR #905, opened 2024-04-09
fix: default options (re-renders not firing by default)

toddbaert:
When I was adding the general `suspend` option as @lukas-reining suggested (a way to enable/disable all suspense features) I broke the default options. I found this writing tests in a different branch.

Tests for the entire SDK including this are incoming shortly (today), but this is a fix for the issue.

### DIFF
--- a/packages/react/src/common/options.ts
+++ b/packages/react/src/common/options.ts
@@ -64,14 +64,19 @@ export const DEFAULT_OPTIONS: ReactFlagEvaluationOptions = {
  */
 export const normalizeOptions: (options?: ReactFlagEvaluationOptions) => NormalizedOptions = (options?: ReactFlagEvaluationOptions) => {
   const defaultOptionsIfMissing = !options ? {} : options;
-  // fall-back the suspense options
+
+  const updateOnContextChanged = defaultOptionsIfMissing.suspendWhileReconciling;
+  const updateOnConfigurationChanged = defaultOptionsIfMissing.updateOnConfigurationChanged;
+
+  // fall-back the suspense options to the catch-all `suspend` property 
   const suspendUntilReady = 'suspendUntilReady' in defaultOptionsIfMissing ? defaultOptionsIfMissing.suspendUntilReady : defaultOptionsIfMissing.suspend;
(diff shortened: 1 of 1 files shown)


packages/react/src/common/options.ts
```diff
@@ -64,14 +64,19 @@ export const DEFAULT_OPTIONS: ReactFlagEvaluationOptions = {
  */
 export const normalizeOptions: (options?: ReactFlagEvaluationOptions) => NormalizedOptions = (options?: ReactFlagEvaluationOptions) => {
   const defaultOptionsIfMissing = !options ? {} : options;
```
lukas-reining, line 66: Just a nit, you can do `options ?? {}` here.

packages/react/src/common/options.ts
```diff
@@ -64,14 +64,19 @@ export const DEFAULT_OPTIONS: ReactFlagEvaluationOptions = {
...
   const suspendWhileReconciling = 'suspendWhileReconciling' in defaultOptionsIfMissing ? defaultOptionsIfMissing.suspendWhileReconciling : defaultOptionsIfMissing.suspend;
+
   return {
-    updateOnContextChanged: defaultOptionsIfMissing.updateOnContextChanged,
-    updateOnConfigurationChanged: defaultOptionsIfMissing.updateOnConfigurationChanged,
     // only return these if properly set (no undefined to allow overriding with spread)
     ...(typeof suspendUntilReady === 'boolean' && {suspendUntilReady}),
     ...(typeof suspendWhileReconciling === 'boolean' && {suspendWhileReconciling}),
+    ...(typeof updateOnContextChanged === 'boolean' && {updateOnContextChanged}),
+    ...(typeof updateOnConfigurationChanged === 'boolean' && {updateOnConfigurationChanged}),
```
toddbaert, line 80: This is important because we use the `...` operator to default the options - so we don't want to return any fields withe value `undefined`, or they override `true` fields.

I will be adding specific tests for this today.

lukas-reining (approved):
Looks good, left a nit.

packages/react/src/common/options.ts
toddbaert: absolutely. I hate when I see this. Thanks.

toddbaert: I'm also seeing the var name is now misleading, I'll fix that too.

toddbaert: Actually I can just default this at the param level!

Issue #748, opened 2024-01-09
[FEATURE] Support setting global evaluation context when registering a provider


beeme1mr:
### Requirements

This is essentially semantic sugar on top of existing API. However, it's useful because it helps users avoid accidentally setting global context after initializing the provider. On the client, this may lead to a second async call that could have been avoided.

PR #919, opened 2024-04-25
feat: set context during provider init on web

beeme1mr:
## This PR

- overloads the set provider methods to support defining context in the web SDK
- updates the web sdk readme

## Related Issues

Fixes https://github.com/open-feature/js-sdk/issues/748

## Notes

I decided to only support setting context in the web SDK because it is less valuable on the server and the expected behavior was less clear due to `domains`.

The behavior may need to be spec'd out. An issue in the spec repo has been created.
https://github.com/open-feature/spec/issues/219


packages/client/README.md
```diff
@@ -230,6 +237,24 @@ const domainScopedClient = OpenFeature.getClient("my-domain");
 Domains can be defined on a provider during registration.
 For more details, please refer to the [providers](#providers) section.
 
+#### Manage evaluation context for domains
+
+By default, domain-scoped clients use the global context.
+This can be overridden by explicitly setting context when registering the provider or by references the domain when updating context.
+
+```ts
+OpenFeature.setProvider("my-domain", new NewCachedProvider(), { isCache: true });
```
lukas-reining, line 246: Could we also set the targeting key here? I think this makes the example more obvious.
`isCache: true` can be confused for options or so.

lukas-reining (approved):
Looks good to me!

packages/client/README.md
toddbaert: Ya, I agree. I think something more personal like "`email`" might be clearer.

packages/client/README.md
```diff
@@ -230,6 +237,24 @@ const domainScopedClient = OpenFeature.getClient("my-domain");
...
+```
+
+Change context after the provider has been registered by using `setContext` with a name.
```
toddbaert, line 249:
```suggestion
To change context after the provider has been registered, use `setContext` with a name:
```

packages/client/README.md
```diff
@@ -230,6 +237,24 @@ const domainScopedClient = OpenFeature.getClient("my-domain");
 Domains can be defined on a provider during registration.
 For more details, please refer to the [providers](#providers) section.
 
+#### Manage evaluation context for domains
+
+By default, domain-scoped clients use the global context.
+This can be overridden by explicitly setting context when registering the provider or by references the domain when updating context.
```
toddbaert, line 243:
```suggestion
This can be overridden by explicitly setting context when registering the provider or by references the domain when updating context:
```

packages/client/src/open-feature.ts
```diff
@@ -61,6 +68,142 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<ClientProviderStatus, P
...
+    await this.setAwaitableProvider(domain, provider);
+  }
+
+  /**
+   * Sets the default provider for flag evaluations.
+   * This provider will be used by domainless clients and clients associated with domains to which no provider is bound.
+   * Setting a provider supersedes the current provider used in new and existing unbound clients.
+   * @param {Provider} provider The provider responsible for flag evaluations.
+   * @returns {this} OpenFeature API
+   */
```
toddbaert, line 143: Where applicable, can you move these back to the abstract and use [@inheritdoc](https://jsdoc.app/tags-inheritdoc)?

beeme1mr: I tried but couldn't get it working cleanly. VSCode joined the docs for all the overloaded methods in the IntelliSense popup.

lukas-reining: I had the same thought @toddbaert, but was not sure and tried and saw the same behavior @beeme1mr found.

beeme1mr:
I've addressed your feedback. @toddbaert, any other concerns?

.vscode/settings.json
beeme1mr, line 1: I accidentally included this in the PR, but we already have a `.vscode` folder. Should I remove it and add `settings.json` to the gitignore?

packages/client/src/open-feature.ts
```diff
@@ -61,6 +68,142 @@ export class OpenFeatureAPI extends OpenFeatureCommonAPI<ClientProviderStatus, P
...
+        this._context = context;
+      }
+    }
+
+    const maybePromise = this.setAwaitableProvider(domain, provider);
+
+    // The setProvider method doesn't return a promise so we need to catch and
+    // log any errors that occur during provider initialization to avoid having
+    // an unhandled promise rejection.
+    if (maybePromise instanceof Promise) {
```
moredip, line 199: I'm not sure that this is a foolproof way to check whether a value is a promise. Based on [this](https://stackoverflow.com/questions/27746304/how-to-check-if-an-object-is-a-promise) I would suggest something like `Promise.resolve(maybePromise).catch(...)` instead.

beeme1mr: I haven't seen Bluebird Promises in the Wild lately, but your proposal is definitely better than what I had; thanks!

packages/client/src/open-feature.ts
toddbaert: bummer :(

beeme1mr:
@toddbaert @moredip any other concerns?

toddbaert (approved):

packages/client/src/open-feature.ts
lukas-reining: Yes totally @toddbaert.

packages/server/README.md
```diff
@@ -234,6 +240,26 @@ const domainScopedClient = OpenFeature.getClient("my-domain");
 Domains can be defined on a provider during registration.
 For more details, please refer to the [providers](#providers) section.
 
+#### Manage evaluation context for domains
+
+TODO: See how we want to handle this behavior.
```
beeme1mr, line 245: Address TODO

PR #921, opened 2024-04-26
fix: delayed suspense causes "flicker"

### LINKED ISSUE #920: [BUG] React SDK suspends after useEffect rather than during a component's first render, causing components to flicker
### Observed behavior

The current implementation of Suspense in the React SDK seems kind of buggy.

A component is meant to suspend **during** rendering, but the React SDK makes the component suspend **after** the inital render, after running useEffect. 

This not only breaks React APIs like `startTransition`, but it leads to a noticeable "blinking" effect.

open-feature-suspense-bug

So, the current behavior seems to be:

1. Component renders once, very quickly, with invalid values, while the Client is `NOT_READY`.
2. useEffect runs, sets the state to a function that throws a promise.
2. Component rerenders and suspend (because it suspends in the second render, it's too late for startTransition to work now ❌)
3. Component renders again, once the Client is `READY`.

### Expected Behavior

A component should suspend immediatly during its first render, not in a second render after the effects have fired.

### Steps to reproduce

**CodeSandbox:** https://codesandbox.io/embed/openfeature-suspense-bug-5j7yll?fontsize=14&hidenavigation=1&theme=dark

The test in [/packages/react/test/provider.spec.tsx](https://github.com/open-feature/js-sdk/blob/86a9230addb8b7a0bb34ca7ef6d69b27209128f2/packages/react/test/provider.spec.tsx#L96C9-L98C66) show a false positive, because React Testing Library only shows the state of the DOM **after** the effects have run. 

To make sure that the `'👎'` truly never appears, we can add a mutable variable that tracks wether or not the component suspend properly. If the component suspends properly, `isReady` will never be `false`, because `useWhenProviderReady` should suspend before the function completes.

Here is a modified version of the `'should suspend until ready and then return provider status'` test that catches the component not suspending properly:

```js
it('should suspend until ready and then return provider status', async () => {
  OpenFeature.setProvider(SUSPENSE_ON, suspendingProvider());

  let renderedWhileNotReady = false;

  function Component() {
    const isReady = useWhenProviderReady();

    if (!isReady) {
      renderedWhileNotReady = true;
    }

    return (
      <>
        <div>{isReady ? '👍' : '👎'}</div>
      </>
    );
  }

  render(
    <OpenFeatureProvider domain={SUSPENSE_ON}>
      <React.Suspense fallback={<div>{FALLBACK}</div>}>
        <Component></Component>
      </React.Suspense>
    </OpenFeatureProvider>,
  );


  // should see fallback initially
  expect(renderedWhileNotReady).toBe(false);
  expect(screen.queryByText('👎')).not.toBeInTheDocument();
  expect(screen.queryByText(FALLBACK)).toBeInTheDocument();
  // eventually we should the value
  await waitFor(() => expect(screen.queryByText('👍')).toBeVisible(), { timeout: DELAY * 2 });
});
```

tjosepo:
## This PR

- Add an internal `useOpenFeatureClientStatus()` function to make it slightly easier to keep track of the state of the current Client, and know when to suspend or not.

- Replaces the `suspend()` function with a `suspendUntilReady()` function. The `suspendUntilReady()`  function fires immediatly during the first render or when the component cannot render due to the client having changed status.

- Put the Client status and the feature flag details inside React states. This is required to make APIs like `startTransition` work properly with the library

### Related Issues

Fixes #920

### Follow-up Tasks

- [x] Add more tests. Suspense and Transitions are very tricky. While these changes don't seem to have broken any existing tests, we should add new tests to make sure Suspense and Transitions continue to work properly in the future.

- [ ] ~I think there might still be some reactivity issues with some parts of the codebase. The client returned by `useOpenFeatureClient()` didn't seem to update properly when the client changed, which made `useOpenFeatureClientStatus()` not update properly.~ Out of the scope of this PR

### How to test

I built the application with `npm build`, then `npm pack`. I used the modified code on a local project to see if it fixed the issues.

I built projects in CodeSanbox that demonstrate the issues.

For Suspense: https://codesandbox.io/embed/openfeature-suspense-bug-5j7yll?fontsize=14&hidenavigation=1&theme=dark

For Transitions: https://codesandbox.io/embed/openfeature-suspense-bug-forked-lqhyf3?fontsize=14&hidenavigation=1&theme=dark

### DIFF
--- a/packages/react/src/common/suspend.ts
+++ b/packages/react/src/common/suspend.ts
@@ -1,75 +1,20 @@
 import { Client, ProviderEvents } from '@openfeature/web-sdk';
-import { Dispatch, SetStateAction } from 'react';
-
-enum SuspendState {
-  Pending,
-  Success,
-  Error,
-}
 
 /**
- * Suspend function. If this runs, components using the calling hook will be suspended.
- * DO NOT EXPORT PUBLICLY 
- * @internal
- * @param {Client} client the OpenFeature client
- * @param {Function} updateState the state update function
- * @param {ProviderEvents[]} resumeEvents list of events which will resume the suspend
- */
-export function suspend(
-  client: Client,
-  updateState: Dispatch<SetStateAction<object | undefined>>,
-  ...resumeEvents: ProviderEvents[]
-) {
-  let suspendResolver: () => void;
-
-  const suspendPromise = new Promise<void>((resolve) => {
-    suspendResolver = () => {
-      resolve();
-      resumeEvents.forEach((e) => {
-        client.removeHandler(e, suspendResolver); // remove handlers once they've run
-      });
-      client.removeHandler(ProviderEvents.Error, suspendResolver);
-    };
-    resumeEvents.forEach((e) => {
-      client.addHandler(e, suspendResolver);
-    });
-    client.addHandler(ProviderEvents.Error, suspendResolver); // we never want to throw, resolve with errors - we may make this configurable later
-  });
-  updateState(suspenseWrapper(suspendPromise));
-}
-
-/**
- * Promise wrapper that throws unresolved promises to support React suspense.
+ * Suspends until the client is ready to evaluate feature flags.
  * DO NOT EXPORT PUBLICLY
- * @internal
- * @param {Promise<T>} promise to wrap
- * @template T flag type
- * @returns {Function} suspense-compliant lambda
+ * @param client
  */
-export function suspenseWrapper<T>(promise: Promise<T>) {
-  let status: SuspendState = SuspendState.Pending;
-  let result: T;
-
-  const suspended = promise
-    .then((value) => {
-      status = SuspendState.Success;
-      result = value;
-    })
-    .catch((error) => {
-      status = SuspendState.Error;
-      result = error;
-    });
-
-  return () => {
-    switch (status) {
-      case SuspendState.Pending:
-        throw suspended;
-      case SuspendState.Success:
-        return result;
-      case SuspendState.Error:
-        throw result;
-      default:
-        throw new Error('Suspending promise is in an unknown state.');
-    }
-  };
+export function suspendUntilReady(client: Client): Promise<void> {
+  let resolve: (value: unknown) => void;
+  let reject: () => void;
+  throw new Promise((_resolve, _reject) => {
+    resolve = _resolve;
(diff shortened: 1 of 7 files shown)


toddbaert:
This looks great!

Don't try to do everything in one go though, I would recommend you just deal with the suspense issues for now, if you want to fix any reactivity issues with the other hooks, try to do that separately.

toddbaert:
I also locally built and linked this, and tested it with our[ little demo app](https://github.com/open-feature/react-test-app), and it worked exactly as expected! Please carry on!

Be careful not to leak any new handlers!

Thanks again :pray:

tjosepo:
Perfect! I'll try adding a few more tests. I'll try to get the PR ready to review by the end of the day.

toddbaert:
> Perfect! I'll try adding a few more tests. I'll try to get the PR ready to review by the end of the day.

No rush!

tjosepo:
I modified the Suspense tests to make sure the component suspended immediately when they should.

I also added two tests to make sure transitions work properly (transitions are tricky to get right, so I'm happy tests now cover it!)

toddbaert:
@tjosepo Thanks again! I will review this no later than tomorrow.

toddbaert (approved):
@tjosepo I tried but I can't find anything to complain about :grin:. I also tested it again in our [test app](https://github.com/open-feature/react-test-app). Thanks for identifying this issue and for the fix! :pray: 

Would you like an invite to the OpenFeature org? It comes with no obligation, but allows us to ping you and keep you in the loop on Github. It's the first step on the [contributor ladder](https://github.com/open-feature/community/blob/main/CONTRIBUTOR_LADDER.md).

beeme1mr (approved):

tjosepo:
> Would you like an invite to the OpenFeature org? It comes with no obligation, but allows us to ping you and keep you in the loop on Github. It's the first step on the [contributor ladder](https://github.com/open-feature/community/blob/main/CONTRIBUTOR_LADDER.md).

Sure! We are very likely to be using OpenFeature at my job soon, so I'd love to contribute to it when I can!

lukas-reining (approved):
Looks really good @tjosepo, thanks! :)

toddbaert:
@tjosepo created a PR to add you to the org: https://github.com/open-feature/community/pull/372

PR #922, opened 2024-04-30
test: add testing for HookFlagQuery

### LINKED ISSUE #915: [TESTING] increase React SDK test coverage
We're missing some coverage of the `HookFlagQuery` class (an internal class that supports our "React query style" API). Add coverage to the uncovered lines by extending the React SDK test suite. This probably can be done most easily by extracting this class into its own file, so it can be imported into tests and tested independently by creating some instances from test data and making some basic assertions. If extracted, be careful not to export this class from any `/index` files, **we don't want it to be public**.

See image below for missing coverage:

image

Link to source: https://github.com/open-feature/js-sdk/blob/e4287ed1a59a24c15efc071afc9148a034d27c4a/packages/react/src/evaluation/use-feature-flag.ts#L303-L348

See the interface this class implements for more details.

ro1hep:
## This PR

- introduces tests for `HookFlagQuery`

### Related Issues

Resolves #915 

### Notes
Placed these tests within `evaluation.spec.tsx`, happy to move them into their own file if more appropriate


toddbaert:
Thanks @ro1hep ! We'll review.

packages/react/test/evaluation.spec.tsx
```diff
@@ -577,5 +580,76 @@ describe('evaluation', () => {
...
+          flagKey: 'flag-key',
+          flagMetadata : {},
+          value: 'string'
+        };
+
+        const hookFlagQuery = new HookFlagQuery(details);
+        expect(hookFlagQuery.details).toEqual(details);
+      });
+
+      it('should return is authorative correctly', () => {
```
james-milligan, line 630: could potentially refactor this test and the above to use an `it.each()` but it might be overkill 🤷

toddbaert: wow. I've been using Jest for years and didn't know about `.each`, Go tabular testing in JS :sunglasses:

toddbaert:
> 100% Statements 56/56100% Branches 15/15100% Functions 12/12100% Lines 56/56

Nice.

packages/react/test/evaluation.spec.tsx
```diff
@@ -577,5 +580,76 @@ describe('evaluation', () => {
...
+        const flagMetadata = {
+          'ping': 'pong'
+        };
+        const details: EvaluationDetails<boolean> = {
+          flagKey: 'with-metadata',
+          flagMetadata,
+          value: true,
+        };
+        const hookFlagQuery = new HookFlagQuery(details);
+        expect(hookFlagQuery.flagMetadata).toEqual(flagMetadata);
```
toddbaert, line 616: `expect.ObjectContaining` is a less-brittle way of testing this sort of thing, if a new prop is added it won't fail... on the other hand sometimes that's good because it forces you to update the test.

packages/react/test/evaluation.spec.tsx
toddbaert, line 630:
```suggestion
      it('should return is authorative if Reason != STALE/DISABLED and errorCode unset', () => {
```

I try to avoid vagueness like this in test descriptions, it's helpful to future readers if the `it` properly describes the rule.

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -6,14 +6,15 @@ import {
...
-  StandardResolutionReasons
 } from '@openfeature/web-sdk';
 import { useEffect, useState } from 'react';
 import { DEFAULT_OPTIONS, ReactFlagEvaluationOptions, normalizeOptions } from '../common/options';
 import { suspend } from '../common/suspend';
 import { useProviderOptions } from '../provider/context';
 import { useOpenFeatureClient } from '../provider/use-open-feature-client';
 import { FlagQuery } from '../query';
+import { HookFlagQuery } from './hook-flag-query';
+
```
toddbaert, line 17:
```suggestion
```

toddbaert (approved):

toddbaert:
@ro1hep Thanks a lot for your contribution. I left a comment and a few tiny suggestions, but approving as is.

packages/react/test/evaluation.spec.tsx
ro1hep: Great suggestion I didn't think of that - I'll amend now!

ro1hep:
```
react/src/evaluation                                   |     100 |      100 |     100 |     100 |                                                             
  hook-flag-query.ts                                    |     100 |      100 |     100 |     100 |    
```

Made the changes requested, used tabular testing for the error tests and updated the name

packages/react/test/evaluation.spec.tsx
ro1hep: Added

packages/react/test/evaluation.spec.tsx
ro1hep: Introduced tabular testing for this test case and the error test case

james-milligan: LGTM 🚀

james-milligan (approved):
@ro1hep nice work, LGTM! 🚀

toddbaert:
Fixed a minor conflict with a recent change. Merging.

PR #926, opened 2024-05-01
fix: run error hook when provider returns reason error or error code

### LINKED ISSUE #925: [BUG] Error hooks only run if the provider throws
### Observed behavior

The error hook only runs if the provider throws an exception. However, it's possible for a provider to return an error code in the `ResolutionDetails`.

### Expected Behavior

The error hook runs if the `ResolutionDetails` contains an `errorCode`

### Steps to reproduce

1. Register a provider that returns a `ResolutionDetails` object with an `errorCode`
2. Register a hook that relies on the error hook
3. Evaluate a flag

beeme1mr:
## This PR

- runs error hook when provider returns reason error or error code

### Related Issues

Fixes #925

### Notes

Based on a conversation in Slack: https://cloud-native.slack.com/archives/C06E4DE6S07/p1714581197391509

### DIFF
--- a/packages/client/src/client/open-feature-client.ts
+++ b/packages/client/src/client/open-feature-client.ts
@@ -226,12 +226,15 @@ export class OpenFeatureClient implements Client {
         flagKey,
       };
 
-      if (evaluationDetails.errorCode || evaluationDetails.reason === StandardResolutionReasons.ERROR) {
-        const errorCode = evaluationDetails.errorCode || ErrorCode.GENERAL;
-        this.errorHooks(allHooksReversed, hookContext, instantiateErrorByErrorCode(errorCode), options);
+      if (evaluationDetails.errorCode) {
+        this.errorHooks(
+          allHooksReversed,
+          hookContext,
+          instantiateErrorByErrorCode(evaluationDetails.errorCode),
+          options,
+        );
         return {
           ...evaluationDetails,
-          errorCode,
           reason: StandardResolutionReasons.ERROR,
         };
       }
--- a/packages/client/src/client/open-feature-client.ts
+++ b/packages/client/src/client/open-feature-client.ts
@@ -227,16 +227,7 @@ export class OpenFeatureClient implements Client {
       };
 
       if (evaluationDetails.errorCode) {
-        this.errorHooks(
-          allHooksReversed,
-          hookContext,
-          instantiateErrorByErrorCode(evaluationDetails.errorCode),
-          options,
-        );
-        return {
-          ...evaluationDetails,
-          reason: StandardResolutionReasons.ERROR,
-        };
+        throw instantiateErrorByErrorCode(evaluationDetails.errorCode);
       }
 
       this.afterHooks(allHooksReversed, hookContext, evaluationDetails, options);
--- a/packages/client/src/client/open-feature-client.ts
+++ b/packages/client/src/client/open-feature-client.ts
@@ -15,7 +15,7 @@ import {
   ResolutionDetails,
   SafeLogger,
   StandardResolutionReasons,
-  instantiateErrorByKey,
+  instantiateErrorByErrorCode,
   statusMatchesEvent,
 } from '@openfeature/core';
 import { FlagEvaluationOptions } from '../evaluation';
@@ -209,7 +209,7 @@ export class OpenFeatureClient implements Client {
 
     try {
       this.beforeHooks(allHooks, hookContext, options);
-      
+
       // short circuit evaluation entirely if provider is in a bad state
       if (this.providerStatus === ProviderStatus.NOT_READY) {
         throw new ProviderNotReadyError('provider has not yet initialized');
@@ -228,7 +228,7 @@ export class OpenFeatureClient implements Client {
 
       if (evaluationDetails.errorCode || evaluationDetails.reason === StandardResolutionReasons.ERROR) {
         const errorCode = evaluationDetails.errorCode || ErrorCode.GENERAL;
-        this.errorHooks(allHooksReversed, hookContext, instantiateErrorByKey(errorCode), options);
+        this.errorHooks(allHooksReversed, hookContext, instantiateErrorByErrorCode(errorCode), options);
         return {
           ...evaluationDetails,
           errorCode,
--- a/packages/client/src/client/open-feature-client.ts
+++ b/packages/client/src/client/open-feature-client.ts
@@ -15,7 +15,8 @@ import {
   ResolutionDetails,
   SafeLogger,
   StandardResolutionReasons,
-  statusMatchesEvent
+  instantiateErrorByKey,
+  statusMatchesEvent,
 } from '@openfeature/core';
 import { FlagEvaluationOptions } from '../evaluation';
 import { ProviderEvents } from '../events';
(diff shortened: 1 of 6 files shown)


packages/client/src/client/open-feature-client.ts
```diff
@@ -225,6 +226,16 @@ export class OpenFeatureClient implements Client {
...
       };
 
+      if (evaluationDetails.errorCode || evaluationDetails.reason === StandardResolutionReasons.ERROR) {
+        const errorCode = evaluationDetails.errorCode || ErrorCode.GENERAL;
+        this.errorHooks(allHooksReversed, hookContext, instantiateErrorByKey(errorCode), options);
+        return {
+          ...evaluationDetails,
+          errorCode,
+          reason: StandardResolutionReasons.ERROR,
+        };
```
beeme1mr, line 236: I'm also setting the reason to error. This will override the reason set by the provider. Does this make sense or should we return the reason defined in the evaluation details?

packages/client/test/hooks.spec.ts
```diff
@@ -206,6 +208,48 @@ describe('Hooks', () => {
...
+        const mockErrorHook = jest.fn();
+
+        const details = client.getBooleanDetails(FLAG_KEY, false, {
+          hooks: [{ error: mockErrorHook }],
+        });
+
+        expect(mockErrorHook).toHaveBeenCalled();
+        expect(details).toEqual(
+          expect.objectContaining({
+            errorCode: ErrorCode.GENERAL,
```
beeme1mr, line 248: Error code is automatically set to `GENERAL` if the provider didn't explicitly define a code.

packages/client/src/client/open-feature-client.ts
lukas-reining: I think error is good here!

lukas-reining (approved):
Nice, I like the change!

packages/shared/src/errors/index.ts
```diff
@@ -1,9 +1,45 @@
...
+import { InvalidContextError } from './invalid-context-error';
+import { OpenFeatureError } from './open-feature-error-abstract';
+import { ParseError } from './parse-error';
+import { ProviderFatalError } from './provider-fatal-error';
+import { ProviderNotReadyError } from './provider-not-ready-error';
+import { TargetingKeyMissingError } from './targeting-key-missing-error';
+import { TypeMismatchError } from './type-mismatch-error';
+
+const instantiateErrorByKey = (key: ErrorCode, message?: string): OpenFeatureError => {
+  switch (key) {
```
beeme1mr, line 14:
```suggestion
const instantiateErrorByErrorCode = (errorCode: ErrorCode, message?: string): OpenFeatureError => {
  switch (errorCode) {
```

packages/server/src/client/open-feature-client.ts
```diff
@@ -278,6 +279,16 @@ export class OpenFeatureClient implements Client, ManageContext<OpenFeatureClien
         flagKey,
       };
 
+      if (evaluationDetails.errorCode || evaluationDetails.reason === StandardResolutionReasons.ERROR) {
+        const errorCode = evaluationDetails.errorCode || ErrorCode.GENERAL;
+        await this.errorHooks(allHooksReversed, hookContext, instantiateErrorByErrorCode(errorCode), options);
+        return {
+          ...evaluationDetails,
```
toddbaert, line 286: Should we add our own `errorMessage` if one wasn't supplied by the provider?

I suspect in most cases they would populate one, and if we added one it might not be very informative... we could add something like `"provider returned error code"`... Leaning towards no, but wanted to mention it.

packages/server/src/client/open-feature-client.ts
```diff
@@ -278,6 +279,16 @@ export class OpenFeatureClient implements Client, ManageContext<OpenFeatureClien
         flagKey,
       };
 
+      if (evaluationDetails.errorCode || evaluationDetails.reason === StandardResolutionReasons.ERROR) {
```
toddbaert, line 282: This is the only thing that gives me pause.

I wonder if it might be better to predicate this failure _only_ on the `errorCode` being present. Having a very concise rule is arguably better than one with an "or".

ie: `if the provider sets an error code, error hooks must run and the default must be returned` is more cut and dry than `if the provider sets an error code or set's the reason to "ERROR", error hooks must run and the default must be returned`.

maxveldink: Yeah, I agree that the few places we check for errors in the Ruby code, it's on the presence of _anything_ in `error_code`. `reason` seems like a much more ambiguous field that some providers might not fill out on error, so I think ignoring it for error flows makes sense to me.

packages/server/src/client/open-feature-client.ts
beeme1mr: I don't think that provides any additional information that would be useful. If the provider sets an error message, it will be retained, but setting a default message would basically rephrase the same information that's already available.

packages/server/src/client/open-feature-client.ts
beeme1mr: Okay, thanks for the feedback. I've updated the PR with this change.

packages/client/src/client/open-feature-client.ts
```diff
@@ -225,6 +226,19 @@ export class OpenFeatureClient implements Client {
         flagKey,
       };
 
+      if (evaluationDetails.errorCode) {
+        this.errorHooks(
```
federicobond, line 230: In the Python SDK we construct and raise an appropriate exception here to avoid duplicating the error handling code from the catch clause. Would it make sense to do something similar?

toddbaert (approved):

packages/client/src/client/open-feature-client.ts
toddbaert: I did a little investigation into this and it looks like in node, stack traces are captured in the error constructor. So calling `new Error()` captures the stack trace and results in significant performance penalty ([reference](https://nodejs.org/api/errors.html#class-error)). This is news to me, I actually thought it happened with the throw (something c# does).

Provider authors returning an error the way we do in this PR has some benefits, in that if it's a error condition that's well understood and doesn't benefit from a stack trace and a proper exception, they could return an error and avoid the performance hit. If we instantiate our own exception anyway, we've negated that performance improvement.

We _could_ instantiate and throw a non-error, but that could be confusing and lead to unexpected bugs.

federicobond: But that's what `instantiateErrorByErrorCode` does a couple of lines below, or am I missing something? If I understand you correctly, the performance cost of instantiating this error would be paid in both cases.

toddbaert: Ya, I was being very unclear.

This PR _does_ carry the performance penalty I mentioned already. The way to avoid that would be to construct something shaped like an error that doesn't capture the stack trace, that's what I meant by:

> We could instantiate and throw a non-error, but that could be confusing and lead to unexpected bugs.

Irrespective of that, the flow could be improved by doing what you suggest I think; sorry, I got carried away by my own research how the stack trace is built and got away from your original point.

beeme1mr: I've made this change. Thanks for the suggestion.

toddbaert (approved):

Kavindu-Dodan:
@toddbaert @beeme1mr 

In Go, there's no concept of throwing, so we rely on error details [^1] already

I created a follow-up for Java SDK [^2]

[^1]: - https://github.com/open-feature/go-sdk/blob/v1.11.0/openfeature/client.go#L747-L748
[^2]: - https://github.com/open-feature/java-sdk/issues/928

Issue #933, opened 2024-05-06
[React] Disable suspense by default, add `useSuspenseFlag` hooks


toddbaert:
Currently, the React SDK suspends by default. Suspense is a REALLY cool feature, and very useful for our library in particular. However, this means that users see errors if they don't set up a suspense boundary. I'm proposing we make 2 changes:

- disable suspense by default
- add dedicated "useSuspenseXxx" versions of all relevant hooks
  - the existing hooks could still be used with suspense, but you would have to pass `{ suspense: true}` in the options
- mark all suspense features as experimental in jsdocs
  - from react 18 docs: `Suspense-enabled data fetching without the use of an opinionated framework is not yet supported. The requirements for implementing a Suspense-enabled data source are unstable and undocumented. An official API for integrating data sources with Suspense will be released in a future version of React` (https://react.dev/reference/react/Suspense#usage) 

The above proposals are consistent with popular libraries that use suspense on data fetching:

TanStack:
 - has [useQuery](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery) and [useSuspenseQuery](https://tanstack.com/query/latest/docs/framework/react/reference/useSuspenseQuery); `useSuspenseQuery` is just [useQuery-v4](https://tanstack.com/query/v4/docs/framework/react/reference/useQuery) with suspense enabled

Apollo:
- has [useSuspenseQuery](https://www.apollographql.com/docs/react/data/suspense/#fetching-with-suspense) which is a "a Suspense-ready replacement for `useQuery` that lets you take advantage of React's Suspense features while fetching during render"

There is only a handful of libraries that actually support suspense on data fetching, since it's an experimental feature.

:warning: This would be a behaviorally breaking change (things would compile, but suspense options would need to be inverted to maintain the same behavior).

toddbaert:
cc @beeme1mr @lukas-reining @tjosepo @thisthat @Kavindu-Dodan @luizgribeiro @thomaspoignant

tjosepo:
I support the proposed changes.

As much as I like Suspense, I can't deny that it's a very new concept and most of the React community isn't fully accustomed to it yet. Requiring users of the library to use a Suspense boundary by default can be seen as a small frustration.

I like the way TanStack and Appollo separate their hooks in suspense and non-suspense version.

moredip:
[...]

PR #940, opened 2024-05-07
feat!: disable suspense by default, add suspense hooks

toddbaert:
This PR:

- inverts the suspense defaults (we now do not suspend by default, you have to add `suspend:true` in options)
- adds `useSuspenseFlag` (analogous to `useSuspenseXxx`) in other libraries, which behaves the same as `useFlag` with `{ suspend: true }`
- updates README (specifically encourages use use "query-style" hooks over type-specific hooks
- adds `@experimental` jsdoc marker to all suspense options and hooks
- associated tests

Things to consider:

- I did not add `useSuspense{Type}FlagValue` and `useSuspense{Type}FlagDetails` hooks; we could do this if we wanted, but IMO these are already not the primary APIs we want to push users toward in react - we want them to use the generic `useFlag` and `useSuspenseFlag` which return the react query interfaces.

Fixes: https://github.com/open-feature/js-sdk/issues/933

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -199,10 +199,14 @@ You can disable this feature in the hook options (or in the [OpenFeatureProvider
 
 ```tsx
 function Page() {
-  const showNewMessage = useBooleanFlagValue('new-message', false, { updateOnContextChanged: false });
+  const { value: showNewMessage } = useFlag('new-message', false, { updateOnContextChanged: false });
   return (
-    <MyComponents></MyComponents>
-  )
+    <div className="App">
+      <header className="App-header">
+        {showNewMessage ? <p>Welcome to this OpenFeature-enabled React app!</p> : <p>Welcome to this React app.</p>}
+      </header>
+    </div>
+  );
 }
 ```
 
@@ -216,10 +220,14 @@ You can disable this feature in the hook options (or in the [OpenFeatureProvider
 
 ```tsx
 function Page() {
-  const showNewMessage = useBooleanFlagValue('new-message', false, { updateOnConfigurationChanged: false });
+  const { value: showNewMessage } = useFlag('new-message', false, { updateOnConfigurationChanged: false });
   return (
-    <MyComponents></MyComponents>
-  )
+    <div className="App">
+      <header className="App-header">
+        {showNewMessage ? <p>Welcome to this OpenFeature-enabled React app!</p> : <p>Welcome to this React app.</p>}
+      </header>
+    </div>
+  );
 }
 ```
 
@@ -227,9 +235,14 @@ Note that if your provider doesn't support updates, this configuration has no im
 
 #### Suspense Support
 
+> [!NOTE]  
+> React suspense is an experimental feature and subject to change in future versions.
+
+
 Frequently, providers need to perform some initial startup tasks.
 It may be desireable not to display components with feature flags until this is complete, or when the context changes.
 Built-in [suspense](https://react.dev/reference/react/Suspense) support makes this easy.
+Use `useSuspenseFlag` or pass `{ suspend: true }` in the hook options to leverage this functionality.
 
 ```tsx
 function Content() {
@@ -242,8 +255,8 @@ function Content() {
 }
 
 function Message() {
-  // component to render after READY.
-  const showNewMessage = useBooleanFlagValue('new-message', false);
+  // component to render after READY, equivalent to useFlag('new-message', false, { suspend: true });
+  const { value: showNewMessage } = useSuspenseFlag('new-message', false);
 
   return (
     <>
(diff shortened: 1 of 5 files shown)


packages/react/README.md
```diff
@@ -216,20 +220,29 @@ You can disable this feature in the hook options (or in the [OpenFeatureProvider
 
 ```tsx
 function Page() {
-  const showNewMessage = useBooleanFlagValue('new-message', false, { updateOnConfigurationChanged: false });
+  const { value: showNewMessage } = useFlag('new-message', false, { updateOnConfigurationChanged: false });
```
toddbaert, line 223: updated the README to push people toward the query-style API.

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -31,6 +31,9 @@ type ConstrainedFlagQuery<T> = FlagQuery<
           : JsonValue
 >;
 
+// suspense options removed for the useSuspenseFlag hooks
```
toddbaert, line 34: Remove suspense from the options of the `useSuspenseFlag` hook (it's always on).

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -70,6 +73,26 @@ T extends boolean
   return query as unknown as ConstrainedFlagQuery<T>;
 }
 
+// alias to the return value of useFlag, used to keep useSuspenseFlag consistent
+type ReturnTypeOfUseFlag<T extends FlagValue> = ReturnType<typeof useFlag<T>>
```
toddbaert, line 77: Aliased this type so `useFlag` and `useSuspense` flag stay synced up.

packages/react/test/evaluation.spec.tsx
```diff
@@ -254,14 +255,13 @@ describe('evaluation', () => {
      */
     const DELAY = 100;
 
-    const SUSPENSE_ON = 'suspense';
```
toddbaert, line 257: I improved the isolation of the test suites a bit.

packages/react/test/evaluation.spec.tsx
```diff
@@ -454,48 +470,119 @@ describe('evaluation', () => {
...
+        await waitFor(() => expect(screen.queryByText(TARGETED_FLAG_VALUE)).toBeInTheDocument(), {
+          timeout: DELAY * 2,
+        });
+      });
     });
 
-    it('should show the previous UI while reconciling during a transition', async () => {
-      await OpenFeature.setContext(SUSPENSE_OFF, {});
-      OpenFeature.setProvider(SUSPENSE_ON, suspendingProvider());
+    describe.each([useFlag, useSuspenseFlag])('suspend=true', () => {
```
toddbaert, line 512: Note this runs multiple times via `.each`.

packages/react/src/evaluation/use-feature-flag.ts
beeme1mr: Could the `suspendWhileReconciling` still be useful to control?

toddbaert: I think it's best to make the suspense hooks enable all suspense options by default. I think if more fine grained control is needed I'd use the 'useFlag' hook. I think in most cases , you either want all suspense or none.

packages/react/README.md
```diff
@@ -199,10 +199,14 @@ You can disable this feature in the hook options (or in the [OpenFeatureProvider
...
-    <MyComponents></MyComponents>
-  )
+    <div className="App">
+      <header className="App-header">
+        {showNewMessage ? <p>Welcome to this OpenFeature-enabled React app!</p> : <p>Welcome to this React app.</p>}
+      </header>
+    </div>
+  );
 }
 ```
```
beeme1mr, line 212: You could link the static context paradigm concepts page now if you want.
https://openfeature.dev/docs/reference/concepts/sdk-paradigms

beeme1mr (approved):

packages/react/src/evaluation/use-feature-flag.ts
lukas-reining: Just a naming thing but I would expect a type name like `UseFlagReturn` or even `UseFlagResult`.

lukas-reining (approved):
Change looks good to me, left a thought on return type naming.

packages/react/src/evaluation/use-feature-flag.ts
toddbaert, line 77:
```suggestion
type UseFlagReturn<T extends FlagValue> = ReturnType<typeof useFlag<T>>
```

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -70,6 +73,26 @@ T extends boolean
...
+
+/**
+ * Equivalent to {@link useFlag} with `options: { suspend: true }`
+ * @experimental Suspense is an experimental feature subject to change in future versions.
+ * @param {string} flagKey the flag identifier
+ * @template {FlagValue} T A optional generic argument constraining the default.
+ * @param {T} defaultValue the default value; used to determine what resolved type should be used.
+ * @param {NoSuspenseOptions} options for this evaluation
+ * @returns { ReturnTypeOfUseFlag<T> } a queryable object containing useful information about the flag.
```
toddbaert, line 86:
```suggestion
 * @returns { UseFlagReturn<T> } a queryable object containing useful information about the flag.
```

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -70,6 +73,26 @@ T extends boolean
...
+ */
+export function useSuspenseFlag<T extends FlagValue = FlagValue>(
+  flagKey: string,
+  defaultValue: T,
+  options?: NoSuspenseOptions,
+): ReturnTypeOfUseFlag<T> {
```
toddbaert, line 92:
```suggestion
): UseFlagReturn<T> {
```

toddbaert:
@tjosepo Just FYI b/c you did some recent related work. I plan on merging/releasing this EOD.

PR #957, opened 2024-06-10
feat: lower compilation target to es2015

### LINKED ISSUE #956: [FEATURE] Consider Alternative Implementations for cause in ErrorOptions to Maintain Lower Compilation Target Compatibility
### Requirements

A while back, PR #765 was merged, introducing support for the `cause` property in `ErrorOptions`. This was implemented by changing the compilation target from `es2015` to `es2022`. However, this change may cause compilation errors for SDK users who are compiling to a lower target and have `skipLibChecks:false`.

While the `cause` functionality is a valuable addition, I'm concerned about the potential impact on runtime support. Is the added functionality worth the potential decrease in compatibility?

I'm wondering if there might be alternative ways to implement this feature without changing the compilation target. For instance, could we manually set the `cause` property where needed? This approach might be compatible with both older and newer runtimes, and could potentially avoid the issue altogether.

I'm interested in hearing your thoughts, and open to contributing a PR.

beeme1mr:
## This PR

- replace the es2022 error cause with a custom implementation
- lower compilation target from es2022 to es2015

### Related Issues

Fixes #956

### Notes

The tests pass, but I still want to manually build and test the outputs in a real application to ensure everything works as expected.


weyert (approved):
LGTM and I think its a reasonable change to have

toddbaert (approved):
LGTM.

nicklasl (approved):

lukas-reining (approved):

PR #971, opened 2024-07-26
feat: add test provider

toddbaert:
Adds some testing utilities, specifically an `<OpenFeatureTestProvider/>` react context provider.

See README for details.


toddbaert:
I recommend we merge https://github.com/open-feature/js-sdk/pull/972 first.

aepfli (approved):
lgtm...

... but as i am not so familiar with the react sdk, do we have a concept of domains also here? and if we do, how could i set it up with the testing provider? More out of curiosity, rather than we should tackle this wihtin this pr. i do think that this adds a lot of benefits to the users of the lib

PR #972, opened 2024-07-26
fix: null derefs on missing metadata.name

toddbaert:
We can be more defensive around missing provider metadata (typings make them required, but there's always JS).

### DIFF
--- a/packages/client/src/client/internal/open-feature-client.ts
+++ b/packages/client/src/client/internal/open-feature-client.ts
@@ -78,7 +78,7 @@ export class OpenFeatureClient implements Client {
         handler({
           clientName: this.metadata.name,
           domain: this.metadata.domain,
-          providerName: this._provider.metadata.name,
+          providerName: this._provider.metadata?.name,
         });
       } catch (err) {
         this._logger?.error('Error running event handler:', err);
--- a/packages/server/src/client/internal/open-feature-client.ts
+++ b/packages/server/src/client/internal/open-feature-client.ts
@@ -80,9 +80,9 @@ export class OpenFeatureClient implements Client {
       // run immediately, we're in the matching state
       try {
         handler({
-          clientName: this.metadata.name,
-          domain: this.metadata.domain,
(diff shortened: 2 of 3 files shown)


beeme1mr (approved):

lukas-reining (approved):
I am not sure about this one @toddbaert.
In the TS world, this is "wrong" in the sense that the type does not declare it as optional.
If we were to make this change, I think we have more places where we have nested props that are accessed non optionally where in JS, nothing would prevent from making it null or undefined.
The change is fine to me but I think we should be able to rely on TS here as we would otherwise have to ignore TS on several other places 🤔

toddbaert:
> I am not sure about this one @toddbaert. In the TS world, this is "wrong" in the sense that the type does not declare it as optional, as you already said. If we were to make this change, I think we have more places where we have nested props that are accessed non optionally where in JS, nothing would prevent from making it null or undefined. The change is fine to me but I think we should be able to rely on TS here as we would otherwise have to ignore TS on several other places 🤔

I see your point, and if we were just writing an internal TS library I would be OK with that. But don't we want to support JS as well? And if so, shouldn't we generally be defensive about this sort of thing? If I'm a JS dev, it's it quite possible I forget something as simple as this metadata?

toddbaert:
Draft for now.

lukas-reining:
> I see your point, and if we were just writing an internal TS library I would be OK with that. But don't we want to support JS as well? And if so, shouldn't we generally be defensive about this sort of thing? If I'm a JS dev, it's it quite possible I forget something as simple as this metadata?

I think we should support JS, but I also think that a TS interface simply has to be implemented correctly in JS and otherwise the error is not really avoidable. I mean we could go as fa as checking the existence of the evaluation methods if we wanted to check the provider interface compliance but I would not say this should be responsibility of the SDK.
Most IDEs/editors will check the type compliance by using the typedefs anyways.

Did you encounter an issue with these properties @toddbaert? I have seen it mentioned in #971 but there it was due to the casting of the incomplete type right?

PR #975, opened 2024-08-02
feat: updated the Metadata type to allow for custom properties

juanparadox:
## This PR
- Updates the `Metadata` type to a `Record` so that Providers can provide more than just the `name` property as part of the `metadata`.

### Notes
Conversation [here](https://cloud-native.slack.com/archives/C03J36ZP020/p1721747906370629).

### DIFF
--- a/packages/shared/src/client/client.ts
+++ b/packages/shared/src/client/client.ts
@@ -1,7 +1,6 @@
-import { Metadata } from '../types';
 import { ProviderMetadata } from '../provider/provider';
 
-export interface ClientMetadata extends Metadata {
+export interface ClientMetadata {
   /**
    * @deprecated alias of "domain", use domain instead
    */
--- a/packages/shared/src/provider/provider.ts
+++ b/packages/shared/src/provider/provider.ts
@@ -82,7 +82,7 @@ export { ClientProviderStatus as AllProviderStatus };
 /**
  * Static data about the provider.
  */
-export interface ProviderMetadata extends Metadata {
+export interface ProviderMetadata extends Readonly<Metadata> {
   readonly name: string;
 }
 
@@ -99,7 +99,7 @@ export interface CommonProvider<S extends ClientProviderStatus | ServerProviderS
   /**
    * @deprecated the SDK now maintains the provider's state; there's no need for providers to implement this field.
    * Returns a representation of the current readiness of the provider.
-   * 
+   *
    * _Providers which do not implement this method are assumed to be ready immediately._
    */
   readonly status?: S;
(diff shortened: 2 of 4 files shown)


packages/shared/src/client/client.ts
```diff
@@ -1,7 +1,6 @@
-import { Metadata } from '../types';
 import { ProviderMetadata } from '../provider/provider';
 
-export interface ClientMetadata extends Metadata {
```
juanparadox, line 4: `Metadata` wasn't doing much here since it was an empty interface.

packages/shared/src/provider/provider.ts
```diff
@@ -82,7 +82,7 @@ export { ClientProviderStatus as AllProviderStatus };
 /**
  * Static data about the provider.
  */
-export interface ProviderMetadata extends Metadata {
+export interface ProviderMetadata extends Readonly<Metadata> {
```
beeme1mr, line 85: Will read-only prevent providers from updating metadata? The metadata should be read-only to consumers but I think a provider should be able to update metadata at runtime.

packages/shared/src/types/metadata.ts
```diff
@@ -1,2 +1 @@
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface Metadata {}
+export type Metadata = Record<string, string>;
```
beeme1mr, line 1: @lukas-reining do you remember why this was an empty interface?

packages/shared/src/provider/provider.ts
juanparadox: I'll double check to see if TS prevents it for consumers and allows it for the provider. I am almost certain the provider can still set its own data. I had read the comment for `ProviderMetadata` which says it's static and so kept it `Readonly`.

toddbaert: This won't cause an issue for providers except if they want to write/mutate inner props of metadata, but this can be easily worked-around.

toddbaert (approved):

packages/shared/src/types/metadata.ts
lukas-reining: No, I think this was like this from the beginning, I think I just moved it when refactoring the SDK.
We should be able to just change it.

lukas-reining (approved):

packages/shared/src/client/client.ts
toddbaert: Maybe we should do `extends Readonly<Metadata>` here as well, so we can add arbitrary stuff in client's just the same as providers? I know it wasn't doing anything before, but with your changes now it actually adds functionality.

If you agree, we can make this change and merge.

beeme1mr (approved):

packages/shared/src/client/client.ts
beeme1mr: I'm not sure how custom metadata would be set on the client.

PR #980, opened 2024-08-22
fix: race condition in test provider with suspense

toddbaert:
Fixes an issues with the React `OpenFeatureTestProvider` where the provider was not ready until the next event loop tick when the `flagValueMap` was used.

Also removes the initialization in the client in-memory provider, since it was only doing some un-needed validation inconsistent with the server provider.

### DIFF
--- a/packages/react/src/provider/test-provider.tsx
+++ b/packages/react/src/provider/test-provider.tsx
@@ -43,6 +43,15 @@ type TestProviderProps = Omit<React.ComponentProps<typeof OpenFeatureProvider>, 
 
 // internal provider which is basically the in-memory provider with a simpler config and some optional fake delays
 class TestProvider extends InMemoryProvider {
+
+  // initially make this undefined, we still set it if a delay is specified
+  initialize: Provider['initialize'] = undefined;
+
+  // "place-holder" init function which we only assign if want a delay
+  private delayedInitialize = async () => {
+    await new Promise<void>((resolve) => setTimeout(resolve, this.delay));
+  };
+
   constructor(
     flagValueMap: FlagValueMap,
     private delay = 0,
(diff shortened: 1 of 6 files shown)


packages/react/test/test-provider.spec.tsx
```diff
@@ -73,27 +72,60 @@ describe('OpenFeatureTestProvider', () => {
...
       );
 
-      expect(await screen.findByText('👎')).toBeInTheDocument();
-      expect(await screen.findByText(/No-op/)).toBeInTheDocument();
+      expect(screen.getByText('👎')).toBeInTheDocument();
+      expect(screen.getByText(/No-op/)).toBeInTheDocument();
+    });
+  });
+
+  describe('component under test suspends', () => {
```
toddbaert, line 97: These failed before the fix.

packages/react/src/provider/test-provider.tsx
```diff
@@ -61,10 +70,8 @@ class TestProvider extends InMemoryProvider {
       };
     }, {});
     super(flagConfig);
-  }
-
-  async initialize(context?: EvaluationContext | undefined): Promise<void> {
-    await Promise.all([super.initialize(context), new Promise<void>((resolve) => setTimeout(resolve, this.delay))]);
+    // only define and init if there's a non-zero delay specified
+    this.initialize = this.delay ? this.delayedInitialize.bind(this) : undefined;
```
toddbaert, line 74: This is the fix. Before, we always delayed, even if only for 0 seconds. Because the `initialize` function was defined, and due to that delay, the provider was sometimes not ready by the time assertions ran leading to suspend or not-ready errors when they shouldn't be there.

packages/react/test/test-provider.spec.tsx
```diff
@@ -45,22 +45,21 @@ describe('OpenFeatureTestProvider', () => {
...
       render(
         <OpenFeatureTestProvider delayMs={delay} flagValueMap={{ [FLAG_KEY]: true }}>
-          <TestComponent />
+          <TestComponent suspend={false} />
         </OpenFeatureTestProvider>,
       );
 
       // should only be resolved after delay
-      expect(await screen.findByText('👎')).toBeInTheDocument();
+      expect(screen.getByText('👎')).toBeInTheDocument();
```
toddbaert, line 53: `getByText` is synchronous, which is much more reliable for these suspense-sensitive tests. I've updated all assertions to use this since `findByText` was hiding some bugs.

aepfli (approved):

beeme1mr (approved):

PR #987, opened 2024-08-26
feat(react): prevent rerenders when value is unchanged

beeme1mr:
## This PR

- prevents components from rendering if the resolved value hasn't been modified.

### Notes

The deep comparison may not cover all edge cases. However, I thought that was acceptable because a false negative would only trigger an unnecessary rerender.

## How to test

Tests are a work in progress.


packages/react/src/common/is-equal.ts
```diff
@@ -0,0 +1,14 @@
...
+ * @param {unknown} value First value to compare
+ * @param {unknown} other Second value to compare
+ * @returns {boolean} True if the values are equal
+ */
+export function isEqual(value: unknown, other: unknown): boolean {
+  if (value === other) {
+    return true;
+  }
+
+  return JSON.stringify(value) === JSON.stringify(other);
```
toddbaert, line 13: For performance, I don't think we should attempt this serialization if we're dealing with a primitive. Currently, this will attempt things like  `JSON.stringify(1) === JSON.stringify(2)`;

lukas-reining, line 13: Not sure if we care about this but if the object is not ordered the same order, this will resolve to false:
Could be considered correct, but also wrong here, really not sure.

```js
obj = {}; obj.name="X"; obj.os="linux";
JSON.stringify(obj);
obj = {}; obj.os="linux"; obj.name="X";
JSON.stringify(obj);
```

beeme1mr: Yeah, I considered that but thought it was unlikely to be an issue. I could either update the check to handle this use or update the JSDocs to mention the limitation. I could also remove this entirely and just do a reference comparison.

ajwootto: there's also the option of deep checking the object by traversing through it. Would a reference comparison for an object always return "false" since an updated configuration from the provider would usually end up re-parsing the incoming configuration and creating new objects from it?

beeme1mr: I've updated the check to traverse the object as @ajwootto suggested.

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -283,12 +286,17 @@ function attachHandlersAndResolve<T extends FlagValue>(
...
   );
 
   const updateEvaluationDetailsRef = () => {
-    setEvaluationDetails(resolver(client).call(client, flagKey, defaultValue, options));
+    const updatedEvaluationDetails = resolver(client).call(client, flagKey, defaultValue, options);
+
+    // Avoid re-rendering if the value hasn't changed
+    if (!isEqual(updatedEvaluationDetails.value, evaluationDetails.value)) {
+      setEvaluationDetails(updatedEvaluationDetails);
+    }
```
beeme1mr, line 299: This is the only real change in this file.

packages/react/src/common/is-equal.ts
beeme1mr: I've refactored the check.

aepfli (approved):

lukas-reining (approved):

PR #991, opened 2024-08-27
chore: move client/ dir to web/

toddbaert:
I've been meaning to do this for a while... it just renames the `client/` dir to `web/`. I think this is better because there will be less confusion around the OpenFeature client object (which has it's own dirs) and because it's more consistent with the associated artifact name "web-sdk".


lukas-reining (approved):

beeme1mr (approved):
We'll need to update the script in the docs once this is merged.

Issue #976, opened 2024-08-05
[FEATURE] Angular SDK implementation


juanparadox:
### Requirements

We should implement an Angular SDK. The SDK would provide an "Angular-friendly" API and features, built on top of the [Web SDK](https://github.com/open-feature/js-sdk/tree/main/packages/client).

The list is not conclusive and is up for discussion, including the removal, update, or addition to this list:

- [ ] An [injectable service](https://angular.dev/guide/di/creating-injectable-service) which will allow for the exposure of `OpenFeature` methods such as `setProvider` and `setContext` and for a way to get the provider's status
- [ ] A way to subscribe to a client so that the user can get a flag's value. For example:
```
this.openFeatureService.client$.subscribe(client => {
  client.getBooleanValue('key', false);
})
```
- [ ] Add handlers for when the context changes so that the subscribed client provides new flag values

I'm a little rusty with Angular since React is more my space, but I would like some opinions on how the subscription part of the developer experience should work. Reason being is because with my suggestions above, the user has to subscribe to a client inside of every consumer - is this the best way to allow the user to evaluate a value for a flag or can we do this a better way with Angular?

Another point to keep in mind is whether to build this using signals or observables?

Non-functional requirements:

- Consistent terminology and naming with Web SDK and OpenFeature spec
- Observe Angular idioms and patterns
- Helpful documentation around service injection, subscriptions and what the user should/can subscribe to
- Typings if applicable

lukas-reining:
Hey @juanparadox great that you started on this!
I have some questions: 

Where is the difference between these two requirements? Do you have an example for the second one?
- A way to subscribe to a client so that the user can get a flag's value.
- Add handlers for when the context changes so that the subscribed client provides new flag values

For the injection I could imagine injecting the client being good enough. We do something similar in the Nest SDK: 
https://github.com/open-feature/js-sdk/tree/main/packages/nest#usage

Maybe we could also add an injection of individual flag values e.g. as an observable? We do something similar with flag value promises in the Nest SDK. But with the static context we could maybe go for flag value observables that update automatically? 

> An [injectable service](https://angular.dev/guide/di/creating-injectable-service) which will allow for the exposure of OpenFeature methods such as setProvider and setContext and for a way to get the provider's status

In the Nest SDK we do this only on module init, but I think a thin injectable service to provide that functionality could be good.

juanparadox:
> Hey @juanparadox great that you started on this! I have some questions:
> 
[...]

PR #997, opened 2024-09-03
feat: Angular SDK

lukas-reining:
## This PR

Adds an Angular SDK as discussed in #976.
It mainly provides directives that conditionally render based on flag value and provider state.

I did not inject any OpenFeature service or anything like that, as I think the directives will be the way to access feature flags in the templates, and the OpenFeature global API can be used directly in services. But I am not 100% sure on that one.

Happy to receive feedback and opinions :)

A small test app can be found here: https://github.com/open-feature/angular-test-app
This can be used with [yalc](https://github.com/wclr/yalc).

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,10 +8,10 @@
     "e2e-server": "git submodule update --init --recursive && shx cp test-harness/features/evaluation.feature packages/server/e2e/features && jest --selectProjects=server-e2e --verbose",
     "e2e-web": "git submodule update --init --recursive && shx cp test-harness/features/evaluation.feature packages/web/e2e/features && jest --selectProjects=web-e2e --verbose",
     "e2e": "npm run e2e-server && npm run e2e-web",
-    "lint": "npm run lint --workspace=packages/shared --workspace=packages/server --workspace=packages/web --workspace=packages/react --workspace=packages/nest",
+    "lint": "npm run lint --workspace=packages/shared --workspace=packages/server --workspace=packages/web --workspace=packages/react --workspace=packages/angular --workspace=packages/nest",
     "clean": "shx rm -rf ./dist",
-    "build": "npm run build --workspace=packages/shared --workspace=packages/server --workspace=packages/web --workspace=packages/react --workspace=packages/nest",
-    "publish-all": "npm run publish-if-not-exists --workspace=packages/shared --workspace=packages/server --workspace=packages/web --workspace=packages/react --workspace=packages/nest",
+    "build": "npm run build --workspace=packages/shared --workspace=packages/server --workspace=packages/web --workspace=packages/react --workspace=packages/angular --workspace=packages/nest",
+    "publish-all": "npm run publish-if-not-exists --workspace=packages/shared --workspace=packages/server --workspace=packages/web --workspace=packages/react --workspace=packages/angular --workspace=packages/nest",
     "docs": "typedoc",
     "core-version": "npm run version --workspace=packages/shared",
     "update-core-peers": "export OPENFEATURE_CORE_VERSION=$(npm run --silent core-version) && npm run update-core-peer --workspace=packages/server --workspace=packages/web"
@@ -76,6 +76,7 @@
     "packages/server",
     "packages/web",
     "packages/react",
+    "packages/angular",
     "packages/nest"
   ]
 }
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "description": "OpenFeature SDK for JavaScript",
   "scripts": {
-    "test": "jest --selectProjects=shared --selectProjects=server --selectProjects=web --selectProjects=react --silent",
+    "test": "jest --selectProjects=shared --selectProjects=server --selectProjects=web --selectProjects=react --selectProjects=angular --silent",
     "e2e-server": "git submodule update --init --recursive && shx cp test-harness/features/evaluation.feature packages/server/e2e/features && jest --selectProjects=server-e2e --verbose",
     "e2e-web": "git submodule update --init --recursive && shx cp test-harness/features/evaluation.feature packages/web/e2e/features && jest --selectProjects=web-e2e --verbose",
     "e2e": "npm run e2e-server && npm run e2e-web",
@@ -43,8 +43,6 @@
     "@types/node": "^20.11.16",
     "@types/react": "^18.2.55",
     "@types/uuid": "^9.0.8",
-    "@typescript-eslint/eslint-plugin": "^7.0.0",
-    "@typescript-eslint/parser": "^7.0.0",
     "esbuild": "^0.20.0",
     "eslint": "^8.56.0",
     "eslint-config-prettier": "^9.1.0",
@@ -60,16 +58,22 @@
     "jest-environment-jsdom": "^29.7.0",
     "jest-environment-node": "^29.7.0",
     "jest-junit": "^16.0.0",
+    "ng-packagr": "^17.3.0",
     "prettier": "^3.2.5",
     "react": "^18.2.0",
     "rollup": "^3.29.4",
     "rollup-plugin-dts": "^5.3.1",
+    "rxjs": "~7.8.0",
     "shx": "^0.3.4",
     "ts-jest": "^29.1.2",
     "ts-node": "^10.9.2",
+    "tslib": "^2.3.0",
     "typedoc": "^0.25.7",
     "typescript": "^4.7.4",
     "uuid": "^9.0.1"
+  },
+  "overrides": {
+    "typescript": "^4.7.4"
   },
   "workspaces": [
     "packages/shared",
--- a/packages/angular/projects/angular-sdk/README.md
+++ b/packages/angular/projects/angular-sdk/README.md
@@ -30,30 +30,30 @@
 </p>
 <!-- x-hide-in-docs-start -->
 
-[OpenFeature](https://openfeature.dev) is an open specification that provides a vendor-agnostic, community-driven API for feature flagging that works with your favorite feature flag management tool or in-house solution.
+[OpenFeature](https://openfeature.dev) is an open specification that provides a vendor-agnostic, community-driven API
+for feature flagging that works with your favorite feature flag management tool or in-house solution.
 
 <!-- x-hide-in-docs-end -->
 
 ## Overview
 
-The OpenFeature Angular SDK adds Angular-specific functionality to the [OpenFeature Web SDK](https://openfeature.dev/docs/reference/technologies/client/web).
-
-In addition to the feature provided by the [web sdk](https://openfeature.dev/docs/reference/technologies/client/web), capabilities include:
+The OpenFeature Angular SDK adds Angular-specific functionality to
+the [OpenFeature Web SDK](https://openfeature.dev/docs/reference/technologies/client/web).
 
 ## Table of Contents
 
 - [Quick start](#quick-start)
-  - [Requirements](#requirements)
-  - [Install](#install)
-    - [npm](#npm)
-    - [yarn](#yarn)
-    - [Required peer dependencies](#required-peer-dependencies)
-  - [Minimal Example](#minimal-example)
-  - [Flag Types](#flag-types)
-    - [Boolean Feature Flag](#boolean-feature-flag)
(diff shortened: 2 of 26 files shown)


beeme1mr:
I've reviewed the flagged licenses and added exceptions in FOSSA.

package.json
```diff
@@ -60,22 +58,29 @@
...
     "shx": "^0.3.4",
     "ts-jest": "^29.1.2",
     "ts-node": "^10.9.2",
+    "tslib": "^2.3.0",
     "typedoc": "^0.25.7",
     "typescript": "^4.7.4",
     "uuid": "^9.0.1"
   },
+  "overrides": {
+    "typescript": "^4.7.4"
```
beeme1mr, line 76: What is this for?

packages/angular/projects/angular-sdk/README.md
```diff
@@ -0,0 +1,206 @@
...
+    - [Object Feature Flag](#object-feature-flag)
+- [Inputs](#inputs)
+- [Contributing](#contributing)
+- [License](#license)
+
+## Quick start
+
+### Requirements
+
+- ES2022-compatible web browser (Chrome, Edge, Firefox, etc)
```
beeme1mr, line 65: We updated React to be compatible with ES2015. Can we do that here, too?

packages/angular/projects/angular-sdk/README.md
```diff
@@ -0,0 +1,206 @@
...
+    // Other components
+  ],
+  imports: [
+    CommonModule,
+    OpenFeatureModule.forRoot({
+      provider: yourFeatureProvider,
+      domainBoundProviders: {
+        domain1: new YourOpenFeatureProvider(),
+        domain2: new YourOtherOpenFeatureProvider(),
+      },
```
beeme1mr, line 114: It may be worth adding a comment saying domains are optional.

beeme1mr (commented):
I've left some quick feedback, but I would like to play around with your demo app before approving it.

package.json
lukas-reining: `"jest-preset-angular": "^14.2.2"` requires typescript <= 5.2 but it does not really need it.
To have tests working in the whole npm workspace, this is needed to be able to use the "workspace ts" version instead of the one required. But it is not actually a problem for `jest-preset-angular`.
I also considered updating TS to 5.x but there were other libs involved then.

packages/angular/projects/angular-sdk/README.md
lukas-reining: added it :)

packages/angular/projects/angular-sdk/README.md
lukas-reining: did so

packages/angular/projects/angular-sdk/src/public-api.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Public API Surface of angular
+ */
+
+export * from './lib/feature-flag.directive';
+export * from './lib/open-feature.module';
+
+// re-export the web-sdk so consumers can access that API from the angular-sdk
```
toddbaert, line 8: Thanks for this!

packages/angular/projects/angular-sdk/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+<!-- x-hide-in-docs-end -->
+<!-- The 'github-badges' class is used in the docs -->
+<p align="center" class="github-badges">
+  <a href="https://github.com/open-feature/spec/releases/tag/v0.8.0">
+    <img alt="Specification" src="https://img.shields.io/static/v1?label=specification&message=v0.8.0&color=yellow&style=for-the-badge" />
+  </a>
+  <!-- x-release-please-start-version -->
+  <a href="https://github.com/open-feature/js-sdk/releases/tag/angular-sdk-v0.0.0">
+    <img alt="Release" src="https://img.shields.io/static/v1?label=release&message=v0.4.0&color=blue&style=for-the-badge" />
```
toddbaert, line 20:
```suggestion
    <img alt="Release" src="https://img.shields.io/static/v1?label=release&message=v0.0.0&color=blue&style=for-the-badge" />
```

packages/angular/projects/angular-sdk/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+[OpenFeature](https://openfeature.dev) is an open specification that provides a vendor-agnostic, community-driven API for feature flagging that works with your favorite feature flag management tool or in-house solution.
+
+<!-- x-hide-in-docs-end -->
+
+## Overview
+
+The OpenFeature Angular SDK adds Angular-specific functionality to the [OpenFeature Web SDK](https://openfeature.dev/docs/reference/technologies/client/web).
+
+In addition to the feature provided by the [web sdk](https://openfeature.dev/docs/reference/technologies/client/web), capabilities include:
```
toddbaert, line 41: Are we missing something here? Probably mention the directives?

packages/angular/projects/angular-sdk/README.md
```diff
@@ -0,0 +1,206 @@
...
+<ng-template #numberFeatureElse>
+  This is shown when the feature flag does not match the specified discount rate.
+</ng-template>
+<ng-template #numberFeatureInitializing>
+  This is shown when the feature flag is initializing.
+</ng-template>
+<ng-template #numberFeatureReconciling>
+  This is shown when the feature flag is reconciling.
+</ng-template>
+```
```
toddbaert, line 172: Something that I think could be nice here, is to add an `exportAs` value for this directive (`exportAs: 'evaluationDetails'`), and bind the flag details to that... then in the templates, you could access and render the flag values like `#details="evaluationDetails"` and then `{{ details.value }}`

I'm not sure if this would be valuable for booleans, but I think it would be for others.

toddbaert:
This all looks really, good, but I'm wondering what you think about this: https://github.com/open-feature/js-sdk/pull/997/files#r1746120679

packages/angular/projects/angular-sdk/README.md
toddbaert: Thinking about it more, I think that the first 2 templates don't make so much sense for non-booleans. I think most users won't actually be interested in showing a template if the string/number/object value matches. I think it would be more useful just to have 3 templates for these: a primary(with flag value injected)/initializing/reconciling template.

I'm not also not 100% sure there's much value in differentiating between the Initializing and reconciling templates; in react, we just show the suspense in both cases and I suspect that's probably all people would need.

lukas-reining: > Something that I think could be nice here, is to add an exportAs value for this directive (exportAs: 'evaluationDetails'), and bind the flag details to that... then in the templates, you could access and render the flag values like #details="evaluationDetails" and then {{ details.value }}

Definitely, but it was already in. I added some docs for that :) 
https://github.com/open-feature/js-sdk/pull/997/files#diff-41afcfae7e1b2d20dffab3b5bd15a8858a79acb6f7b468e45f154e9eecf2d539R231

lukas-reining: > Thinking about it more, I think that the first 2 templates don't make so much sense for non-booleans. I think most users won't actually be interested in showing a template if the string/number/object value matches. I think it would be more useful just to have 3 templates for these: a primary(with flag value injected)/initializing/reconciling template.

I see the point. We actually have the use case of conditionally rendering based on the value with an else.

I changed the `value` to be optional, so it can be used in that way.
If you don't give a value, the template is always rendered (besides if initializing and reconciling are given) and if you give a value and it does not match either noting, or if given, the else template is rendered. 

I also described that a bit more in the docs.
This API allows all the modes I think.

```html
<div *stringFeatureFlag="'themeColor'; value: 'dark'; default: 'light'; else: stringFeatureElse; let value; let details = evaluationDetails">
  It was a match!
  The theme color is {{ value }} because of {{ details.reason }}
</div>
<ng-template #stringFeatureElse let-value let-details="evaluationDetails">
  It was no match!
  The theme color is {{ value }} because of {{ details.reason }}
</ng-template>
```

```html
<div *stringFeatureFlag="'themeColor'; default: 'light'; let value;">
  The theme color is {{ value }}.
</div>
```

lukas-reining: >I'm not also not 100% sure there's much value in differentiating between the Initializing and reconciling templates; in react, we just show the suspense in both cases and I suspect that's probably all people would need.

I was also wondering about it but I think the way it is here gives most flexibility.
We had cases were we did not want to see the reconciling but the loading, you can do that by simply omitting the reconciling template here. 
If you want to have it for both, you can simply set the same "loading template" to both parameters.

```
<div *booleanFeatureFlag="'flag'; default: false; initializing: loading; reconciling: loading">
  This is shown when the feature flag matches the specified user configuration.
</div>
<ng-template #loading>
  This is shown when the feature flag is initializing or reconciling.
</ng-template>
```

toddbaert: > Definitely, but it was already in. Or do you mean it differently @toddbaert? I added some docs for that :)

Oh, I'm sorry! I missed that.

toddbaert:
I think this looks like enough for an initial release!

toddbaert (approved):

release-please-config.json
```diff
@@ -27,6 +27,14 @@
       "extra-files": ["README.md"],
       "versioning": "default"
     },
+    "packages/angular/projects/angular-sdk": {
+      "release-type": "node",
+      "prerelease": false,
```
toddbaert, line 32:
```suggestion
      "prerelease": true,
```

beeme1mr (approved):

packages/angular/projects/angular-sdk/README.md
```diff
@@ -0,0 +1,258 @@
...
+This parameter is _optional_.
+
+The template referenced in `initializing` and `reconciling` will be rendered if OpenFeature provider is in the
+corresponding states.
+This parameter is _optional_, if omitted, the `then` and `else` templates will be rendered according to the flag value.
+
+##### Boolean Feature Flag
+
+```html
+<div *booleanFeatureFlag="'isFeatureEnabled'; default: true; domain: 'userDomain'; else: booleanFeatureElse; initializing: booleanFeatureInitializing; reconciling: booleanFeatureReconciling">
```
juanparadox, line 166: Praise: I really like how you can account for the initializing and reconciling states! Super cool.

packages/angular/projects/angular-sdk/README.md
```diff
@@ -0,0 +1,258 @@
...
+@NgModule({
+  declarations: [
+    // Other components
+  ],
+  imports: [
+    CommonModule,
+    OpenFeatureModule.forRoot({
+      provider: yourFeatureProvider,
+      // domainBoundProviders are optional, mostly needed if more than one provider is needed
+      domainBoundProviders: {
```
juanparadox, line 113: Question: Should we include how to set a new Provider outside of using `OpenFeatureModule.forRoot`? For example, if someone needs to set a provider after the app initializes?

toddbaert: Not a bad thing to add but also not something I'd consider a blocker. I think most Angular people would "just get it".

toddbaert:
I'm comfortable with releasing this, but we could also wait until after this weeks meeting - and maybe get a demo there, though I don't want to volunteer you for anything @lukas-reining

packages/angular/projects/angular-sdk/README.md
lukas-reining: I will add a sentence to this and also explaining that a service can simply be taken from the GlobalApi.

lukas-reining:
> I'm comfortable with releasing this, but we could also wait until after this weeks meeting - and maybe get a demo there, though I don't want to volunteer you for anything @lukas-reining

This sounds good, I will do a quick demo :) 
Will polish the test app a bit to show off the features and then we can merge maybe the day after the community meeting if nothing comes up.

Building a nice demo is the hard part actually :D

juanparadox (approved):

lukas-reining:
I would merge it this afternoon if noting comes up until then @toddbaert @beeme1mr @juanparadox :)

PR #1020, opened 2024-09-20
feat: implement tracking as per spec

toddbaert:
:mega: This was a draft for a while, but is now ready for review! :mega: 

This implements tracking as per spec, in the server, web, and react SDKs.
I don't think the Angular or Nest SDKs need specific implementations, but please advise (cc @luizgribeiro @lukas-reining).

Fixes: https://github.com/open-feature/js-sdk/issues/1033
Fixes: https://github.com/open-feature/js-sdk/issues/1034


packages/shared/src/tracking/occurrence.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * A container for arbitrary contextual data that can be used as a basis for dynamic evaluation
+ */
+export type OccurrenceDetails = {
+  /**
+   * A string uniquely identifying the subject (end-user, or client service) of a flag evaluation.
+   * Providers may require this field for fractional flag evaluation, rules, or overrides targeting specific users.
+   * Such providers may behave unpredictably if a targeting key is not specified at flag resolution.
+   */
```
toddbaert, line 18:
```suggestion
  /**
   * A numeric value associated with this event.
   */
```

packages/web/src/tracking/tracking.ts
```diff
@@ -0,0 +1,11 @@
+import { OccurrenceDetails } from '@openfeature/core';
```
beeme1mr, line 1:
```suggestion
import types { OccurrenceDetails } from '@openfeature/core';
```

packages/web/src/provider/provider.ts
```diff
@@ -71,4 +71,12 @@ export interface Provider extends CommonProvider<ClientProviderStatus> {
...
     logger: Logger,
   ): ResolutionDetails<T>;
+
+  /**
+   * Track a thing
+   * @param occurrenceKey
+   * @param context
+   * @param occurrenceDetails
+   */
+    track?(occurrenceKey: string, context: EvaluationContext, occurrenceDetails: OccurrenceDetails): void;
```
moredip, line 81: I think `event` or `eventName` is more idiomatic and recognizable than `occurenceKey` (closer to the defacto standard of [Segment's SDK](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#track)) 
```suggestion
    track?(event: string, context: EvaluationContext, occurrenceDetails: OccurrenceDetails): void;
```

packages/web/src/tracking/tracking.ts
weyert: `import type`

packages/server/src/provider/provider.ts
```diff
@@ -57,4 +65,12 @@ export interface Provider extends CommonProvider<ServerProviderStatus> {
     context: EvaluationContext,
     logger: Logger,
   ): Promise<ResolutionDetails<T>>;
+
+  /**
+   * Track a thing
+   * @param occurrenceKey
+   * @param context
+   * @param occurrenceDetails
```
lukas-reining, line 73: Are we doing these later?

packages/shared/src/tracking/occurrence.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type OccurrenceValue =
+  | PrimitiveValue
+  | Date
+  | { [key: string]: OccurrenceValue }
+  | OccurrenceValue[];
+
+/**
+ * A container for arbitrary contextual data that can be used as a basis for dynamic evaluation
+ */
```
lukas-reining, line 12: This is copy-paste right? Seems to make for sense for evaluation context.

toddbaert: Yep - this is still quite rough - we aren't settle on terminology yet and comments aren't updated.

packages/server/src/provider/provider.ts
toddbaert: https://github.com/open-feature/js-sdk/pull/1020#discussion_r1779083185

packages/shared/src/tracking/occurrence.ts
lukas-reining: Sure this is great, just wanted to be sure!

aepfli (approved):

packages/web/src/tracking/tracking.ts
toddbaert: I've fixed this is a separate PR with a new eslint rule, nice catch.

toddbaert:
This is ready for review now.

packages/web/src/client/internal/open-feature-client.ts
```diff
@@ -181,6 +182,21 @@ export class OpenFeatureClient implements Client {
     return this.evaluate<T>(flagKey, this._provider.resolveObjectEvaluation, defaultValue, 'object', options);
   }
 
+  track(occurrenceKey: string, occurrenceDetails: TrackingEventDetails): void {
+
+    this.shortCircuitIfNotReady();
```
beeme1mr, line 187: Do we want to throw if the provider is unhealthy? I expected the track method to never throw.

packages/web/src/client/internal/open-feature-client.ts
```diff
@@ -181,6 +182,21 @@ export class OpenFeatureClient implements Client {
...
+
+    const context = {
```
beeme1mr, line 189: It may be worth adding a comment explaining why you're creating a new context object. I'm assuming it's because you don't want the provider to be able to modify context in the track method.

packages/server/src/client/internal/open-feature-client.ts
```diff
@@ -223,6 +224,19 @@ export class OpenFeatureClient implements Client {
...
   }
 
+  track(occurrenceKey: string, context: EvaluationContext, occurrenceDetails: TrackingEventDetails): void {
+
+    this.shortCircuitIfNotReady();
+
+    const mergedContext = this.mergeContexts(context);
+
+    if (typeof this._provider.track === 'function') {
+      return this._provider.track?.(occurrenceKey, mergedContext, occurrenceDetails);
```
beeme1mr, line 234: Shouldn't we log and swallow any exceptions thrown by the provider?

toddbaert: Definitely; this would fall into our spec rules for client ops not abnormally terminating. I will add some error handling.

packages/web/src/client/internal/open-feature-client.ts
toddbaert: We want to throw here but catch it, I think. I just made it short circuit like all the other methods.

packages/server/src/client/internal/open-feature-client.ts
toddbaert: Added

packages/web/src/client/internal/open-feature-client.ts
toddbaert: I did this and also froze the context.

packages/web/src/client/internal/open-feature-client.ts
toddbaert: done

packages/server/src/client/internal/open-feature-client.ts
```diff
@@ -380,4 +383,23 @@ export class OpenFeatureClient implements Client {
...
+    // merge global and client contexts
+    return {
+      ...OpenFeature.getContext(),
+      ...OpenFeature.getTransactionContext(),
+      ...this._context,
+      ...invocationContext,
+    };
+  }
+
+  private shortCircuitIfNotReady() {
```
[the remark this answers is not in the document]
toddbaert: This is harder than it sounds because the server/web actually both use different enums since there's more types in client.

packages/react/src/tracking/use-track.ts
```diff
@@ -0,0 +1,29 @@
+import type { TrackingEventDetails } from '@openfeature/web-sdk';
+import { useCallback } from 'react';
+import { useOpenFeatureClient } from '../provider';
+
+export type Track = {
```
beeme1mr, line 5: Couldn't this type live in core?

packages/web/src/client/internal/open-feature-client.ts
```diff
@@ -181,6 +182,26 @@ export class OpenFeatureClient implements Client {
...
 
+  track(occurrenceKey: string, occurrenceDetails: TrackingEventDetails): void {
+    try {
+      this.shortCircuitIfNotReady();
+
+      // copy and freeze the context
+      const context = {
+        ...OpenFeature.getContext(this?.options?.domain),
+      };
+      Object.freeze(context);
```
beeme1mr, line 193: Please move this under the if condition.

packages/web/src/client/internal/open-feature-client.ts
```diff
@@ -317,4 +333,13 @@ export class OpenFeatureClient implements Client {
   private get _logger() {
     return this._clientLogger || this.globalLogger();
   }
+
+  private shortCircuitIfNotReady() {
```
beeme1mr, line 337: Perhaps we could move this to a utils method in core to avoid duplication.

packages/react/src/tracking/use-track.ts
toddbaert: It's used only in React; it's the type of the exported track hook - it happens to look like the track method but it's not a method.

packages/web/src/client/internal/open-feature-client.ts
toddbaert: See: https://github.com/open-feature/js-sdk/pull/1020#discussion_r1804558887

packages/react/src/tracking/use-track.ts
toddbaert: I can do `track: Tracking['track'];` actually, since this function actually matches that type.

beeme1mr (approved):

packages/web/src/provider/in-memory-provider/in-memory-provider.ts
```diff
@@ -38,8 +38,7 @@ export class InMemoryProvider implements Provider {
    */
   async putConfiguration(flagConfiguration: FlagConfiguration) {
     try {
-      const flagsChanged = Object.entries(flagConfiguration)
-        .filter(([key, value]) => this._flagConfiguration[key] !== value)
+      const flagsChanged = Object.entries({...flagConfiguration, ...this._flagConfiguration})
```
wichopy, line 41: Since you are trying to merge old and new configs, shouldn't the new config come last?
```suggestion
      const flagsChanged = Object.entries({...this._flagConfiguration,...flagConfiguration })
```

wichopy: maybe need a test case for this?

toddbaert: reverted

wichopy (approved):
Lgtm. Would be nice to hook it up to @beeme1mr s test app?

aepfli (approved):

toddbaert:
> Lgtm. Would be nice to hook it up to @beeme1mr s test app?

That's the plan!

packages/server/src/client/internal/open-feature-client.ts
```diff
@@ -380,4 +386,23 @@ export class OpenFeatureClient implements Client {
   private get _logger() {
     return this._clientLogger || this.globalLogger();
   }
+
+  private mergeContexts(invocationContext: EvaluationContext) {
+    // merge global and client contexts
+    return {
+      ...OpenFeature.getContext(),
+      ...OpenFeature.getTransactionContext(),
```
lukas-reining, line 394: Maybe instead if using the OpenFeature global API to get the context we could add a context accessor function as we do here: https://github.com/open-feature/js-sdk/pull/1020/files#diff-8a151eb52e8ba25881ad3b837c98a32807dabf7b7f771518248edb6206e5add9R51
I would like this as we would not have a circular dependency between the global API and the client then.

But this would also apply to `OpenFeature.getHooks` used in this class and it might be out of context for this API. So we can resolve this, just wanted to mention it and see if we want to do it now.

packages/server/src/provider/provider.ts
```diff
@@ -58,4 +66,12 @@ export interface Provider extends CommonProvider<ServerProviderStatus> {
...
     logger: Logger,
   ): Promise<ResolutionDetails<T>>;
+
+  /**
+   * Track a user action or application state, usually representing a business objective or outcome.
+   * @param trackingEventName
+   * @param context
+   * @param trackingEventDetails
+   */
+  track?(trackingEventName: string, context?: EvaluationContext, trackingEventDetails?: TrackingEventDetails): void;
```
lukas-reining, line 76: This is the same for server and client, I think this could live in the `CommonProvider` like `initialize` and `onClose`.

lukas-reining (approved):
Looks good! Left two comments :)

packages/server/src/provider/provider.ts
toddbaert: Oh you are right I think. Will do.

toddbaert:
I'm merge this tomorrow unless I hear objections

packages/server/src/client/internal/open-feature-client.ts
lukas-reining: Great just a nice thing to "clean up"! :)

PR #1024, opened 2024-09-24
feat(react): avoid re-resolving flags unaffected by a change event

beeme1mr:
## This PR

- avoid re-resolving flags unaffected by a change event

### Notes

If the provider sends the change event payload, the React SDK uses the list of change flags. If the list is missing or empty, all events are triggered. This is a way to avoid sending unnecessary evaluation telemetry data.

### DIFF
--- a/packages/react/src/common/is-empty.ts
+++ b/packages/react/src/common/is-empty.ts
@@ -1,0 +1,24 @@
+/**
+ * Check if a value is empty
+ * @param {unknown} value - Value to check
+ * @returns {boolean} True if the value is empty
+ */
+export function isEmpty(value: unknown): boolean {
+  if (value === null || value === undefined) {
+    return true;
+  }
+
+  if (typeof value === 'string' && value.trim() === '') {
+    return true;
+  }
+
+  if (Array.isArray(value) && value.length === 0) {
+    return true;
+  }
+
+  if (typeof value === 'object' && Object.keys(value).length === 0) {
+    return true;
+  }
+
+  return false;
+}
--- a/packages/react/test/evaluation.spec.tsx
+++ b/packages/react/test/evaluation.spec.tsx
@@ -1,17 +1,15 @@
-import {
-  EvaluationContext,
-  Hook,
-  InMemoryProvider,
-  OpenFeature,
-  StandardResolutionReasons,
-  EvaluationDetails,
-  ErrorCode,
-} from '@openfeature/web-sdk';
 import '@testing-library/jest-dom'; // see: https://testing-library.com/docs/react-testing-library/setup
 import { act, render, renderHook, screen, waitFor } from '@testing-library/react';
 import * as React from 'react';
 import {
+  ErrorCode,
+  EvaluationContext,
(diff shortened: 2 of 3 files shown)


packages/react/test/evaluation.spec.tsx
```diff
@@ -415,6 +447,26 @@ describe('evaluation', () => {
...
+        });
+
+        expect(screen.queryByTestId('render-count')).toHaveTextContent('3');
+        expect(resolverSpy).toHaveBeenNthCalledWith(
+          3,
+          BOOL_FLAG_KEY,
+          expect.anything(),
+          expect.anything(),
+          expect.anything(),
+        );
```
beeme1mr, line 469: This assertion would have failed before [yesterday's hotfix](https://github.com/open-feature/js-sdk/pull/1021).

packages/react/src/common/is-empty.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (Array.isArray(value) && value.length === 0) {
+    return true;
+  }
+
+  if (typeof value === 'object' && Object.keys(value).length === 0) {
+    return true;
+  }
+
+  return false;
+}
```
lukas-reining, line 24: I know that this is just for checking if the `changedFlags` are empty but why are we checking for empty strings here? It feels quite generic for the concrete use case as we are just interested in `changedFlags` being `undefined` or empty `string[]` right? [[1]](https://github.com/open-feature/js-sdk/blob/e6adae04a5f3481317a855968ad846c72af4a087/packages/shared/src/events/eventing.ts#L26 )

And if I am overseeing something here, I am not sure about the semantic of `isEmpty` here as e.g. lodash defines   `isEmpty` differently regarding the non enumerable values [[2]](https://github.com/lodash/lodash/blob/6a2cc1dfcf7634fea70d1bc5bd22db453df67b42/src/isEmpty.ts#L72 ) but this might be fine, just want to bring it up.

[[1]](https://github.com/open-feature/js-sdk/blob/e6adae04a5f3481317a855968ad846c72af4a087/packages/shared/src/events/eventing.ts#L26 ): https://github.com/open-feature/js-sdk/blob/e6adae04a5f3481317a855968ad846c72af4a087/packages/shared/src/events/eventing.ts#L26
[[2]](https://github.com/lodash/lodash/blob/6a2cc1dfcf7634fea70d1bc5bd22db453df67b42/src/isEmpty.ts#L72 ): https://github.com/lodash/lodash/blob/6a2cc1dfcf7634fea70d1bc5bd22db453df67b42/src/isEmpty.ts#L72

lukas-reining (approved):
Looks good! Just left a small nit/question.

packages/react/src/common/is-empty.ts
beeme1mr: Yeah, I agree. I'll try to find a cleaner way to handle this.

toddbaert (approved):

toddbaert:
I removed the (react) from the title since this will be added only to the react changelog.

PR #1029, opened 2024-10-01
chore(angular): add module mapper to angular tests

beeme1mr:
Attempt to fix: https://github.com/open-feature/js-sdk/actions/runs/11131081055/job/30931849487


toddbaert:
See https://github.com/open-feature/js-sdk/pull/1030 instead.

PR #1031, opened 2024-10-06
feat: use mutate context hook

### LINKED ISSUE #968: [FEATURE] React SDK: Provide hook to set relevant context "`useContextMutator`"
### Background

The `OpenFeatureProvider` React context provider binds an OpenFeature client to a React Context. Components within that boundary use the associated client for evaluations, etc. The client can be associated with a domain (see https://github.com/open-feature/js-sdk/blob/main/packages/react/README.md#multiple-providers-and-domains and https://github.com/open-feature/js-sdk/blob/main/packages/client/README.md#domains)

We do not currently have a hook that makes it easy to update this context. It's necessary to use the global object to do this. We should add a hook that allows the correct context for the enclosed scope to be set.

For example, this current implementation:

```tsx
<OpenFeatureProvider domain={'my-domain'}>
  <button
    onClick={() => {
      OpenFeature.setContext('my-domain', updatedContext);
    }}
  >
      click here
  </button>
</OpenFeatureProvider>
```

Could be instead be:

```tsx
<OpenFeatureProvider domain={'my-domain'}>
  <button
    onClick={() => {
      mutateContext(updatedContext); // function returned by new hook, associated with the 'my-domain' domain
    }}
  >
      click here
  </button>
</OpenFeatureProvider>
```

This would make things easier for users, and fit better into react idioms. Please test your solution but using it in the sample react app [here](https://github.com/open-feature/react-test-app/blob/85b48aafc50eae2d6dab1e24189b8e4c0fd44e08/src/demos/ContextChangeDemo.tsx#L82-L84).

### Requirements

- new hook added which returns function to mutate context (please come up with a reasonable name for the hook and returned function)
- associated tests
- associated README updates
- React demo updated [here](https://github.com/open-feature/react-test-app/blob/85b48aafc50eae2d6dab1e24189b8e4c0fd44e08/src/demos/ContextChangeDemo.tsx#L82-L84)
- returned func(s) should probably be async

### Additional Considerations

I'd love to see considerations from the implementor on the following:

- should we export a couple different functions? `setContext`, `mergeContext` etc, which merge the context?
- should we ignore changes if the reference hasn't changed?

wichopy:
## This PR

New hook for mutating context instead of using the global object

### Related Issues

Fixes #968

### DIFF
--- a/packages/react/src/provider/use-context-mutator.ts
+++ b/packages/react/src/provider/use-context-mutator.ts
@@ -1,5 +1,6 @@
 import { useCallback, useContext, useRef } from 'react';
-import { OpenFeature, EvaluationContext } from '@openfeature/web-sdk';
+import type { EvaluationContext } from '@openfeature/web-sdk';
+import { OpenFeature } from '@openfeature/web-sdk';
 import { Context } from './context';
 
 /**
--- a/packages/react/src/provider/use-context-mutator.ts
+++ b/packages/react/src/provider/use-context-mutator.ts
@@ -8,19 +8,21 @@ import { Context } from './context';
  *
  */
 export function useContextMutator() {
-  const { domain } = useContext(Context) || {};
+    const { domain } = useContext(Context) || {};
+    const previousContext = useRef(null);
 
-  async function mutateContext(updatedContext: EvaluationContext): Promise<void> {
-    if (!domain) {
-      // Set the global context
-      OpenFeature.setContext(updatedContext);
-      return;
-    }
+    const mutateContext = useCallback(async (updatedContext: EvaluationContext) => {
+        if (previousContext.current !== updatedContext) {
+            if (!domain) {
+                OpenFeature.setContext(updatedContext);
+            } else {
+                OpenFeature.setContext(domain, updatedContext);
+            }
+            previousContext.current = updatedContext;
+        }
+    }, [domain]);
 
-    OpenFeature.setContext(domain, updatedContext);
-  }
-
-  return {
-    mutateContext,
-  };
+    return {
+        mutateContext,
+    };
 }
--- a/packages/react/src/provider/use-context-mutator.ts
+++ b/packages/react/src/provider/use-context-mutator.ts
@@ -1,4 +1,4 @@
-import { useContext } from 'react';
+import { useCallback, useContext, useRef } from 'react';
 import { OpenFeature, EvaluationContext } from '@openfeature/web-sdk';
 import { Context } from './context';
 
@@ -9,7 +9,7 @@ import { Context } from './context';
  */
 export function useContextMutator() {
     const { domain } = useContext(Context) || {};
-    const previousContext = useRef(null);
+    const previousContext = useRef<null | EvaluationContext>(null);
 
     const mutateContext = useCallback(async (updatedContext: EvaluationContext) => {
         if (previousContext.current !== updatedContext) {
--- a/packages/react/src/provider/use-context-mutator.ts
+++ b/packages/react/src/provider/use-context-mutator.ts
@@ -8,11 +8,13 @@ import { Context } from './context';
  *
  */
 export function useContextMutator() {
-  async function mutateContext(updatedContext: EvaluationContext): Promise<void> {
-    const { domain } = useContext(Context) || {};
+  const { domain } = useContext(Context) || {};
 
+  async function mutateContext(updatedContext: EvaluationContext): Promise<void> {
     if (!domain) {
-      throw new Error('No domain set for your context');
+      // Set the global context
+      OpenFeature.setContext(updatedContext);
+      return;
     }
 
     OpenFeature.setContext(domain, updatedContext);
--- a/packages/react/src/provider/use-context-mutator.ts
+++ b/packages/react/src/provider/use-context-mutator.ts
@@ -7,13 +7,20 @@ import { Context } from './context';
  * A hook for accessing context mutating functions.
  *
  */
-export function useContextMutator() {
+export function useContextMutator({
+  setGlobal
+}: {
+  /**
+   * Apply changes to the global context instead of the domain scoped context applied at the React Provider
+   */
+  setGlobal?: boolean;
+} = {}) {
     const { domain } = useContext(Context) || {};
     const previousContext = useRef<null | EvaluationContext>(null);
 
-    const mutateContext = useCallback(async (updatedContext: EvaluationContext) => {
+    const setContext = useCallback(async (updatedContext: EvaluationContext) => {
         if (previousContext.current !== updatedContext) {
-            if (!domain) {
+            if (!domain || setGlobal) {
                 OpenFeature.setContext(updatedContext);
             } else {
                 OpenFeature.setContext(domain, updatedContext);
@@ -23,6 +30,6 @@ export function useContextMutator() {
     }, [domain]);
 
     return {
-        mutateContext,
+        setContext,
     };
 }
--- a/packages/react/src/provider/use-context-mutator.ts
+++ b/packages/react/src/provider/use-context-mutator.ts
@@ -2,28 +2,22 @@ import { OpenFeature, EvaluationContext } from '@openfeature/web-sdk';
 
 type DomainContextMutator = (domain: string, updatedContext: EvaluationContext) => Promise<void>; 
 
-type ContextMutatorReturn = {
-  mutateContext: DomainContextMutator;
-}
 /**
  * 
  * A hook for accessing context mutating functions.
  * 
- * @returns {ContextMutatorReturn} 
  */
-export function useContextMutator(): {
-  mutateContext: DomainContextMutator;
-} {
-  return {
-    /**
-     * 
-     * Mutates the evaluation context for a given domain.
-     * 
-     * @param {string}domain
-     * @param {EvaluationContext} updatedContext
-     */
-    mutateContext: async (domain: string, updatedContext: EvaluationContext) => {
-      await OpenFeature.setContext(domain, updatedContext);
+export function useContextMutator() {
+
+  async function mutateContext(domainOrUpdatedContext: string | EvaluationContext, updatedContextOrUndefined?: EvaluationContext): Promise<void> {
+    if (typeof domainOrUpdatedContext === 'string' && updatedContextOrUndefined) {
+      await OpenFeature.setContext(domainOrUpdatedContext, updatedContextOrUndefined);
+    } else if (typeof domainOrUpdatedContext !== 'string') {
+      OpenFeature.setContext(domainOrUpdatedContext);
     }
+  }
+  
+  return {
+    mutateContext, 
   };
 }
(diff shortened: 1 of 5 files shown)


packages/react/src/provider/use-context-mutator.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    if (!domain) {
+      throw new Error('No domain set for your context');
+    }
+
+    OpenFeature.setContext(domain, updatedContext);
+  }
+
+  return {
+    mutateContext,
```
wichopy, line 22: I saw the comment about merging contexts, I think I need to know more about use cases in the real world. For example I saw one project setting the global context instead of a domain scoped one in react, is that also a valid use case? Should we just stick with updating the context for a domain for now and expand on this hook later?

packages/react/src/provider/use-context-mutator.ts
```diff
@@ -0,0 +1,23 @@
...
+type DomainContextMutator = (domain: string, updatedContext: EvaluationContext) => Promise<void>; 
+
+/**
+ * 
+ * A hook for accessing context mutating functions.
+ * 
+ */
+export function useContextMutator() {
+
+  async function mutateContext(domainOrUpdatedContext: string | EvaluationContext, updatedContextOrUndefined?: EvaluationContext): Promise<void> {
```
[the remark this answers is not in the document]
wichopy: Starting to think about how to write some tests. Looking at the readme I see you can set a context by domain but also have a global context when no domain is set:

https://github.com/open-feature/js-sdk/tree/main/packages/react#openfeatureprovider-context-provider

Could you confirm these testing requirements:

`test('should modify context for the domain if one is present')`

`test('should modify context for global scope when no domain is found')`

beeme1mr:
Could you please sign off on your commits? It's basically a way to tell the CNCF that you're willing and able to donate this code.

To add your Signed-off-by line to every commit in this branch:

- Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
- In your local branch, run: git rebase HEAD~5 --signoff
- Force push your changes to overwrite the branch: git push --force-with-lease origin wichopy/968

> For future commits, you can add `-s` when running `git commit`.

I'll look into the licensing issue.

packages/react/src/provider/use-context-mutator.ts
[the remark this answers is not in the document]
wichopy: test code here. let me know if it would be good to have a test for the global context case as well

https://github.com/open-feature/js-sdk/pull/1031/commits/a7090f099cce3be28565de935cd2119bfeb41372

packages/react/test/provider.spec.tsx
```diff
@@ -136,5 +139,74 @@ describe('OpenFeatureProvider', () => {
         await waitFor(() => expect(screen.queryByText('👍')).toBeInTheDocument(), { timeout: DELAY * 2 });
       });
     });
+
+    describe('useMutateContext', () => {
```
beeme1mr, line 143: These tests look great but could you please add one more for the following setup?

```
<OpenFeatureProvider domain={DOMAIN1}>
  <React.Suspense fallback={<div>{FALLBACK}</div>}>
    <TestComponent name="Will" />
    <OpenFeatureProvider>
      <React.Suspense fallback={<div>{FALLBACK}</div>}>
        <TestComponent name="Todd" />
      </React.Suspense>
    </OpenFeatureProvider>
  </React.Suspense>
</OpenFeatureProvider>
```

The nested component should update the global context.

packages/react/src/provider/use-context-mutator.ts
beeme1mr: Yep, I think it makes sense to add later. This is already a great start. I just wonder if we should name this function "setContext" instead of "mutateContext". @toddbaert do you have an opinion?

packages/react/test/provider.spec.tsx
wichopy: @beeme1mr I have set this up. In this scenario, it will update both TestComponents. Is that the expected behaviour?

beeme1mr: I'd like a test that ensures that updating context in the nested component updates the global context and not the domain scoped context. The other component probably isn't necessary in the test.

packages/react/src/provider/use-context-mutator.ts
```diff
@@ -0,0 +1,26 @@
...
+      return;
+    }
+
+    OpenFeature.setContext(domain, updatedContext);
+  }
+
+  return {
+    mutateContext,
+  };
+}
```
lukas-reining, line 26:
```suggestion
export function useContextMutator() {
    const { domain } = useContext(Context) || {};
    const previousContext = useRef(null);

    const mutateContext = useCallback(async (updatedContext: EvaluationContext) => {
        if (previousContext.current !== updatedContext) {
            if (!domain) {
                OpenFeature.setContext(updatedContext);
            } else {
                OpenFeature.setContext(domain, updatedContext);
            }
            previousContext.current = updatedContext;
        }
    }, [domain]);

    return {
        mutateContext,
    };
}
```

lukas-reining, line 26:
```suggestion
export function useContextMutator() {
    const { domain } = useContext(Context) || {};
    const previousContext = useRef(null);

    const mutateContext = useCallback(async (updatedContext: EvaluationContext) => {
        if (previousContext.current !== updatedContext) {
            if (!domain) {
                OpenFeature.setContext(updatedContext);
            } else {
                OpenFeature.setContext(domain, updatedContext);
            }
            previousContext.current = updatedContext;
        }
    }, [domain]);

    return {
        mutateContext,
    };
}
```

I would like to consider two things here. 

First is to memoize the `mutateContext` function here to avoid re-rendering of dependent components if they pass is downwards. This can just prevent some unnecessary re-renders.

Second and more important is that calling `setContext` can be quite expensive. The reason is outlined [here](https://openfeature.dev/blog/reconciling-with-state/#context-reconciliation) and specified [here](https://openfeature.dev/specification/sections/providers#26-provider-context-reconciliation), but so summarize, in the `static context paradigm` which the Web SDK and so React SDK use, setting the context can lead to a provider refreshing it's cached flags which most times means calling APIs.  
To prevent this being called unnecessarily, we should at least check if the new context is the same we had before. 
In my suggestion I am only checking for referential equality, maybe can deep equal this (cc @beeme1mr @toddbaert ), but I could imagine referential equality being good enough for this problem.

lukas-reining (commented):
Looks good, thank you! :) 
I left one comment to be considered.

packages/react/test/provider.spec.tsx
wichopy: The reason I'm asking is it seems like something is wrong, shouldn't domain scoped context be unchanged even if the global context changes?

packages/react/src/provider/use-context-mutator.ts
wichopy: @lukas-reining I had memoized callback ready for my next commit :) As for checking the object, I think it makes sense and can be added, but probably have to do a deep comparison as we can't assume that the developer will be using the same object reference when they call `setContext`. I found in react-land people like to copy objects everywhere.

beeme1mr: I feel like that logic would be better in the web sdk and could be handled in a separate pr.

lukas-reining: For the deep equality I am with you @beeme1mr. Checking for referential equality should be done directly as even successive renders would lead to reconciliation otherwise.

wichopy: Got it, deep equal in the core lib, and a simple reference check in react to prevent some rerenders.

packages/react/test/provider.spec.tsx
beeme1mr: Yes, you're correct. Is there something wrong with my example?

wichopy: Yes when I set the globally scoped context thats nested under the scoped one, I was expecting "Will says hi" to stay unchanged, but instead I'm getting both "Todd says aloha" and "Will says aloha".

Maybe its out of scope of this PR to narrow down if its the react implementation or the core lib with the issue, but wanted to confirm it was unintended behaviour.

beeme1mr: Ah, okay. I understand the issue. Basically, domain-scoped context is only managed separately if it's explicitly set for the domain. Otherwise, it inherits from the global context. The logic is that most of the time the context should be the same but we wanted to give users the ability to override this behavior if necessary.

I'm now wondering how we should handle this in the hooks. I think I'll need to test it out myself tomorrow before I can provide a solid recommendation.

Thanks for your effort on this.

packages/react/src/provider/use-context-mutator.ts
```diff
@@ -0,0 +1,28 @@
+import { useCallback, useContext, useRef } from 'react';
+import { OpenFeature, EvaluationContext } from '@openfeature/web-sdk';
+import { Context } from './context';
+
+/**
+ *
+ * A hook for accessing context mutating functions.
+ *
+ */
+export function useContextMutator() {
```
toddbaert, line 10: I wonder if we should add an optional param here for using the scoped OR global evaluation context.

By default, we could use the context indicated by the associated `domain`, but if desired, the user could indicate they want to mutate the global evaluation context.

packages/react/test/provider.spec.tsx
toddbaert: See: https://github.com/open-feature/js-sdk/pull/1031/files#r1801753490

packages/react/src/provider/use-context-mutator.ts
beeme1mr: Yes, I think that's a good idea, given how context works with domains.

@wichopy, [here's](https://openfeature.dev/docs/reference/technologies/client/web/#manage-evaluation-context-for-domains) a quick overview of how context works with domains.

wichopy: done!

packages/react/test/provider.spec.tsx
wichopy: @toddbaert adding a property to set global or not doesn't solve the issue I brought up though.

I've added a test with the global context having a different flag key and variants than the domain scoped one. The domain and global provider context evaluators both listen to `user === 'bob@flags.com'`, but now only the global context flag variant changes instead of both global and domain scoped.

When the providers are the same (same flag keys and variants) they both update when global context is nested under domain context. Did I find a bug or is this desired behaviour?

toddbaert: I'm not 100% sure I understand your last comment. But I can confirm what @beeme1mr said here:

>  Basically, domain-scoped context is only managed separately if it's explicitly set for the domain. 

Basically, if you don't set a context for domain X, it will use the default context. If you DO set a context for domain X, it uses that context (this is analogous to how providers are bound to domains as well. Does that clear things up?

My point about the option was somewhat separate: just that users might want to update either of these from their component, and either is valid.

wichopy: i think you cleared it up! i think testing is all set then unless we need more cases.

beeme1mr (approved):

packages/react/test/provider.spec.tsx
toddbaert: Yes, looks that way to me.

PR #1045, opened 2024-10-17
chore: add js docs for context mutator hook

toddbaert:
Minor refactors, but mostly adding js-doc after https://github.com/open-feature/js-sdk/pull/1031.


packages/react/src/context/use-context-mutator.ts
```diff
@@ -0,0 +1,46 @@
+import { useCallback, useContext, useRef } from 'react';
```
toddbaert, line 1: This file was moved into a new dir.

The only changes are additional js-docs and I changed the `setGlobal` option to `default` to keep in line with our other terminology (we call this the default context not the global context when we are talking about domains)

packages/react/src/context/use-context-mutator.ts
```diff
@@ -0,0 +1,46 @@
...
+import type { EvaluationContext } from '@openfeature/web-sdk';
+import { OpenFeature } from '@openfeature/web-sdk';
+import { Context } from '../common';
+
+export type ContextMutationOptions = {
+  /**
+   * Apply changes to the default context instead of the domain scoped context applied at the <OpenFeatureProvider/>.
+   * Note, if the <OpenFeatureProvider/> has no domain specified, the default is used.
+   */
+  default?: boolean;
```
toddbaert, line 11: changed from `setGlobal` to `default`.

beeme1mr: Can you come up with a more descriptive name than `default`?

toddbaert: I've changed to `defaultContext`.

beeme1mr (approved):

packages/react/src/context/use-context-mutator.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Context } from '../common';
+
+export type ContextMutationOptions = {
+  /**
+   * Mutate the default context instead of the domain scoped context applied at the `<OpenFeatureProvider/>`.
+   * Note, if the `<OpenFeatureProvider/>` has no domain specified, the default is used.
+   * See the {@link https://openfeature.dev/docs/reference/technologies/client/web/#manage-evaluation-context-for-domains|documentation} for more information.
+   * @default false
+   */
+  defaultContext?: boolean;
```
toddbaert, line 13: @wichopy @lukas-reining I've changed this from `setGlobal` to `defaultContext` for 2 reasons:

1 - I eventually want to return 2 functions from this hook, 1 for `set` and one for `merge`, so "set" seemed wrong
2 - we call this the `default` context mostly in reference to other domain-scoped context

packages/react/src/context/use-context-mutator.ts
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * A function to set the desired context (see: {@link ContextMutationOptions} for details).
+   * @param updatedContext
+   * @returns
+   */
+  setContext: (updatedContext: EvaluationContext) => Promise<void>
+};
+
+/**
+ * Get function(s) for mutating the evaluation context associated with this domain, or the default context if `global: true`.
```
wichopy, line 28: this `global: true` bit will need to line up with your finalize name in `ContextMutationOptions`

packages/react/src/context/use-context-mutator.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export type ContextMutation = {
+  /**
+   * A function to set the desired context (see: {@link ContextMutationOptions} for details).
+   * There's generally no need to await the result of this function; flag evaluation hooks will re-render when the context is updated.
+   * This promise never rejects.
+   * @param updatedContext
+   * @returns Promise for awaiting the context update
+   */
+  setContext: (updatedContext: EvaluationContext) => Promise<void>;
```
toddbaert, line 24: @wichopy @lukas-reining one thing perhaps worth talking about before we release is that we could make this return `void` instead of `Promise<void>`... it would give developers less power to "do something" after the context update is complete, but generally, developers should need to await this since things re-render if the context is changed.

IMO we should keep the promise exposed, but I wanted to point it out.

packages/react/src/context/use-context-mutator.ts
toddbaert: Nice catch!

toddbaert: fixed

toddbaert:
I'm going to merge this since it's clashing with some other work I'm doing, but please feel free to open new issues or make further comments before we release.

PR #1052, opened 2024-10-21
fix(angular): fix race condition on initialization

lukas-reining:
## This PR

Fixes a race condition when the directive is first rendered.
Initially the initialization code in the constructor before the `@Input`s were bound to the variables. 
This meant that in some cases providers were called with `undefined` as flag key, which most providers accept, but for flipt-web the engine paniced due to the value being `unit` instead of a string.

### DIFF
--- a/packages/angular/projects/angular-sdk/src/lib/feature-flag.directive.ts
+++ b/packages/angular/projects/angular-sdk/src/lib/feature-flag.directive.ts
@@ -5,6 +5,7 @@ import {
   Input,
   OnChanges,
   OnDestroy,
+  OnInit,
   TemplateRef,
   ViewContainerRef,
 } from '@angular/core';
@@ -33,7 +34,7 @@ class FeatureFlagDirectiveContext<T extends FlagValue> {
   standalone: true,
   selector: '[featureFlag]',
 })
-export abstract class FeatureFlagDirective<T extends FlagValue> implements OnDestroy, OnChanges {
+export abstract class FeatureFlagDirective<T extends FlagValue> implements OnInit, OnDestroy, OnChanges {
   protected _featureFlagDefault: T;
   protected _featureFlagDomain: string | undefined;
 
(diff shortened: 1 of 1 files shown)


packages/angular/projects/angular-sdk/src/lib/feature-flag.directive.ts
```diff
@@ -69,14 +70,17 @@ export abstract class FeatureFlagDirective<T extends FlagValue> implements OnDes
     templateRef: TemplateRef<FeatureFlagDirectiveContext<T>>,
   ) {
     this._thenTemplateRef = templateRef;
-    this.initClient();
   }
 
   set featureFlagDomain(domain: string | undefined) {
     this._featureFlagDomain = domain;
     this.initClient();
```
beeme1mr, line 77: Can this be removed now?

toddbaert:
Not a blocker for me, but if you can think of a quick test for this it might be good.

toddbaert (approved):

beeme1mr (approved):

packages/angular/projects/angular-sdk/src/lib/feature-flag.directive.ts
lukas-reining: No it sadly can not. 
If we remove this, angular will not see the changes in the domain and so it will not use the new one.

beeme1mr: I'm not sure I follow. Perhaps it's worth adding a comment so that future us won't accidentally remove it 😄

lukas-reining: Added a comment @beeme1mr.

packages/angular/projects/angular-sdk/src/lib/feature-flag.directive.ts
```diff
@@ -69,14 +70,21 @@ export abstract class FeatureFlagDirective<T extends FlagValue> implements OnDes
...
   ) {
     this._thenTemplateRef = templateRef;
-    this.initClient();
   }
 
   set featureFlagDomain(domain: string | undefined) {
+    /**
+     * We have to handle the change of the domain explicitly because we need to get a new client when the domain changes.
+     * This can not be done if we simply relay the onChanges method.
+     */
```
lukas-reining, line 79: @beeme1mr does that work for you?

PR #1063, opened 2024-10-28
feat(angular): add Angular 18 support

lukas-reining:
## This PR

- allows Angular 18 peer dependency


packages/angular/projects/angular-sdk/package.json
```diff
@@ -8,8 +8,8 @@
     "prepack": "shx cp ./../../../../LICENSE ./LICENSE"
   },
   "peerDependencies": {
-    "@angular/common": "^16.2.12 || ^17.3.0",
-    "@angular/core": "^16.2.12 || ^17.3.0",
+    "@angular/common": "^16.2.12 || ^17.3.0 || ^18.2.9",
```
beeme1mr, line 11: Could this be `^18.0.0`?

packages/angular/package.json
```diff
@@ -14,33 +14,30 @@
...
     "jest-preset-angular": "^14.2.2",
-    "ng-packagr": "^17.3.0",
+    "ng-packagr": "^18.2.1",
     "rxjs": "~7.8.0",
     "tslib": "^2.3.0",
-    "zone.js": "~0.14.3",
-    "typescript": "^4.7.4"
-  },
-  "overrides": {
-    "typescript": "^4.7.4"
```
beeme1mr, line 44: Why is this no longer needed?

beeme1mr (approved):
Can you please change the PR title to something like "add Angular 18 support"?

packages/angular/package.json
lukas-reining: Because we have the override already in the root `package.json`.

packages/angular/projects/angular-sdk/package.json
lukas-reining: If I am not mistaken, this should not matter right?
Or is it the minimum version still with this?

lukas-reining: Yes, just found out too and committed a change :D

PR #1074, opened 2024-11-01
perf: avoid using exceptions for flow control

beeme1mr:
## This PR

- avoids using error codes as flow control during an evaluation

### DIFF
--- a/packages/server/src/client/internal/open-feature-client.ts
+++ b/packages/server/src/client/internal/open-feature-client.ts
@@ -294,26 +294,17 @@ export class OpenFeatureClient implements Client {
       };
 
       if (evaluationDetails.errorCode) {
-        throw instantiateErrorByErrorCode(evaluationDetails.errorCode);
+        const err = instantiateErrorByErrorCode(evaluationDetails.errorCode);
+        await this.errorHooks(allHooksReversed, hookContext, err, options);
+        return this.getErrorEvaluationDetails(flagKey, defaultValue, err);
       }
 
       await this.afterHooks(allHooksReversed, hookContext, evaluationDetails, options);
 
       return evaluationDetails;
     } catch (err: unknown) {
-      const errorMessage: string = (err as Error)?.message;
-      const errorCode: ErrorCode = (err as OpenFeatureError)?.code || ErrorCode.GENERAL;
-
       await this.errorHooks(allHooksReversed, hookContext, err, options);
-
-      return {
-        errorCode,
-        errorMessage,
-        value: defaultValue,
-        reason: StandardResolutionReasons.ERROR,
-        flagMetadata: Object.freeze({}),
-        flagKey,
-      };
+      return this.getErrorEvaluationDetails(flagKey, defaultValue, err);
     } finally {
       await this.finallyHooks(allHooksReversed, hookContext, options);
     }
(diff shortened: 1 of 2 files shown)


aepfli (approved):

lukas-reining (approved):
Nice change! Even though this is more a "stylistic" change than a performance benefit.

beeme1mr:
> Nice change! Even though this is more a "stylistic" change than a performance benefit

It should have some minor performance improvements but I need to run proper performance tests to prove it. 😄

Here's a fun snippet from Stack Overflow to see how impactful throwing an error is on performance.

```js
import { performance } from "perf_hooks";

const error = new Error("invalid exception");
function ThrowException() {
  throw error;
}
const _ATTEMPT = 1000000;
function ThrowingExceptions() {
  const p1 = performance.now();
  for (let i = 0; i < _ATTEMPT; i++) {
    try {
      ThrowException();
    } catch (ex) {
      // log error
    }
  }
  const p2 = performance.now();
  console.log(`ThrowingExceptions: ${p2 - p1}`);
}
function ReturnException() {
  return { error };
}
function ReturningExceptions() {
  const p1 = performance.now();
  for (let i = 0; i < _ATTEMPT; i++) {
    const ex = ReturnException();
    if (ex.error) {
      // log error
    }
  }
  const p2 = performance.now();
  console.log(`ReturningExceptions: ${p2 - p1}`);
}

function Process() {
  ThrowingExceptions();
  ReturningExceptions();
  ThrowingExceptions();
  ReturningExceptions();
}

Process();
```
And here are the results on my computer.

ThrowingExceptions: 365.008125
ReturningExceptions: 3.436661000000015
ThrowingExceptions: 363.11218099999996
ReturningExceptions: 1.329001000000062

lukas-reining:
> It should have some minor performance improvements but I need to run proper performance tests to prove it. 😄
>
>Here's a fun snippet from Stack Overflow to see how impactful throwing an error is on performance.

Oh this is interesting @beeme1mr. I thought for V8 this would not affect the performance, had it in mind from some discussion I had in the past.
I do not think that this will affect our performance but I am also getting a little less than 100x performance for the test :D 

````
ThrowingExceptions: 183.6151249408722
ReturningExceptions: 2.1167919635772705
ThrowingExceptions: 196.92433404922485
ReturningExceptions: 2.2545419931411743
````

When I inline the error creation, I consistently get faster times with the throw (and sure long times due to Error creation overhead):

```
ThrowingExceptions: 1596.49795794487
ReturningExceptions: 1687.1295830011368
ThrowingExceptions: 1605.3998750448227
ReturningExceptions: 1646.5373330116272
```

```js
import { performance } from 'perf_hooks';

const _ATTEMPT = 1000000;

function ThrowingExceptions() {
  const p1 = performance.now();
  for (let i = 0; i < _ATTEMPT; i++) {
    try {
      throw new Error('invalid exception');
    } catch (ex) {
      // log error
    }
  }
  const p2 = performance.now();
  console.log(`ThrowingExceptions: ${p2 - p1}`);
}

function ReturnException() {
  return new Error('invalid exception');
}

function ReturningExceptions() {
  const p1 = performance.now();
  for (let i = 0; i < _ATTEMPT; i++) {
    const ex = ReturnException();
    if (ex) {
      // log error
    }
  }
  const p2 = performance.now();
  console.log(`ReturningExceptions: ${p2 - p1}`);
}

function Process() {
  ThrowingExceptions();
  ReturningExceptions();
  ThrowingExceptions();
  ReturningExceptions();
}

Process();
```

beeme1mr:
What version of node are you using? Perhaps performance has improved in recent versions. 🤔

lukas-reining:
> What version of node are you using? Perhaps performance has improved in recent versions. 🤔

Tried Node 16, 20 and 22, all the same, and my Chrome so several V8 versions, I guess it simply is notable overhead when testing it in the synthetic benchmark. I guess in reality the error creation alone compensates the performance benefit, but as we are not creating errors here anymore I guess we can actually have a little performance improvement :)

PR #1077, opened 2024-11-06
fix: update OpenFeature core version to 1.5.0

beeme1mr:
## This PR

- bumps OpenFeature core version to match the current release

### Notes

This should have been automatically by the CI but it didn't work for some reason. I'll look into why after the issue is resolved.

https://github.com/open-feature/js-sdk/commit/fe3ad8eeb9219ff08ba287cab228016da0b88e88#diff-e35631c960979816b4b2bda7950788e968930fcaf2cf39b482ff23117cd13888R50

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -47,7 +47,7 @@
   },
   "homepage": "https://github.com/open-feature/js-sdk#readme",
   "peerDependencies": {
-    "@openfeature/core": "1.4.0"
+    "@openfeature/core": "1.5.0"
   },
   "devDependencies": {
     "@openfeature/core": "1.5.0"


beeme1mr:
The script to bump the peer version is very inconsistent but I'm not sure why. The web SDK fairly constantly has the dev dependency version bumped but not the peer version. Running the script again properly bumps the peer. If I update the root package.json to update web before server, the server peer version isn't bumped. It must be some very odd race condition. Any ideas @toddbaert @lukas-reining?

beeme1mr:
We may want to loosen the peer dependency requirements if possible. It's a bit painful to run both the server and web SDK in a single project due to constant version mismatches.

toddbaert (approved):

toddbaert:
> We may want to loosen the peer dependency requirements if possible. It's a bit painful to run both the server and web SDK in a single project due to constant version mismatches.

It's not possible without introducing significant risk.

PR #1082, opened 2024-11-10
export use open feature client status with suspense support

wichopy:
## This PR

- adds this new feature

### Related Issues

Fixes #1234523


packages/react/src/provider/use-open-feature-client-status.ts
```diff
@@ -28,6 +32,12 @@ export function useOpenFeatureClientStatus(): ProviderStatus {
       client.removeHandler(ProviderEvents.Reconciling, updateStatus);
     };
   }, [client]);
+  // highest priority > evaluation hook options > provider options > default options > lowest priority
+  const defaultedOptions = { ...DEFAULT_OPTIONS, ...useProviderOptions(), ...normalizeOptions(options) };
+
+  if (defaultedOptions.suspendUntilReady && status === ProviderStatus.NOT_READY) {
+    suspendUntilReady(client);
```
wichopy, line 39: Let me know if I'm doing this right @beeme1mr, I assume whenever the provider needs to do some asynciness, the status is NOT_READY.

packages/react/src/provider/use-open-feature-client-status.ts
```diff
@@ -1,15 +1,20 @@
...
+
+type Options = Pick<ReactFlagEvaluationOptions, 'suspendUntilReady'>
 
 /**
  * Get the {@link ProviderStatus} for the OpenFeatureClient.
+ * @param options
  * @returns {ProviderStatus} status of the client for this scope
  */
-export function useOpenFeatureClientStatus(): ProviderStatus {
+export function useOpenFeatureClientStatus(options?: Options) {
```
wichopy, line 15: the great debate of return types 😅

https://www.youtube.com/watch?v=nwSe95uFN8E&t=47s

packages/react/src/provider/use-open-feature-client-status.ts
```diff
@@ -28,6 +33,12 @@ export function useOpenFeatureClientStatus(): ProviderStatus {
       client.removeHandler(ProviderEvents.Reconciling, updateStatus);
     };
   }, [client]);
+  // highest priority > evaluation hook options > provider options > default options > lowest priority
+  const defaultedOptions = { ...DEFAULT_OPTIONS, ...useProviderOptions(), ...normalizeOptions(options) };
+
+  if (defaultedOptions.suspendUntilReady && status === ProviderStatus.NOT_READY) {
+    suspendUntilReady(client);
```
toddbaert, line 40: Some of the code here is identical to the `useWhenProviderReady` hook - we might want to commonize it into a util.

packages/react/src/provider/use-open-feature-client-status.ts
toddbaert: One thing to consider; we might also want to support suspense for this hook when the provider status is `RECONCILING`, since it's use-case more broad than `useWhenProviderReady`... Which would mean checking for that state too, and adding the related option(s).

@beeme1mr @aepfli @lukas-reining Any thoughts on this?

packages/react/src/provider/use-open-feature-client-status.ts
toddbaert, line 15: We call this hook internally from `useWhenProviderReady`; we should probably explicitly disable suspense in the call from that hook so we don't suspend when we don't want to (since that hook also has it's own suspense logic).

toddbaert:
@wichopy Sorry for the delay - KubeCon really slowed things down for us, still playing some catch-up!

I still think this is valuable, and I think you're on the right track. I left some thoughts.

We also need some tests (probably very similar to the other tests involving hooks and syspense).

packages/react/src/provider/use-open-feature-client-status.ts
guidobrei: I'd rather keep this hook clean without any suspense support. 

If we want to have suspense support we might want to extend the useWhenProviderReady hook to also return the current state additionally to the boolean.

toddbaert: Ya, @wichopy I think I am starting to agree with @guidobrei on this one. I think I was wrong to attempt to add suspense here - I don't think it makes as much sense in this use-case - I think we should just remove the suspense and export the hook basically as is (though I think you've have too absorb some changes from main)

Sorry about the back-and-forth.

wichopy: done, please PTAL when you get another chance

packages/react/src/provider/use-open-feature-client-status.ts
```diff
@@ -1,15 +1,15 @@
 import { useEffect, useState } from 'react';
 import { useOpenFeatureClient } from './use-open-feature-client';
-import type { ProviderStatus } from '@openfeature/web-sdk';
-import { ProviderEvents } from '@openfeature/web-sdk';
```
beeme1mr, line 4: Is there a reason to combine these two imports? I personally like to keep types and runtime imports separate.

packages/react/src/provider/use-open-feature-client-status.ts
```diff
@@ -1,15 +1,15 @@
...
+import { ProviderEvents, type ProviderStatus, } from '@openfeature/web-sdk';
+
 
 /**
  * Get the {@link ProviderStatus} for the OpenFeatureClient.
  * @returns {ProviderStatus} status of the client for this scope
  */
-export function useOpenFeatureClientStatus(): ProviderStatus {
+export function useOpenFeatureClientStatus() {
```
beeme1mr, line 10: I see [your comment](https://github.com/open-feature/js-sdk/pull/1082#discussion_r1835698885) about removing the return type, but I think it's worth keeping. Explicitly defining the return type ensures you don't accidentally return an unexpected value. In this case, we want to ensure we only return a `ProviderStatus` and not a generic `string` type.

beeme1mr (approved):

wichopy:
@toddbaert shall we merge?

PR #1084, opened 2024-11-21
feat(angular): add angular 19 support

### LINKED ISSUE #1083: [FEATURE] angular-sdk: Add Angular v19 support
### Requirements

At present, attempting to install the Angular SDK in an Angular v19 project results in an error because the SDK is dependent on older versions of Angular.

lukas-reining:
## This PR

-  adds Angular 19 as possible peer dependency

fixes #1083


beeme1mr (approved):
I've looked into the license issues and added exceptions for all of them. They were transitive dependencies used in the Angular CLI.

PR #1087, opened 2024-11-21
feat: add evaluation details to finally hook

beeme1mr:
## This PR

- adds evaluation details to the `finally` stage in hooks.

### Notes

This breaks the signature of the `finally` stages based on [this spec enhancement](https://github.com/open-feature/spec/pull/280). It is **not** considered a breaking change to the SDK because hooks are marked as experimental in the spec, and the change has no impact on known hooks.

The noteworthy change to the interface is:

```diff
- finally?(hookContext: Readonly<HookContext<T>>, hookHints?: HookHints): HooksReturn;
+ finally?(hookContext: Readonly<HookContext<T>>, evaluationDetails: EvaluationDetails<T>, hookHints?: HookHints): HooksReturn;
```

### Follow-up Tasks

- Update the JS contribs repo


packages/server/src/client/internal/open-feature-client.ts
```diff
@@ -279,35 +281,38 @@ export class OpenFeatureClient implements Client {
       logger: this._logger,
     };
 
-    try {
-      const frozenContext = await this.beforeHooks(allHooks, hookContext, options);
+    const evaluationDetails = await(async () => {
```
toddbaert, line 284: I'm totally OK with the logic here, but I have a very slight concern that this could be creating a large amount of memory churn because we will be creating this anon function (and needing to clean it up) at every flag evaluation.

toddbaert: If you added this IIFE just to avoid a mutable var, I would just add the mutable var.

lukas-reining: I also think it might be more simple to just accept the mutation here. Additionally as @toddbaert say we can try to save some memory allocs here, even though it might be okay.

lukas-reining (approved):
The change looks good to me!
I am with @toddbaert regarding the IIFEs, and I could live with duplicated lines or mutations.

toddbaert (approved):

packages/server/src/client/internal/open-feature-client.ts
beeme1mr: @lukas-reining, could you please quickly re-review to see if you prefer this change? Thanks!

lukas-reining (approved):
Yes, looks good @beeme1mr :)

kinyoklion (approved):

PR #1090, opened 2024-11-21
ci: add id token write permissions

beeme1mr:
Adds missing permissions

https://docs.npmjs.com/generating-provenance-statements#publishing-packages-with-provenance-via-github-actions


.github/workflows/release-please.yml
```diff
@@ -54,6 +54,9 @@ jobs:
     needs: release-please
     runs-on: ubuntu-latest
     if: ${{ needs.release-please.outputs.release_created }}
+    permissions:
```
beeme1mr, line 57: I believe these are the only permissions we'll need, but I didn't have an easy way to confirm.

PR #1095, opened 2024-12-04
feat: re-render if flagsChanged is falsy

toddbaert:
Adds an improvement to the React SDK which supports re-renders if the [flags changed](https://open-feature.github.io/js-sdk/types/_openfeature_server_sdk.ConfigChangeEvent.html) array from a provider event is falsy.

Since some providers have no knowledge of flags which are changed, this allows them to support dynamic re-rendering by not defining this property. If the prop is null/undefined, we diff all flags... If the property is explicitly set to an empty array, that means no flags have changed and the React SDK skips all diff checks.

### DIFF
--- a/packages/react/src/evaluation/use-feature-flag.ts
+++ b/packages/react/src/evaluation/use-feature-flag.ts
@@ -267,7 +267,8 @@ export function useObjectFlagDetails<T extends JsonValue = JsonValue>(
 
 // determines if a flag should be re-evaluated based on a list of changed flags
 function shouldEvaluateFlag(flagKey: string, flagsChanged?: string[]): boolean {
-  return !!flagsChanged && flagsChanged.includes(flagKey);
+  // if flagsChange is missing entirely, we don't know what to re-render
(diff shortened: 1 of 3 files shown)


packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -267,7 +267,8 @@ export function useObjectFlagDetails<T extends JsonValue = JsonValue>(
 
 // determines if a flag should be re-evaluated based on a list of changed flags
 function shouldEvaluateFlag(flagKey: string, flagsChanged?: string[]): boolean {
-  return !!flagsChanged && flagsChanged.includes(flagKey);
+  // if flagsChange is missing entirely, we don't know what to re-render
+  return !flagsChanged || flagsChanged.includes(flagKey);
```
toddbaert, line 271: This is the change.

beeme1mr (approved):
Thanks! I think this is a good change.

lukas-reining (approved):

PR #1100, opened 2024-12-09
refactor: improve track interface for providers

beeme1mr:
## This PR

- updates the context and trackingEventDetails on the tracking interface to not be optional since they're always supplied by the SDK.

### Notes

This is a small QoL improvement for provider devs implementing the tracking interface.

### How to test

The SDK still compiles, and the tests don't need to be modified.

### DIFF
--- a/packages/server/test/client.spec.ts
+++ b/packages/server/test/client.spec.ts
@@ -859,6 +859,18 @@ describe('OpenFeatureClient', () => {
         }).not.toThrow();
       });
 
+      it('provide empty tracking details to provider if not supplied in call', async () => {
+        await OpenFeature.setProviderAndWait({ ...MOCK_PROVIDER });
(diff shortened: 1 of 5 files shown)


packages/server/test/client.spec.ts
```diff
@@ -859,6 +859,18 @@ describe('OpenFeatureClient', () => {
         }).not.toThrow();
       });
 
+      it('provide empty tracking details to provider if not supplied in call', async () => {
```
toddbaert, line 862: This test fails before the change to default the tracking details.

toddbaert:
I pushed up some changes to actually default the tracking contents to match your types. I thought we were doing this, but we weren't. I also added a test.

toddbaert (approved):

lukas-reining (approved):

PR #1101, opened 2024-12-10
refactor: export public option types

beeme1mr:
## This PR

- renames the `shared` folder to `internal` to prevent accident exports
- export public option types

### Notes

This will make it easier for devs who need access to the option types.


toddbaert (approved):

toddbaert:
I'm just going to merge this since it's just renames and adding one export. I can't see any import cycles.

Issue #998, opened 2024-09-04
[FEATURE] Implement logging hook


toddbaert:
### Requirements

Implement a logging hook as described in appendix A of the OpenFeaure spec: https://github.com/open-feature/spec/blob/main/specification/appendix-a-included-utilities.md#logging-hook

Rahilsiddique:
So what I understand is that we have to create a hook that will be used to log the messages during flag life-cycle 
i.e before, after, error and finally 
Hey I would like to give it a try 🤚 
Thanks !

toddbaert:
> So what I understand is that we have to create a hook that will be used to log the messages during flag life-cycle i.e before, after, error and finally Hey I would like to give it a try 🤚 Thanks !

Yep - and also remove any non-debug logging in hot code-paths.

beeme1mr:
Hi @Rahilsiddique, is this something you're still interested in working on?

kevinmlong:
@beeme1mr / @toddbaert - I'm happy to take this issue on as my first issue.

beeme1mr:
Thanks @kevinmlong!

kevinmlong:
@beeme1mr / @toddbaert  - trying to understand a few things when working on this issue.

It appears that `DefaultLogger` doesn't actually log anything when `info` or `debug` are called. Looking at the linked implementation from the Java SDK, it use debug for `before` and `after`. Pretty easy to fill out the implementation, but then this causes the a few tests to fail. What's the reasoning here for why `console.info` and `console.debug` shouldn't be used for the respective implementations in `DefaultLogger`?

kevinmlong:
Looking at the spec (https://github.com/open-feature/spec/blob/main/specification/appendix-a-included-utilities.md#logging-hook) seems we should be using `debug` and `info` for these stages as well, so perhaps we should have the `DefaultLogger` implement these.

beeme1mr:
`console.debug` is an alias for `console.log` in [node.js](https://nodejs.org/api/console.html#consoledebugdata-args). That's why we decided to no-op debug by default. In this case, debug feels like the correct log level but it would definitely be annoying to see nothing if you're using the default logger. @toddbaert @lukas-reining, would either of you have a suggestion?

kevinmlong:
[...]

PR #1114, opened 2024-12-19
feat: logging hook

kevinmlong:
## This PR
This PR provides a LoggingHook implementation and the related tests. Additionally, this PR adds a constructor with two parameters to allow users to forcefully show the `info` and `debug` logging levels.

### Related Issues
Implements #998 

### How to test
Test harness now has a `logger-hook.spec.ts` that appropriately checks that the appropriate logging levels are called and provide the expected output.


packages/shared/src/hooks/logging-hook.ts
```diff
@@ -0,0 +1,70 @@
...
+const EVALUATION_CONTEXT_KEY = 'evaluation_context';
+const ERROR_CODE_KEY = 'error_code';
+const ERROR_MESSAGE_KEY = 'error_message';
+const REASON_KEY = 'reason';
+const VARIANT_KEY = 'variant';
+const VALUE_KEY = 'value';
+
+export class LoggingHook implements BaseHook {
+  readonly includeEvaluationContext: boolean = false;
+  readonly logger = new SafeLogger(new DefaultLogger(true, true));
```
beeme1mr, line 23: Please allow a custom logger to be passed in the constructor. If one isn't provided, we can fall back on what you have here.

packages/shared/src/hooks/logging-hook.ts
```diff
@@ -0,0 +1,70 @@
...
+    const payload: LoggerPayload = { stage: 'error' };
+
+    payload[ERROR_MESSAGE_KEY] = error.message;
+    payload[ERROR_CODE_KEY] = error.code;
+
+    this.addCommonProps(payload, hookContext);
+    this.logger.error(payload);
+  }
+
+  finally(hookContext: Readonly<HookContext<FlagValue>>, hookHints?: HookHints): void {
```
beeme1mr, line 56: Please make sure to rebase your PR. This method signature has changed recently.

https://github.com/open-feature/js-sdk/pull/1087

packages/shared/src/logger/default-logger.ts
```diff
@@ -3,6 +3,15 @@
 import type { Logger } from './logger';
 
 export class DefaultLogger implements Logger {
+  
+  private readonly showInfo : boolean = false;
+  private readonly showDebug : boolean = false;
+
+  constructor(showInfo: boolean = false, showDebug: boolean = false){
```
beeme1mr, line 10: Perhaps a log-level argument could work well. It could default to "warn" by default.

lukas-reining, line 10: https://github.com/open-feature/js-sdk/issues/998#issuecomment-2541526479

Please insider this @kevinmlong.
Having the level configurable via env vars is common and most useful when deployed. Having to rebuild your Dockerfile or even commit a changed log level in case you need to observe something in a deployed instance would not be too practical to me.

packages/shared/src/hooks/logging-hook.ts
kevinmlong: @beeme1mr This has been addressed with latest commit.

packages/shared/src/hooks/logging-hook.ts
kevinmlong: @beeme1mr This has been addressed with latest commit.

packages/shared/src/logger/default-logger.ts
kevinmlong: @beeme1mr This has been addressed with latest commit.

kevinmlong: @lukas-reining - I'm all for being able to pass in configuration from the environment. Perhaps I'm confused though. Wouldn't the applications making use of the hook and the DefaultLogger be where the parameters passed in come from the environment? Perhaps the switch to passing in a LogLevel parameter resolved this as the the client implementation of the SDK would pass in `process.env.LOG_LEVEL` where appropriate.

lukas-reining: Hey sorry for the delay @kevinmlong, I was on vacation :)

Mh, I think the problem to me here is that I could imagine the case of the `LoggingHook` being registered in every deployment, but it might be preferred to be able to turn on the logging on demand without rebuilding. 
If a user deployed a version without explicitly setting a logger, the hook would be statically configured to be in `LogLevel.DEBUG`. 

My idea is that the default OpenFeature logger could just always consume the `LOG_LEVEL` from the environment variables. With this, the "simple" usage of the hook would always be configurable from the env. 
The other approach would require the user to read the env variable, validate it, construct a new logger with it and then create the hook. I find this acceptable but "out-of-the-box configurability" would be nice here. 

Another option could be to add a `setLogLevel` method to the logger interface. 
The `LoggingHook` could then read, parse and validate the LOG_LEVEL from the environment variable and apply it to the configured `Logger`. Maybe this would be even better in terms of consistency as this would make every logger implementation accept the same env variable through the `LoggingHook`.

What do you think of these three options? @kevinmlong 

cc @toddbaert @beeme1mr

kevinmlong: Many apologies @lukas-reining / @toddbaert / @beeme1mr. It's been a crazy start to the year, and I'm finally finding some time to get back to this.

I'm good with a standard out-of-the-box variable that is read from the environment and defaults to a particular level if not set. I like the idea of doing this via a setLogLevel method on the Logger interface and will give that a whirl.

kevinmlong: @lukas-reining - As I started to noodle through this, it occurred to me that "pulling from the environment" would work great on server side use of this, but web apps (e.g., React, Vue) that don't use SSR, are essentially pre-bundled and the "environment" is pre-determined. I'm definitely a fan of config based deployments so that I don't have to re-build stuff to change things like this, but not sure what the best path forward is here. @toddbaert / @beeme1mr - could you way in here please?

lukas-reining: > web apps (e.g., React, Vue) that don't use SSR, are essentially pre-bundled and the "environment" is pre-determined

Yeah I did not consider them here. I think it is still fine to change the log behavior during build of the SPA/static web app as this is in-line with most other configs for them. 
But it strengthens the point that we need constructor and in my view also env config options. I could also accept leaving the env variable reading to the SDK users but I still support some default.

beeme1mr:
I'm going to close the PR for now since it's still a draft and has been open for a while. Feel free to reopen when it's ready for a review.

PR #1120, opened 2025-01-28
feat: add telemetry helper utils

### LINKED ISSUE #1118: Implement TelemetryHook in JS SDK
## Description

Implement a `TelemetryHook` that can be shared across the various JavaScript SDKs. Ensure it complies with the OpenTelemetry specification and meets the technical requirements, including those from [Appendix D](https://openfeature.dev/specification/appendix-d) of the OpenFeature Specification and the [OpenTelemetry Semantic Conventions for Feature Flags](https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/).

## Technical requirements

- Must comply with the OTel specification.
- Must be included in the SDK.
- Must not include any third-party dependencies.
- Must be extensible, allowing users to add their own capturing mechanisms and properties.
- Must adhere to the observability conformance requirements in Appendix D of the OpenFeature Specification.

beeme1mr:
## This PR

- adds a method to core that returns a semantically valid flag evaluation event

### Related Issues

Fixes #1118

## Notes

I've omitted value type because it is likely to be declined in the OTel spec and adds complexity. We should consider removing that section from [appendix d](https://openfeature.dev/specification/appendix-d#evaluation-details).


packages/shared/src/telemetry/attributes.ts
```diff
@@ -0,0 +1,78 @@
+/**
+ * The attributes of an OpenTelemetry compliant event for flag evaluation.
+ * @see https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/
+ */
+export const TELEMETRY_ATTRIBUTE = {
```
dyladan, line 5: I still think constants like `export const TELEMETRY_ATTRIBUTE_FEATURE_FLAG_KEY` are better for minification and tree shaking.

packages/shared/src/telemetry/attributes.ts
```diff
@@ -0,0 +1,78 @@
...
+   */
+  KEY: 'feature_flag.key',
+  /**
+   * Describes a class of error the operation ended with.
+   *
+   * - type: `string`
+   * - requirement level: `conditionally required`
+   * - example: `flag_not_found`
+   */
+  ERROR_CODE: 'error.type',
```
dyladan, line 21: one is code the other is type. is that intentional?

packages/shared/src/telemetry/evaluation-event.ts
```diff
@@ -0,0 +1,67 @@
+import { ErrorCode, StandardResolutionReasons, type EvaluationDetails, type FlagValue } from '../evaluation/evaluation';
+import type { HookContext } from '../hooks/hooks';
+import type { JsonValue } from '../types';
+import { TELEMETRY_ATTRIBUTE } from './attributes';
+import { TELEMETRY_EVALUATION_DATA } from './evaluation-data';
+import { TELEMETRY_FLAG_METADATA } from './flag-metadata';
+
+interface EvaluationEvent {
```
dyladan, line 8: I prefer to use `type` over `interface` when a type won't be implemented by a class.

packages/shared/src/telemetry/attributes.ts
```diff
@@ -0,0 +1,78 @@
...
+   * - type: `string`
+   * - requirement level: `required`
+   * - example: `logo-color`
+   */
+  KEY: 'feature_flag.key',
+  /**
+   * Describes a class of error the operation ended with.
+   *
+   * - type: `string`
+   * - requirement level: `conditionally required`
```
lukas-reining, line 18: I do not think it is definitely need but maybe for `conditionally required` we can add a sentence about the condition.
But I am fine if we just leave it out.

dyladan (commented):
Looks semantically correct to me. Left a couple comments about style but they're really just nits. The one where the names don't match (code/type) is the only one that I think is an actual mistake.

packages/shared/src/telemetry/evaluation-event.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @returns {EvaluationEvent} An evaluation event object containing the event name and attributes
+ */
+export function createEvaluationEvent(
+  hookContext: Readonly<HookContext<FlagValue>>,
+  evaluationDetails: EvaluationDetails<FlagValue>,
+): EvaluationEvent {
+  const attributes: EvaluationEvent['attributes'] = {
+    [TELEMETRY_ATTRIBUTE.KEY]: hookContext.flagKey,
+    [TELEMETRY_ATTRIBUTE.PROVIDER]: hookContext.providerMetadata.name,
+    [TELEMETRY_ATTRIBUTE.REASON]: (evaluationDetails.reason ?? StandardResolutionReasons.UNKNOWN).toLowerCase(),
```
lukas-reining, line 29: In the spec we say: 

> The error.type and feature_flag.evaluation.reason enumerations use a lowercase "snake_case" convention (see [OpenTelemetry feature-flag log records](https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/)). OpenFeature [error codes](https://openfeature.dev/specification/types#error-code) and [resolution reasons](https://openfeature.dev/specification/types#resolution-reason) should be transformed accordingly by integrations which include this data.

I do not find this explicitly, but if we want to be strict about this, `toLowerCase()` might not be enough as  `evaluationDetails.reason` is `keyof typeof StandardResolutionReasons | (string & Record<never, never>);` which includes any string.

packages/shared/src/telemetry/evaluation-event.ts
```diff
@@ -0,0 +1,67 @@
...
+    attributes[TELEMETRY_ATTRIBUTE.SET_ID] = setId;
+  }
+
+  const version = evaluationDetails.flagMetadata[TELEMETRY_FLAG_METADATA.VERSION];
+  if (version) {
+    attributes[TELEMETRY_ATTRIBUTE.VERSION] = version;
+  }
+
+  if (evaluationDetails.reason === StandardResolutionReasons.ERROR) {
+    attributes[TELEMETRY_ATTRIBUTE.ERROR_CODE] = (evaluationDetails.errorCode ?? ErrorCode.GENERAL).toLowerCase();
```
lukas-reining, line 56: Same question regarding the `toLowerCase()` as above

lukas-reining (approved):
Looks good to me!
Left one question.

packages/shared/src/telemetry/attributes.ts
dyladan: I'd include it. Makes it a lot easier if you don't have to go look the condition up elsewhere.

packages/shared/src/telemetry/evaluation-event.ts
dyladan: In otel we usually make it as easy as possible to do the right thing, but allow for customization in case you really need it for your case. In this case, providing an unknown value will produce _something_ so normalizing it with lowercase is probably fine.

kinyoklion (approved):

packages/shared/src/telemetry/attributes.ts
beeme1mr: I made the switch but some of the the names are so long that they don't fit in the intellisense dropdown.

packages/shared/src/telemetry/attributes.ts
beeme1mr: Yes it was intentional but I changed it anyways. The difference will now be in the hook instead of the attribute const.

packages/shared/src/telemetry/evaluation-event.ts
beeme1mr: i left this behavior unchanged.

packages/shared/src/telemetry/attributes.ts
dyladan: It's up to you. It's your library. This is just the style we use in otel for bundle size reasons.

beeme1mr: I reverted back after talking to @toddbaert. We agreed that the ergonomic improves outweigh the slight tree shaking penalty.

packages/shared/src/telemetry/flag-metadata.ts
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * The context identifier returned in the flag metadata uniquely identifies
+   * the subject of the flag evaluation. If not available, the targeting key
+   * should be used.
+   */
+  CONTEXT_ID: 'contextId',
+  /**
+   * 	A logical identifier for the flag set.
+   */
+  SET_ID: 'flagSetId',
```
toddbaert, line 15:
```suggestion
  FLAG_SET_ID: 'flagSetId',
```

toddbaert (approved):

PR #1134, opened 2025-02-06
fix: msg missing when providers return err resolutions

toddbaert:
Fixes an issue where the error message was lost if a provider returned a resolution which had an error code (instead of throwing).

Frankly I don't think testing this is worth it. I debating making the second arg to this function `message: string | undefined` instead of `message?: string` (which would force a second arg to be set) but that would be breaking and we'd need to release a core v2 to avoid possible tsc issues in consumers, which I don't think is worth it ether.


beeme1mr (approved):

beeme1mr:
Merging because it's low risk and fixes a bug.

PR #1151, opened 2025-02-19
feat: add support for abort controllers to event handlers

beeme1mr:
## This PR

- adds an optional abort signal option to the `addHandler` API both globally and on the client.
- updates the React useEffect hook to leverage the abort signal

### Notes

This change was inspired by [this blog](https://kettanaito.com/blog/dont-sleep-on-abort-controller). It allows us to use abort controllers to easily unregister event handlers. I've checked [MDN](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) to confirm that we shouldn't have any compatibility issues.

### Follow-up Tasks

We'll need to bump the min peer dependency versions when releasing to avoid issues.

### How to test

There should be pretty good test coverage of this. Please let me know if you spot any gaps.


lukas-reining (approved):
Nice change!
I really like how this makes it unnecessary to keep a reference to the handler!

PR #1152, opened 2025-02-26
feat: add a top-level method for accessing providers

beeme1mr:
## This PR

- add a top-level method for accessing providers

### Notes

While working on some improvements to the way Suspense works in the React SDK, I ran into a few scenarios were having access to the provider itself was important. I needed a way to confidently tell that a provider was the noop provider since it has special properties like never being in a ready state. There are a few ways could could achieve this but I noticed that the Java SDK [already has methods](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java#L279) that expose the provider. It also allowed me to improve some of our existing tests.

### How to test

Unit tests have been updated accordingly.

### DIFF
--- a/packages/server/src/open-feature.ts
+++ b/packages/server/src/open-feature.ts
@@ -138,6 +138,27 @@ export class OpenFeatureAPI
     return this;
   }
 
+  /**
+   * Get the default provider.
+   *
+   * Note that it isn't recommended to interact with the provider directly, but rather through
+   * an OpenFeature client.
+   * @returns {Provider} Default Provider
+   */
+  getProvider(): Provider;
+  /**
+   * Get the provider bound to the specified domain.
+   *
+   * Note that it isn't recommended to interact with the provider directly, but rather through
+   * an OpenFeature client.
+   * @param {string} domain An identifier which logically binds clients with providers
+   * @returns {Provider} Domain-scoped provider
+   */
+  getProvider(domain?: string): Provider;
+  getProvider(domain?: string): Provider {
+    return this.getProviderForClient(domain);
+  }
+
   setContext(context: EvaluationContext): this {
     this._context = context;
     return this;
--- a/packages/server/test/open-feature.spec.ts
+++ b/packages/server/test/open-feature.spec.ts
@@ -74,8 +74,8 @@ describe('OpenFeature', () => {
     it('should set the default provider if no domain is provided', () => {
       const provider = mockProvider();
       OpenFeature.setProvider(provider);
-      const client = OpenFeature.getClient();
-      expect(client.metadata.providerMetadata.name).toEqual(provider.metadata.name);
+      const registeredProvider = OpenFeature.getProvider();
+      expect(registeredProvider).toEqual(provider);
     });
 
     it('should not change providers associated with a domain when setting a new default provider', () => {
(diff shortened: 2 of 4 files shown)


packages/server/src/open-feature.ts
```diff
@@ -138,6 +138,27 @@ export class OpenFeatureAPI
...
+   * Get the provider bound to the specified domain.
+   *
+   * Note that it isn't recommended to interact with the provider directly, but rather through
+   * an OpenFeature client.
+   * @param {string} domain An identifier which logically binds clients with providers
+   * @returns {Provider} Domain-scoped provider
+   */
+  getProvider(domain?: string): Provider;
+  getProvider(domain?: string): Provider {
+    return this.getProviderForClient(domain);
```
beeme1mr, line 159: Unfortunately, this method is named poorly for historical reasons but I wanted to avoid breaking changes in core.

packages/server/test/open-feature.spec.ts
```diff
@@ -85,11 +85,11 @@ describe('OpenFeature', () => {
...
+      const defaultProvider = OpenFeature.getProvider();
+      const domainSpecificProvider = OpenFeature.getProvider(domain);
 
-      expect(defaultClient.metadata.providerMetadata.name).toEqual(provider.metadata.name);
-      expect(domainSpecificClient.metadata.providerMetadata.name).toEqual(fakeProvider.metadata.name);
+      expect(defaultProvider).toEqual(provider);
+      expect(domainSpecificProvider).toEqual(fakeProvider);
     });
 
     it('should bind a new provider to existing clients in a matching domain', () => {
```
beeme1mr, line 95: This test could also be improved if we add a method on the client to get the provider. I'm just not sure if it's a good idea.

toddbaert (approved):
We have this in Java and other SDKs, and since it's not on the client I think the chances of averaging application authors using it appropriately is relatively low. 

There's utility here enough to do it IMO.

packages/server/src/open-feature.ts
lukas-reining: Do we consider this breaking for core? 
I considered core to be free of the need to not break as long as the web and server SDKs are also fixed/not broken. 
I do not think core is considered a real library for 3rd parties, rather a way to share componets between the web and server SDK.

It is not important but I guess we could still "clean up" giving this a fitting name.

lukas-reining (approved):
Looks good to me!

packages/server/src/open-feature.ts
beeme1mr: Yeah, we could 100% make braking changes to core and simply bump the min peer version. The only time there could be issues is if users are managing the core version themselves. I could definitely make the change if we think it's worth it. WDYT @toddbaert @lukas-reining?

toddbaert: Yes we can release breaking versions of core, and in theory, if we are doing our jobs and correctly specifying the peer in web/server, and so are users, we won't have issues...

However, that's just in theory. Of course there will be users who have installed this dep incorrectly and will cause themselves problems, so for me, in situations like this the cost/benefit doesn't justify the change (we risk some bugs just for a method name change).

All that said, in theory it is totally fine, so if you guys want to keep things clean I can support that. I don't care if we are on core v16 next month - it makes little difference except that we have to update our few deps which directly use core.

lukas-reining: > However, that's just in theory. Of course there will be users who have installed this dep incorrectly and will cause themselves problems, so for me, in situations like this the cost/benefit doesn't justify the change (we risk some bugs just for a method name change).
>
>All that said, in theory it is totally fine, so if you guys want to keep things clean I can support that. I don't care if we are on core v16 next month - it makes little difference except that we have to update our few deps which directly use core.

As said, it is not super important, so I also would like to avoid breaking projects for no good reason.

> All that said, in theory it is totally fine, so if you guys want to keep things clean I can support that. I don't care if we are on core v16 next month - it makes little difference except that we have to update our few deps which directly use core.

Besides when `The only time there could be issues is if users are managing the core version themselves` is true, which I expect to happen in reality. 

I just wanted to know how we consider `core, but I would also lean toward avoiding breaking just for fixing the naming here I guess.

PR #1157, opened 2025-02-26
feat: add support for suspense flags in next.js

beeme1mr:
## This PR

- adds an internal `use` polyfill
- refactors suspense support to maintain state across rerenders
- adds support for building Next.JS that use suspending flags.

### Notes

Previously, the Next.JS build process would timeout when using a suspense flag hook. The reason for this is because the noop provider (which is used during a build) is never ready. That meant that the promise thrown to initiate suspense never resolved. To address this, I've added global state using a weak map that's keyed off a provider. A `useRef` won't help because the value is only retained if the component renders successfully.

### How to test

I've added some tests and manually tested in the Toggle Shop.

> [!NOTE]
>  Marking as a draft until #1152 is merged.


packages/react/src/internal/suspense.ts
```diff
@@ -1,21 +1,57 @@
...
+import { NOOP_PROVIDER, ProviderEvents } from '@openfeature/web-sdk';
+import { use } from './use';
+
+/**
+ * A weak map is used to store the global suspense status for each provider. It's
+ * important for this to be global to avoid rerender loops. Using useRef won't
+ * work because the value isn't preserved when a promise is thrown in a component,
+ * which is how suspense operates.
+ */
+const globalProviderSuspenseStatus = new WeakMap<Provider, Promise<unknown>>();
```
lukas-reining, line 11: I am not 100% sure but globals in Next.js code tend to create weird states. 
If this code is run on the server  `globalProviderSuspenseStatus` will be the same for every caller if I recall correctly.
This means when we save a provider here I think we could either get multiple instances of the provider, or the same provider with the same eval context on multiple clients. 

But I am not 100% sure if I am not overseeing anything here that avoids this.

This is what I am not 100% sure about if it applies here: https://github.com/vercel/next.js/discussions/47605

lukas-reining (commented):
This generally looks good to me!
I left one concern that I am not really sure about.

packages/react/src/internal/suspense.ts
beeme1mr: It's worth noting that Vercel's SWR library also maintains a global state using a weakmap.

https://github.com/vercel/swr/blob/975f99118fc11eedc79accceb6eeaead19753db9/src/_internal/utils/global-state.ts

In our case, the weakmap is useful for handling the noop provider and some minor initialization optimizations. I don't think it would be a big deal if we lose state between requests. Basically, it would behave how it current done.

lukas-reining: I think the problem would be more that we might confuse providers. 

From the Issue: 
> Since mutating server-side global variables does some combination of a) not updating when you expect them to b) resetting arbitrarily c) leaking memory and d) leaking user data,

E.g. looks at TanStack Query: https://tanstack.com/query/latest/docs/framework/react/guides/ssr?from=reactQueryV3#initial-setup
They say: 
> // NEVER DO THIS:
// const queryClient = new QueryClient()
//
// Creating the queryClient at the file root level makes the cache shared
// between all requests and means _all_ data gets passed to _all_ users.
// Besides being bad for performance, this also leaks any sensitive data.

lukas-reining: I imagine that when we have a static context provider in this list, this could become a problem here. 
Maybe we have that problem already between different users with our singleton API 🤔

packages/react/src/internal/suspense.ts
```diff
@@ -1,21 +1,57 @@
...
-  }).finally(() => {
-    client.removeHandler(ProviderEvents.Ready, resolve);
-    client.removeHandler(ProviderEvents.Ready, reject);
-  });
+export function suspendUntilInitialized(provider: Provider, client: Client) {
+  const statusPromiseRef = globalProviderSuspenseStatus.get(provider);
+  if (!statusPromiseRef) {
+    // Noop provider is never ready, so we resolve immediately
+    const statusPromise = provider !== NOOP_PROVIDER ? isProviderReady(client) : Promise.resolve();
+    // Storing the promise globally because
```
RedbackThomson, line 26: nit: This comment is unfinished

RedbackThomson (approved):
Don't see anything obviously violating here, but I haven't done the due diligence to run it locally yet

packages/react/src/internal/suspense.ts
beeme1mr: You may be right and perhaps I should update the PR title to not include Next.JS. This PR definitely addresses a build-time issue that's easily reproducible with Next. However, as you stated, we'll likely need to do more to properly support Next.

packages/react/src/internal/suspense.ts
beeme1mr, line 26:
```suggestion
```

packages/react/src/internal/suspense.ts
lukas-reining: > This PR definitely addresses a build-time issue that's easily reproducible with Next.

Totally! 

> You may be right and perhaps I should update the PR title to not include Next.JS
Removing Next from the title would make sense to me. 

The problem is that running Next with the React SDK currently can lead to security problems if I understand correctly.

lukas-reining (approved):

packages/react/src/internal/suspense.ts
toddbaert: I think we should remove the explicit mention of Next here. It seems like based on our other discussions we need that.

But as far as I know, this PR doesn't hurt anything that already exists, so it's probably OK to have as long as we don't say that it explicitly allows us to support Next formally.

lukas-reining: > But as far as I know, this PR doesn't hurt anything that already exists, so it's probably OK to have as long as we don't say that it explicitly allows us to support Next formally.

Yes, this is also what I meant :)

beeme1mr: I've updated the title and comment.

lukas-reining (approved):

PR #1159, opened 2025-03-14
feat: adds RequireFlagsEnabled decorator

kaushalkapasi:
## This PR
-  Feature: Adds a `RequireFlagsEnabled` decorator to allow a simple, reusable way to block access to a specific controller or endpoint based on the value of a list of one, or many, boolean flags

### Notes
- Discussions on the approach & implementation are welcome!

### Follow-up Tasks
- Update OpenFeature NestJS docs to include new `RequireFlagsEnabled` decorator & usage examples

### How to test
`npx jest --selectProject=nest`

### DIFF
--- a/packages/nest/src/require-flags-enabled.decorator.ts
+++ b/packages/nest/src/require-flags-enabled.decorator.ts
@@ -2,6 +2,7 @@ import type { CallHandler, ExecutionContext, HttpException, NestInterceptor } fr
 import { applyDecorators, mixin, NotFoundException, UseInterceptors } from '@nestjs/common';
 import { getClientForEvaluation } from './utils';
 import type { EvaluationContext } from '@openfeature/server-sdk';
+import type { ContextFactory } from './context-factory';
 
 type RequiredFlag = {
   flagKey: string;
@@ -37,6 +38,13 @@ interface RequireFlagsEnabledProps {
    * @see {@link OpenFeature#setContext}
    */
   context?: EvaluationContext;
+
+  /**
+   * A factory function for creating an OpenFeature {@link EvaluationContext} from Nest {@link ExecutionContext}.
+   * For example, this can be used to get header info from an HTTP request or information from a gRPC call to be used in the {@link EvaluationContext}.
+   * @see {@link ContextFactory}
+   */
+  contextFactory?: ContextFactory;
 }
 
 /**
@@ -56,6 +64,11 @@ interface RequireFlagsEnabledProps {
  *   context: {                             // Optional, defaults to the global OpenFeature Context
  *     targetingKey: 'user-id',
  *   },
+ *   contextFactory: (context: ExecutionContext) => {  // Optional, defaults to the global OpenFeature Context. Takes precedence over the context option.
+ *     return {
+ *       targetingKey: context.switchToHttp().getRequest().headers['x-user-id'],
+ *     };
+ *   },
  * })
  * @Get('/')
  * public async handleGetRequest()
@@ -72,7 +85,8 @@ const FlagsEnabledInterceptor = (props: RequireFlagsEnabledProps) => {
 
     async intercept(context: ExecutionContext, next: CallHandler) {
       const req = context.switchToHttp().getRequest();
-      const client = getClientForEvaluation(props.domain, props.context);
+      const evaluationContext = props.contextFactory ? await props.contextFactory(context) : props.context;
+      const client = getClientForEvaluation(props.domain, evaluationContext);
 
       for (const flag of props.flags) {
         const endpointAccessible = await client.getBooleanValue(flag.flagKey, flag.defaultValue ?? false);
--- a/packages/nest/src/require-flags-enabled.decorator.ts
+++ b/packages/nest/src/require-flags-enabled.decorator.ts
@@ -1,21 +1,28 @@
 import type { CallHandler, ExecutionContext, HttpException, NestInterceptor } from '@nestjs/common';
 import { applyDecorators, mixin, NotFoundException, UseInterceptors } from '@nestjs/common';
-import type { Client } from '@openfeature/server-sdk';
-import { OpenFeature } from '@openfeature/server-sdk';
+import { getClientForEvaluation } from './utils';
+import type { EvaluationContext } from '@openfeature/server-sdk';
+
+type RequiredFlag = {
+  flagKey: string;
+  defaultValue?: boolean;
+};
 
 /**
- * Options for injecting a feature flag into a route handler.
+ * Options for using one or more Boolean feature flags to control access to a Controller or Route.
  */
 interface RequireFlagsEnabledProps {
   /**
-   * The key of the feature flag.
+   * The key and default value of the feature flag.
    * @see {@link Client#getBooleanValue}
    */
-  flagKeys: string[];
+  flags: RequiredFlag[];
+
   /**
    * The exception to throw if any of the required feature flags are not enabled.
    * Defaults to a 404 Not Found exception.
    * @see {@link HttpException}
+   * @default new NotFoundException(`Cannot ${req.method} ${req.url}`)
    */
   exception?: HttpException;
 
@@ -24,15 +31,12 @@ interface RequireFlagsEnabledProps {
    * @see {@link OpenFeature#getClient}
    */
   domain?: string;
-}
 
-/**
- * Returns a domain scoped or the default OpenFeature client with the given context.
- * @param {string} domain The domain of the OpenFeature client.
- * @returns {Client} The OpenFeature client.
- */
-function getClientForEvaluation(domain?: string) {
-  return domain ? OpenFeature.getClient(domain) : OpenFeature.getClient();
+  /**
+   * Global {@link EvaluationContext} for OpenFeature.
+   * @see {@link OpenFeature#setContext}
+   */
+  context?: EvaluationContext;
 }
 
 /**
@@ -43,9 +47,15 @@ function getClientForEvaluation(domain?: string) {
  * For example:
  * ```typescript
  * @RequireFlagsEnabled({
- *   flagKeys: ['flagName', 'flagName2'],  // Required, an array of Boolean feature flag keys
- *   exception: new ForbiddenException(),  // Optional, defaults to a 404 Not Found exception
- *   domain: 'my-domain',                  // Optional, defaults to the default OpenFeature client
+ *   flags: [                               // Required, an array of Boolean flags to check, with optional default values (defaults to false)
+ *     { flagKey: 'flagName' },
+ *     { flagKey: 'flagName2', defaultValue: true },
+ *   ],
+ *   exception: new ForbiddenException(),   // Optional, defaults to a 404 Not Found Exception
+ *   domain: 'my-domain',                   // Optional, defaults to the default OpenFeature Client
+ *   context: {                             // Optional, defaults to the global OpenFeature Context
+ *     targetingKey: 'user-id',
+ *   },
  * })
  * @Get('/')
  * public async handleGetRequest()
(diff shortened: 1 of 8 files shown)


toddbaert:
I think this is a good idea, and it "feels right" to me ergonomically.

Make sure you add something to the [test-app](https://github.com/open-feature/js-sdk/blob/main/packages/nest/test/test-app.ts) to test and document this.

packages/nest/src/require-flags-enabled.decorator.ts
```diff
@@ -0,0 +1,80 @@
...
+   * The key of the feature flag.
+   * @see {@link Client#getBooleanValue}
+   */
+  flagKeys: string[];
+  /**
+   * The exception to throw if any of the required feature flags are not enabled.
+   * Defaults to a 404 Not Found exception.
+   * @see {@link HttpException}
+   */
+  exception?: HttpException;
```
lukas-reining, line 20: I think it is okay to constrain the exceptions to HTTP exceptions, even though it might make sense to broaden this to general exceptions. 
Sometimes custom filters are built for more complex cases so I think one could argue for being strict on that or allowing any exception to be thrown.

packages/nest/src/require-flags-enabled.decorator.ts
```diff
@@ -0,0 +1,80 @@
+import type { CallHandler, ExecutionContext, HttpException, NestInterceptor } from '@nestjs/common';
+import { applyDecorators, mixin, NotFoundException, UseInterceptors } from '@nestjs/common';
+import type { Client } from '@openfeature/server-sdk';
+import { OpenFeature } from '@openfeature/server-sdk';
+
+/**
+ * Options for injecting a feature flag into a route handler.
+ */
+interface RequireFlagsEnabledProps {
```
lukas-reining, line 9: I think we could have some optional static evaluation context given here too.

lukas-reining, line 9: Also we could have a context factory here too. Basically what we globally have for the OpenFeatureModule:
 https://github.com/open-feature/js-sdk/blob/cf89e7da24f43e8142e117ae3b662f380a66e514/packages/nest/src/open-feature.module.ts#L138
 
But this would be optional to me. We do not have it for the other decorators too:
https://github.com/open-feature/js-sdk/blob/cf89e7da24f43e8142e117ae3b662f380a66e514/packages/nest/src/feature.decorator.ts#L36

So to me this is optional for this PR but it would be a great addition.

packages/nest/src/require-flags-enabled.decorator.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * Returns a domain scoped or the default OpenFeature client with the given context.
+ * @param {string} domain The domain of the OpenFeature client.
+ * @returns {Client} The OpenFeature client.
+ */
+function getClientForEvaluation(domain?: string) {
+  return domain ? OpenFeature.getClient(domain) : OpenFeature.getClient();
+}
```
lukas-reining, line 36: We already have this function: 
https://github.com/open-feature/js-sdk/blob/cf89e7da24f43e8142e117ae3b662f380a66e514/packages/nest/src/feature.decorator.ts#L65

packages/nest/src/require-flags-enabled.decorator.ts
```diff
@@ -0,0 +1,80 @@
...
+ *   domain: 'my-domain',                  // Optional, defaults to the default OpenFeature client
+ * })
+ * @Get('/')
+ * public async handleGetRequest()
+ * ```
+ * @param {RequireFlagsEnabledProps} props The options for injecting the feature flag.
+ * @returns {ClassDecorator & MethodDecorator} The decorator that can be used to require Boolean Feature Flags to be enabled for a controller or a specific route.
+ */
+export const RequireFlagsEnabled = (props: RequireFlagsEnabledProps): ClassDecorator & MethodDecorator =>
+  applyDecorators(UseInterceptors(FlagsEnabledInterceptor(props)));
```
lukas-reining, line 57: I think there could be value in similar function like: `RequireFlagEquals`.
But this could also be added in another PR.

packages/nest/src/require-flags-enabled.decorator.ts
```diff
@@ -0,0 +1,80 @@
...
+const FlagsEnabledInterceptor = (props: RequireFlagsEnabledProps) => {
+  class FlagsEnabledInterceptor implements NestInterceptor {
+    constructor() {}
+
+    async intercept(context: ExecutionContext, next: CallHandler) {
+      const req = context.switchToHttp().getRequest();
+      const client = getClientForEvaluation(props.domain);
+
+      for (const flagKey of props.flagKeys) {
+        const endpointAccessible = await client.getBooleanValue(flagKey, false);
```
lukas-reining, line 68: Sometimes there are cases where it is better to fall back to `true` as default value if the flag evaluation fails. 
I think it would be worth it adding the default value as a parameter, defaulting to false.

packages/nest/src/require-flags-enabled.decorator.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+interface RequireFlagsEnabledProps {
+  /**
+   * The key of the feature flag.
+   * @see {@link Client#getBooleanValue}
+   */
+  flagKeys: string[];
+  /**
+   * The exception to throw if any of the required feature flags are not enabled.
+   * Defaults to a 404 Not Found exception.
```
lukas-reining, line 17: I am not entirely sure why you chose 404 here. 
Is it to prevent unwanted information disclosure in the case the feature is meant to be hidden?

lukas-reining (approved):
I really like the addition!
Added some optional additions to it. Feel free to add them, otherwise we can add them in another PR :) 
Thank you!

lukas-reining:
I will have a look at what is wrong with the e2e test tomorrow @kaushalkapasi :)

packages/nest/src/require-flags-enabled.decorator.ts
kaushalkapasi: Yeah! I opted to implement this first as the "easiest" option, a follow up would be to take a flag key and expected value combination to support String, Number and Object type flags.

packages/nest/src/require-flags-enabled.decorator.ts
kaushalkapasi: Should I change that to be an exported function in a `utils.ts` file that can be imported across decorators? Additionally, should the decorators live in their own folder or folders, along the lines of: `nest/src/decorators/require-flags-enabled/require-flags-enabled.ts`?

packages/nest/src/require-flags-enabled.decorator.ts
kaushalkapasi: What would this look like? Is there an example you could share (maybe in another SDK) since I'm not fully familiar with that concept.

packages/nest/src/require-flags-enabled.decorator.ts
kaushalkapasi: That's right. Defaulting to 404 as the preferred behaviour is to "hide" the existence of the Controller or Route from the caller entirely. This could also be integrated with any API documentation that the end user may generate.

packages/nest/src/require-flags-enabled.decorator.ts
kaushalkapasi: Sounds good, will do.

toddbaert:
> I will have a look at what is wrong with the e2e test tomorrow @kaushalkapasi :)

Seems to me like the test-harness module was just deleted.

packages/nest/src/require-flags-enabled.decorator.ts
toddbaert: Ya I actually think for security reasons this is the most reasonable default.

packages/nest/src/require-flags-enabled.decorator.ts
toddbaert: Yep that makes sense to me. I think most people would use this current variant, which could basically just be a shorthand for a `RequireFlagEquals` implementation. I think instead of a key/value combination though, a lambda predicate might be easier (a lamda taking the EvalutionDetails which will run to decide to run the request or not).

packages/nest/src/require-flags-enabled.decorator.ts
```diff
@@ -0,0 +1,90 @@
...
+   * The domain of the OpenFeature client, if a domain scoped client should be used.
+   * @see {@link OpenFeature#getClient}
+   */
+  domain?: string;
+
+  /**
+   * Global {@link EvaluationContext} for OpenFeature.
+   * @see {@link OpenFeature#setContext}
+   */
+  context?: EvaluationContext;
```
toddbaert, line 39: I'm a bit confused as to why we'd want to set the global context here?

It seems like a bit of a separation-of-concerns violation. Can somebody help me understand why it's useful here?

packages/nest/test/fixtures.ts
```diff
@@ -23,6 +23,11 @@ export const defaultProvider = new InMemoryProvider({
     variants: { default: { client: 'default' } },
     disabled: false,
   },
+  testBooleanFlag2: {
+    defaultVariant: 'default',
+    variants: { default: false, enabled: true },
+    disabled: false,
+  },
```
toddbaert, line 30: Could we add another integration test where we use a flag with a simple targeting rule, which requires an evaluation context value (which is injected in an interceptor) to be present? That would ensure that we are using a per-request context as expected in our decorator(s).

toddbaert (changes requested):
I have a couple questions, and one [request for an additional test](https://github.com/open-feature/js-sdk/pull/1159#discussion_r2017469514) (please let me know if you think this is not necessary) but this looks really cool. Great work!

Happy to approve from my perspective after these are resolved.

packages/nest/test/fixtures.ts
kaushalkapasi: Sounds good, will add the test.

packages/nest/src/require-flags-enabled.decorator.ts
kaushalkapasi: That's a mistake, it shouldn't be the global context. Will be updated to: `* The {@link EvaluationContext} for evaluating the feature flag.`

lukas-reining:
Sorry for taking time @kaushalkapasi, many of us have been at KubeCon. 
Will have a look tomorrow or the day after :)

packages/nest/src/require-flags-enabled.decorator.ts
lukas-reining: You already have it with the `context` in line 39, I must have missed this last time :)

packages/nest/src/require-flags-enabled.decorator.ts
lukas-reining: @kaushalkapasi do you want to do this in this PR? 
Because you requested reviews again but I could not find it yet. 
We can also leave it out of here, but it might be really good to have here.

kaushalkapasi: I haven't added the lambda function in this PR. My only concern is that it would complicate the usage of this Boolean only provider.

I would definitely add it when we build a decorator for other flag types (String, Number and Object).

Let me know if you think it should be implemented here and I'd be happy to do so.

lukas-reining: If you do it like @toddbaert describes, `RequireFlagsEnabled` would simply be an alias for `RequireFlagEquals` with `evalutationDetails.value === true`.
It would not make the `RequireFlagsEnabled` more complicated in this case.

I would prefer to have it.

kaushalkapasi: Sounds good, I'll update this implementation soon and tag you for a re-review when ready!

toddbaert (approved):
Implementation looks great. Like @lukas-reining says I think we could use some docs so people know how to use this cool new feature... but I'm approving.

beeme1mr (approved):

PR #1164, opened 2025-04-09
feat(react): add FeatureFlag component

weyert:
## This PR
Introduces the FeatureFlag component for React that allow using feature flags in a declarative manner

### Related Issues
No ticket

### Notes
Maybe consider adding a similar component for other supported frameworks?

### How to test
I have written unit tests to cover the main features of the component


packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Props for the Feature component that conditionally renders content based on feature flag state.
+ * @interface FeatureProps
+ */
+interface FeatureProps {
+  /**
+   * The key of the feature flag to evaluate.
+   */
+  featureKey: string;
```
toddbaert, line 13: I think this should be called `key`... or possibly `flagKey`, but I think that's redundant given the name of the component.

packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Optional value to match against the feature flag value.
+   * If provided, the component will only render children when the flag value matches this value.
+   * If a boolean, it will check if the flag is enabled (true) or disabled (false).
+   * If a string, it will check if the flag variant equals this string.
+   */
+  match?: string | boolean;
```
toddbaert, line 21: I think it would be really cool to add support for a predicate function that accepts the `EvaluationDetails` so people can specify exactly how to match.

packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -0,0 +1,91 @@
...
+   * Content to render when the feature flag condition is met.
+   * Can be a React node or a function that receives flag query details and returns a React node.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  children: React.ReactNode | ((details: FlagQuery<any>) => React.ReactNode);
+
+  /**
+   * Optional content to render when the feature flag condition is not met.
+   */
+  fallback?: React.ReactNode;
```
toddbaert, line 39: I think if we are going to only have one alternative child (fallback) we definitely need to do [this](https://github.com/open-feature/js-sdk/pull/1164/files#r2036805701), so that users can have control over to render this in very specific situations (evaluation errors, reasons, etc).

The alternative (which I don't like as muich) would be to also specify `error` components, etc... but I think that isn't as good.

cc @beeme1mr @lukas-reining

toddbaert (commented):
This is really cool. I really like the idea and the concepts. I've made a few minor proposals that I think simply the API but also make it slightly more flexible.

packages/react/src/declarative/FeatureFlag.tsx
lukas-reining: > I think it would be really cool to add support for a predicate function that accepts the EvaluationDetails so people can specify exactly how to match.

+1

Also with this, you could make this function work for all the types. 
And it could have a default that would be 

```ts
function equals<T>(expected: T, actual: EvaluationDetails<T>){
    return expected === actual.value
}
```

And for objects we could actually make this required if we wanted as this default does not make much sense here.
What do you think @weyert @toddbaert?

packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -0,0 +1,91 @@
...
+   * If a boolean, it will check if the flag is enabled (true) or disabled (false).
+   * If a string, it will check if the flag variant equals this string.
+   */
+  match?: string | boolean;
+
+  /**
+   * Default value to use when the feature flag is not found.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  defaultValue: any;
```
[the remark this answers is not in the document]
lukas-reining: With the predicate from above this could be generic. 
The whole component would become nicely typed then as the same type param can be used for the predicate then, so it will always fit the default value.

packages/react/src/declarative/FeatureFlag.tsx
lukas-reining: > The alternative (which I don't like as muich) would be to also specify error components, etc... but I think that isn't as good.

I think for it feels a little more idiomatic to have the matcher that is also able to match for error, as you said @toddbaert.
But both could work for me.

packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Optional content to render when the feature flag condition is not met.
+   */
+  fallback?: React.ReactNode;
+
+  /**
+   * If true, inverts the condition logic (renders children when condition is NOT met).
+   */
+  negate?: boolean;
```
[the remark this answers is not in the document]
lukas-reining: +1

packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -0,0 +1,91 @@
...
+ * @interface FeatureProps
+ */
+interface FeatureProps {
+  /**
+   * The key of the feature flag to evaluate.
+   */
+  featureKey: string;
+
+  /**
+   * Optional value to match against the feature flag value.
```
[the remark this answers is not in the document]
lukas-reining: if we do the default mentioned below, we can just describe it as `By default, strict equality is used`

packages/react/src/declarative/FeatureFlag.tsx
toddbaert: Actually ya, I think that's a great idea.

beeme1mr:
Hey @weyert, could you please respond to the feedback when you have a moment? It looks like this change is almost ready and it would be great to get it out soon.

aepfli:
> I will see if I can fix the issue by merging main back into my branch

a handy little git command which i like to use in this situations
```
git reset $(git merge-base upstream/main $(git branch --show-current))
```
it will reset the branch to upstream main, but will keep all the changes ready to be committed again. replace `upstream/main` with the reference branch of your desire ;)

weyert:
@aepfli Thanks, I will try it out tomorrow. I might message you lol

packages/react/src/declarative/FeatureFlag.tsx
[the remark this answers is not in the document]
weyert: Added in the jsdoc comment

packages/react/test/declarative.spec.tsx
```diff
@@ -0,0 +1,209 @@
...
+  OpenFeature.setProvider(EVALUATION, makeProvider());
+
+  const childText = 'Feature is active';
+  const ChildComponent = () => <div>{childText}</div>;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('<FeatureFlag />', () => {
```
beeme1mr, line 59: What's the expected behavior of a non-boolean flag that doesn't define a `match`? Will this always be truthy or falsy?

packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -0,0 +1,104 @@
...
+import type { FlagQuery } from '../query';
+import type { FlagValue, EvaluationDetails } from '@openfeature/core';
+
+/**
+ * Default predicate function that checks if the expected value equals the actual flag value.
+ * @param {T} expected The expected value to match against
+ * @param {EvaluationDetails<T>} actual The evaluation details containing the actual flag value
+ * @returns {boolean} true if the values match, false otherwise
+ */
+function equals<T extends FlagValue>(expected: T, actual: EvaluationDetails<T>): boolean {
```
beeme1mr, line 13: It may be better to reuse this `is-equal` check.

https://github.com/open-feature/js-sdk/blob/main/packages/react/src/internal/is-equal.ts

packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -0,0 +1,104 @@
...
+interface FeatureFlagProps<T extends FlagValue = FlagValue> {
+  /**
+   * The key of the feature flag to evaluate.
+   */
+  flagKey: string;
+
+  /**
+   * Optional value to match against the feature flag value.
+   * If provided, the component will only render children when the flag value matches this value.
+   * By default, strict equality (===) is used for comparison.
```
beeme1mr: Can this be overridden? The phrase "by default" implies that it can be but I'm not show how.

packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -0,0 +1,104 @@
...
+   * Content to render when the feature flag condition is met.
+   * Can be a React node or a function that receives flag query details and returns a React node.
+   */
+  children: React.ReactNode | ((details: FlagQuery<T>) => React.ReactNode);
+
+  /**
+   * Optional content to render when the feature flag condition is not met.
+   * Can be a React node or a function that receives evaluation details and returns a React node.
+   */
+  fallback?: React.ReactNode | ((details: EvaluationDetails<T>) => React.ReactNode);
```
beeme1mr, line 46: Should we also consider an optional loading component? This may have already been discussed but I didn't notice it in the thread yet.

packages/react/test/declarative.spec.tsx
marcozabel: As we discussed, non-boolean flags should probably not be shown if the match function is not specified.

packages/react/src/declarative/FeatureFlag.tsx
marcozabel: Yes, it can be overriden if a predicate is set.

packages/react/test/declarative.spec.tsx
toddbaert: @marcozabel @beeme1mr I think we can improve the types here to actually indicate this right...? if the `defaultValue` is a non-boolean, we could make the match required. We should still be defensive for the non-TS usage, but I think this will make the API clear from a types perspective.

packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/**
+ * FeatureFlag component that conditionally renders its children based on the evaluation of a feature flag.
+ * @param {FeatureFlagProps} props The properties for the FeatureFlag component.
```
toddbaert:
```suggestion
/**
 * @experimental This API is experimental, and is subject to change.
 * FeatureFlag component that conditionally renders its children based on the evaluation of a feature flag.
 * @param {FeatureFlagProps} props The properties for the FeatureFlag component.
```

toddbaert (approved):
This has been open a while, but I think it's a super cool feature. I'm approving, with 2 recommendations:

- Mark this as experimental: https://github.com/open-feature/js-sdk/pull/1164/changes#r2620705691
- Minor improvements to types to communicate that a `match` should always be provided for non-boolean flags cc @beeme1mr @marcozabel : https://github.com/open-feature/js-sdk/pull/1164#discussion_r2620694816

packages/react/src/declarative/FeatureFlag.tsx
toddbaert: I think this is a good idea, but I would leave this as a new issue - this has already been open for a while. If we agree I'm happy to make it.

marcozabel:
Hey, I improved the typing and jsdoc in this [PR](https://github.com/open-feature/js-sdk/pull/1319). Let me know if I you need any help or something is still missing 👍

packages/react/src/declarative/FeatureFlag.tsx
lukas-reining: I mean with `EvaluationDetails` provided, this can do all a loading component can do. 
So I think it is good to have but can be made in a separate issue as @toddbaert says?

packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -0,0 +1,124 @@
...
+      match?: T | undefined;
+    }
+  : {
+      /**
+       * Value to match against the feature flag value.
+       * Required for non-boolean flags to determine when children should render.
+       * By default, strict equality is used for comparison.
+       */
+      match: T;
+    });
```
lukas-reining, line 73: I think it would be a little more intuitive if the name of the expected value param would be the same for this and for the predicate. 
I am leaning a bit more towards `expected` than `match`.

weyert:
Do we want to merge it and solve the loading component aspect in a separate improvement PR?

packages/react/src/declarative/FeatureFlag.tsx
toddbaert: I agree it would be great if these names were consistent.

However, I'm not 100% sure about "expect"/"expected", especially in JS, this to me strongly implies testing, which I feel like brings in conceptual baggage.

I wonder if going the other way is better, and changing the predicate...

 from:

```ts
  /**
   * Optional predicate function for custom matching logic.
   * If provided, this function will be used instead of the default equality check.
   * @param expected The expected value (from match prop)
   * @param actual The evaluation details
   * @returns true if the condition is met, false otherwise
   */
  predicate?: (expected: T | undefined, actual: EvaluationDetails<T>) => boolean;
```

to: 

```ts
  /**
   * Optional predicate function for custom matching logic.
   * If provided, this function will be used instead of the default equality check.
   * @param match The expected value (from match prop)
   * @param details The evaluation details
   * @returns true if the condition is met, false otherwise
   */
  predicate?: (match: T | undefined, details: EvaluationDetails<T>) => boolean;
```

This is very plain/simple... the names are just what they are - the match prop you specified, and the evaluation details.

toddbaert: Perhaps `matchValue` would be even better for both, since it's specifically the `value` of the details we are matching.

marcozabel: I agree with @toddbaert that "match" is more fitting, as "expected" tends to carry a testing connotation, which doesn’t align with the context of feature flags where there’s no predefined expectation of a value.

toddbaert: I'll make this change.

toddbaert: https://github.com/open-feature/js-sdk/pull/1164/changes/c6e7e1ef0754be8bfb8e563d329da36f150f3037

weyert: Thank you, change makes sense

PR #1189, opened 2025-05-14
refactor(telemetry): update telemetry attributes and remove unused evaluation data

### LINKED ISSUE #1188: Update the OTel-Compatible Telemetry Util Function to latest spec
### Requirements

- [Match latest spec](https://github.com/open-feature/spec/pull/306)
- Update the description to include a disclaimer that this feature is part of the experimental OTel semcon.

> [!NOTE]
> This will be a breaking change but it's required to match the latest spec. It's an experimental feature, so it doesn't require a major release but the change should be denoted in the change log.

beeme1mr:
## This PR

- update the evaluation event to match the latest OTel semcon

### Related Issues

Fixes #1188

### Notes

Update the createEvaluationEvent to match the latest OpenTelemetry spec. Unforutnately, this is a breaking change. I'll denote this in the release notes but not in the library version. That's because we're following the experimental OTel spec and the method itself is currently undocumented except from the generated JS Doc.

It's also worth noting that OTel JS SDK event API doesn't currently support object as attribute values. @dyladan will update the OTel SDK.

### Follow-up Tasks

Update the ToggleShop.

### How to test

`npm run test`


beeme1mr:
Waiting for the OTel JS SDK to support this.

https://github.com/open-telemetry/opentelemetry-js/pull/5707

aepfli (approved):

beeme1mr:
@toddbaert Should we get this into the current release or wait for the next one?

PR #1203, opened 2025-05-29
chore: update node to v20+

toddbaert:
node v18 is now EOL

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,9 +31,6 @@
     "url": "https://github.com/open-feature/js-sdk/issues"
   },
   "homepage": "https://github.com/open-feature/js-sdk#readme",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -31,9 +31,6 @@
     "url": "https://github.com/open-feature/js-sdk/issues"
   },
   "homepage": "https://github.com/open-feature/js-sdk#readme",
-  "engines": {
-    "node": ">=18"
-  },
```
toddbaert, line 36: Pretty sure this is doing nothing since it's at the root.

jonathannorris: yea this is the client side SDK, so yea I don't think this does anything.

PR #1216, opened 2025-06-06
feat: add evaluation-scoped hook data

### LINKED ISSUE #1215: Add Hook Data Support
## Summary

The OpenFeature specification now includes support for hook data, which allows hooks to maintain state across their execution stages (before → after/error → finally). This feature enables use cases like distributed tracing, performance monitoring, and multi-stage validation.

This issue tracks the implementation of hook data support in this SDK to achieve compliance with the latest specification.

## Action Required

SDK maintainers need to implement hook data support according to the specification defined in:
- https://github.com/open-feature/spec/pull/273 - Initial hook data specification
- https://github.com/open-feature/spec/pull/307 - Hook data updates

A reference implementation is available in the .NET SDK PR: https://github.com/open-feature/dotnet-sdk/pull/387

## Hook data specification enables stateful hook operations

The OpenFeature hook data specification provides a mechanism for different stages of the same hook to share state during a single flag evaluation. This feature supports implementing telemetry, performance monitoring, and resource management use cases. The .NET SDK's implementation in https://github.com/open-feature/dotnet-sdk/pull/387 demonstrates a successful approach that other SDKs can follow.

## Reference Links

- **Specification PRs**: 
  - https://github.com/open-feature/spec/pull/273 - Initial hook data specification
  - https://github.com/open-feature/spec/pull/307 - Hook data updates
- **Implementation Example**: https://github.com/open-feature/dotnet-sdk/pull/387
- **Documentation**:
  - [OpenFeature Hook Specification](https://openfeature.dev/specification/sections/hooks/)
  - [Hook Concepts Documentation](https://openfeature.dev/docs/reference/concepts/hooks/)

## Understanding hook data fundamentals

Hook data provides the ability for hooks to maintain state across their execution lifecycle. When a hook starts an operation in the `before` stage (like opening a telemetry span), it needs a way to access that same data in the `after` stage to complete the operation. Without hook data, developers resort to workarounds like global state, weak maps, or thread-local storage, which introduce complexity and potential bugs.

The specification defines hook data as a **mutable, per-hook, per-evaluation** data structure. This scoping means each hook instance gets its own isolated data store that persists only for the duration of a single flag evaluation. Different hooks evaluating the same flag don't share data, preventing interference and ensuring predictable behavior.

### Data structure and access patterns

Hook data **MUST** be implemented as a structure supporting arbitrary properties with string keys and values of any type. The structure must be mutable to allow stages to modify shared state. SDKs typically implement this as a map-like interface with `set(key, value)` and `get(key)` methods.

The data structure must be created before the first hook stage executes and remain available throughout all stages (before → after/error → finally). This lifecycle management is the SDK's responsibility, not the hook developer's.

### Isolation and scoping guarantees

Each hook instance maintains its own hook data - there's no sharing between different hooks. This isolation extends to:
- Different hook types running for the same evaluation
- Multiple instances of the same hook type
- Hooks registered at different levels (global, client, invocation)

The specification explicitly states that hook data is not shared between different hooks, ensuring separation of concerns.

### Integration with hook context

Hook data must be accessible through the hook context parameter passed to all hook stages. The typical pattern follows:

```pseudocode
context.hookData.set("key", value)  // In before stage
value = context.hookData.get("key") // In after stage
```

## Learning from the .NET SDK implementation

The .NET SDK's PR https://github.com/open-feature/dotnet-sdk/pull/387 provides implementation insights. Their approach enhanced the existing `HookContext<T>` class with a `Data` property, maintaining backward compatibility while adding new functionality. Lessons include:

**API Design**: The implementation uses a simple key-value store pattern that's intuitive for developers. Type safety is handled through casting when retrieving values, following .NET conventions.

**No Breaking Changes**: Existing hooks continue working without modification. The hook data feature is additive - hooks can opt-in to using it without requiring changes to hooks that don't need state sharing.

**Examples**: The implementation includes examples like timing hooks that measure execution duration, demonstrating the feature's value to developers.

### Phase 1: Core implementation

Start by implementing the basic hook data structure with these capabilities:
- String-keyed storage with any-type values
- Mutable data access through set/get operations
- Proper lifecycle management (creation before first stage, cleanup after finally)
- Thread-safe access for concurrent evaluations

### Phase 2: Integration and testing

Integrate hook data with your existing hook infrastructure:
- Add data property to hook context
- Ensure data flows through all hook stages
- Implement isolation between hook instances
- Create comprehensive test suites covering edge cases

### Phase 3: Documentation and examples

Provide guidance for hook developers:
- API documentation with type signatures
- Examples (timing, telemetry, validation)
- Best practices and anti-patterns
- Migration guide if introducing any changes

### Memory management

Hook data should be garbage collected after the evaluation completes. Consider implementing weak references or explicit cleanup to prevent memory leaks in long-running applications.

### Thread safety

Many applications evaluate flags concurrently. Ensure hook data access is thread-safe without introducing significant performance overhead. Consider using concurrent data structures appropriate for your language.

### Type safety

In statically-typed languages, consider providing type-safe access patterns while maintaining the flexibility of arbitrary value storage. Generic methods or type assertions can help balance safety with usability.

### Performance impact

Hook data should add minimal overhead to flag evaluations. Consider lazy initialization - only create the data structure when a hook actually uses it. Benchmark the implementation to ensure acceptable performance.

### OpenTelemetry integration

The canonical use case involves creating spans in the `before` stage and ending them in `after`:

```java
// Before stage
Span span = tracer.spanBuilder("feature-flag-evaluation").startSpan();
context.hookData.set("span", span);

// After stage
Span span = (Span) context.hookData.get("span");
span.end();
```

### Performance monitoring

Tracking evaluation duration is a common requirement:

```csharp
// Before stage
context.Data.Set("startTime", DateTime.UtcNow);

// After stage
var duration = DateTime.UtcNow - (DateTime)context.Data.Get("startTime");
logger.LogInformation($"Evaluation took {duration.TotalMilliseconds}ms");
```

### Multi-stage validation

Complex validation scenarios benefit from accumulating results:

```python
# Before stage
context.hook_data.set("validation_errors", [])

# Multiple validation stages can add errors
errors = context.hook_data.get("validation_errors")
if not is_valid(context):
    errors.append("Validation failed")

# Finally stage checks accumulated errors
if context.hook_data.get("validation_errors"):
    logger.warning(f"Validation issues: {errors}")
```

## Testing requirements

Comprehensive test coverage should include:

1. **Basic functionality**: Set and get operations work correctly
2. **Isolation**: Data doesn't leak between hook instances
3. **Lifecycle**: Data persists through all stages but not beyond
4. **Type handling**: Various data types can be stored and retrieved
5. **Edge cases**: Null values, missing keys, concurrent access
6. **Performance**: Minimal overhead for evaluations not using hook data

## Migration and compatibility

When implementing hook data support:

1. **Maintain backward compatibility**: Existing hooks must continue working
2. **Version appropriately**: Follow semantic versioning based on impact
3. **Provide migration guides**: Help developers adopt the new feature
4. **Consider feature flags**: Allow gradual rollout of hook data support

## Success criteria

A complete hook data implementation should:
- Pass all specification compliance tests
- Support the primary use cases (telemetry, monitoring, validation)
- Add minimal performance overhead
- Provide good developer experience
- Include comprehensive documentation
- Maintain backward compatibility

## Conclusion

Hook data provides a feature that enables observability and monitoring capabilities in OpenFeature. By following the specification requirements and learning from successful implementations like the .NET SDK, maintainers can provide consistent, reliable hook data support across all OpenFeature SDKs. The implementation should balance flexibility with safety, performance with functionality, and maintain the isolation guarantees that make hook data predictable and useful.

## Additional Resources

- https://github.com/open-feature/spec
- [Hook Specification Details](https://openfeature.dev/specification/sections/hooks/)
- [Hook Concepts Guide](https://openfeature.dev/docs/reference/concepts/hooks/)
- [.NET SDK Documentation](https://openfeature.dev/docs/reference/technologies/server/dotnet/)
- https://github.com/open-feature/dotnet-sdk/releases

beeme1mr:
## This PR

- adds support for hook data in accordance to the spec
- adds optional type safety to the hook data keys and values

### Related Issues

Fixes #1215

### Notes

I had to add a bit more complexity in order to maintain support for existing hook implementations that used weakmaps.

### How to test

Unit tests have been added

### DIFF
--- a/packages/server/src/client/internal/open-feature-client.ts
+++ b/packages/server/src/client/internal/open-feature-client.ts
@@ -277,17 +277,10 @@ export class OpenFeatureClient implements Client {
 
     const mergedContext = this.mergeContexts(invocationContext);
 
-    // Create hook data instances for each hook
-    const hookDataMap = new WeakMap<Hook, DefaultHookData>();
-    for (const hook of allHooks) {
-      hookDataMap.set(hook, new DefaultHookData());
-    }
-
     // Create hook context instances for each hook (stable object references for the entire evaluation)
     // This ensures hooks can use WeakMaps with hookContext as keys across lifecycle methods
-    const hookContextMap = new WeakMap<Hook, HookContext>();
-    for (const hook of allHooks) {
-      const hookContext: HookContext = {
+    const hookContexts = allHooks.map<HookContext>(() =>
+      Object.freeze({
         flagKey,
         defaultValue,
         flagValueType: flagType,
@@ -295,29 +288,15 @@ export class OpenFeatureClient implements Client {
         providerMetadata: this._provider.metadata,
         context: mergedContext,
         logger: this._logger,
-        hookData: hookDataMap.get(hook)!,
-      };
-      
-      // Make the core properties immutable while allowing context updates
-      Object.defineProperty(hookContext, 'flagKey', { writable: false, configurable: false });
-      Object.defineProperty(hookContext, 'defaultValue', { writable: false, configurable: false });
-      Object.defineProperty(hookContext, 'flagValueType', { writable: false, configurable: false });
-      Object.defineProperty(hookContext, 'clientMetadata', { writable: false, configurable: false });
-      Object.defineProperty(hookContext, 'providerMetadata', { writable: false, configurable: false });
-      Object.defineProperty(hookContext, 'logger', { writable: false, configurable: false });
-      Object.defineProperty(hookContext, 'hookData', { writable: false, configurable: false });
-      
-      hookContextMap.set(hook, hookContext);
-    }
-
-    // Function to get the stable hook context for a given hook
-    const getHookContext = (hook: Hook) => hookContextMap.get(hook)!;
+        hookData: new DefaultHookData(),
+      }),
+    );
 
     let evaluationDetails: EvaluationDetails<T>;
     let frozenContext = mergedContext;
 
     try {
-      frozenContext = await this.beforeHooks(allHooks, getHookContext, mergedContext, options);
+      frozenContext = await this.beforeHooks(allHooks, hookContexts, mergedContext, options);
 
       this.shortCircuitIfNotReady();
 
@@ -332,34 +311,34 @@ export class OpenFeatureClient implements Client {
 
       if (resolutionDetails.errorCode) {
         const err = instantiateErrorByErrorCode(resolutionDetails.errorCode, resolutionDetails.errorMessage);
-        await this.errorHooks(allHooksReversed, getHookContext, err, options);
+        await this.errorHooks(allHooksReversed, hookContexts, err, options);
         evaluationDetails = this.getErrorEvaluationDetails(flagKey, defaultValue, err, resolutionDetails.flagMetadata);
       } else {
-        await this.afterHooks(allHooksReversed, getHookContext, resolutionDetails, options);
+        await this.afterHooks(allHooksReversed, hookContexts, resolutionDetails, options);
         evaluationDetails = resolutionDetails;
       }
     } catch (err: unknown) {
-      await this.errorHooks(allHooksReversed, getHookContext, err, options);
+      await this.errorHooks(allHooksReversed, hookContexts, err, options);
       evaluationDetails = this.getErrorEvaluationDetails(flagKey, defaultValue, err);
     }
 
-    await this.finallyHooks(allHooksReversed, getHookContext, evaluationDetails, options);
+    await this.finallyHooks(allHooksReversed, hookContexts, evaluationDetails, options);
     return evaluationDetails;
   }
 
   private async beforeHooks(
-    hooks: Hook[], 
-    getHookContext: (hook: Hook) => HookContext,
+    hooks: Hook[],
+    hookContexts: HookContext[],
     mergedContext: EvaluationContext,
-    options: FlagEvaluationOptions
+    options: FlagEvaluationOptions,
   ) {
     let accumulatedContext = mergedContext;
-    
-    for (const hook of hooks) {
-      const hookContext = getHookContext(hook);
-      
+
+    for (const [index, hook] of hooks.entries()) {
+      const hookContext = hookContexts[index];
+
       // Update the context on the stable hook context object
-      (hookContext as { context: EvaluationContext }).context = accumulatedContext;
+      Object.assign(hookContext.context, accumulatedContext);
 
       const hookResult = await hook?.before?.(hookContext, Object.freeze(options.hookHints));
       if (hookResult) {
@@ -367,13 +346,11 @@ export class OpenFeatureClient implements Client {
           ...accumulatedContext,
           ...hookResult,
         };
-      }
-    }
 
-    // Update all hook contexts with the final accumulated context and freeze it
-    for (const hook of hooks) {
-      const hookContext = getHookContext(hook);
-      (hookContext as { context: EvaluationContext }).context = Object.freeze(accumulatedContext);
+        for (let i = 0; i < hooks.length; i++) {
+          Object.assign(hookContexts[index].context, accumulatedContext);
+        }
+      }
     }
 
     // after before hooks, freeze the EvaluationContext.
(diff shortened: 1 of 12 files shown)


packages/server/src/client/internal/open-feature-client.ts
```diff
@@ -276,22 +277,47 @@ export class OpenFeatureClient implements Client {
...
+        flagValueType: flagType,
+        clientMetadata: this.metadata,
+        providerMetadata: this._provider.metadata,
+        context: mergedContext,
+        logger: this._logger,
+        hookData: hookDataMap.get(hook)!,
+      };
+      
+      // Make the core properties immutable while allowing context updates
+      Object.defineProperty(hookContext, 'flagKey', { writable: false, configurable: false });
```
beeme1mr: I don't love this part, but we need to freeze everything except the evaluation context.

packages/server/src/client/internal/open-feature-client.ts
```diff
@@ -276,22 +277,47 @@ export class OpenFeatureClient implements Client {
...
-      flagKey,
-      defaultValue,
-      flagValueType: flagType,
-      clientMetadata: this.metadata,
-      providerMetadata: this._provider.metadata,
-      context: mergedContext,
-      logger: this._logger,
-    };
+    // Create hook data instances for each hook
+    const hookDataMap = new WeakMap<Hook, DefaultHookData>();
```
kinyoklion: I don't think a WeakMap is correct here as the only reference to the initial data would be in the weak map, which would allow it to be collected before invocation.

The collection itself is scoped do the duration of hook execution, so a normal map should be fine as we wouldn't be retaining any strong references longer than hook invocation.

I am not completely sure about indexing by the hook instance as it maybe would be somewhat unexpected if you registered the same hook multiple times.

packages/server/src/client/internal/open-feature-client.ts
kinyoklion: I think you can just freeze it. Freezing isn't a "deep" freeze. So you still have interior mutability unless you recursively freeze components.

```
const test = {
  a: "a",
  b: new Map()
};

console.log(test);
// Output
// { a: 'a', b: Map(0) {} }
Object.freeze(test);

// This interior mutability isn't prevented by freeze.
test.b.set("test", "a")

console.log(test)
// Output
// { a: 'a', b: Map(1) { 'test' => 'a' } }

// The freeze does prevent changing a property of the frozen object.
test.a = "b"

console.log(test)
// Output
// { a: 'a', b: Map(1) { 'test' => 'a' } }
// The freeze prevented the value of a from becoming 'b'.
```

beeme1mr: Good point! I'll give that a shot.

packages/server/src/client/internal/open-feature-client.ts
beeme1mr: I think you're right. I probably can use a map in both instances because they'll both be GCed after the evaluation.

packages/web/src/client/internal/open-feature-client.ts
```diff
@@ -261,45 +265,49 @@ export class OpenFeatureClient implements Client {
...
-    hookContext: HookContext,
+    hookContexts: HookContext[],
     evaluationDetails: EvaluationDetails<FlagValue>,
     options: FlagEvaluationOptions,
   ) {
     // run "after" hooks sequentially
-    for (const hook of hooks) {
+    for (const [index, hook] of hooks.entries()) {
+      // Calculating index because after hooks run in reverse order
+      const hookContext = hookContexts[hooks.length - 1 - index];
```
lukas-reining: I would either provide the reversed hooks to `afterHooks`, `errorHooks` and `finallyHooks` or not provide the reversed list of hooks to them. 
Otherwise I think this could lead to problems or confusion later.

packages/shared/src/hooks/hook-data.ts
```diff
@@ -0,0 +1,80 @@
...
+  delete<K extends keyof TData>(key: K): boolean;
+  delete(key: string): boolean;
+  delete(key: string): boolean {
+    return this.data.delete(key);
+  }
+
+  clear(): void {
+    this.data.clear();
+  }
+}
```
lukas-reining: You do not really need the two overrides? `keyof TData` should be unknown when `TData` is not set, isn't it?

I would prefer the following implementation:

```suggestion
/**
 * Default implementation of HookData using a Map.
 * @template TData - A record type that defines the shape of the stored data
 */
export class DefaultHookData<TData = Record<string, unknown>> implements HookData<TData> {
  private readonly data = new Map<keyof TData, TData[keyof TData]>();

  set<K extends keyof TData>(key: K, value: TData[K]): void {
    this.data.set(key, value);
  }

  get<K extends keyof TData>(key: K): TData[K] | undefined {
    return this.data.get(key) as TData[K] | undefined;
  }

  has<K extends keyof TData>(key: K): boolean {
    return this.data.has(key);
  }

  delete<K extends keyof TData>(key: K): boolean {
    return this.data.delete(key);
  }

  clear(): void {
    this.data.clear();
  }
}
```

As the map can not return the correct type for a specific key, we need the cast there, but this does less casting then the implicit casts to `unknown`.

lukas-reining (approved):
Looks good, left two minor suggestions.

packages/web/src/client/internal/open-feature-client.ts
beeme1mr: I reversed the hook context list and special-cased the before hook.

packages/server/src/client/internal/open-feature-client.ts
beeme1mr: Done, thanks!

packages/server/src/client/internal/open-feature-client.ts
```diff
@@ -276,22 +277,27 @@ export class OpenFeatureClient implements Client {
...
+        hookData: new DefaultHookData(),
+      }),
+    );
 
     let evaluationDetails: EvaluationDetails<T>;
+    let frozenContext = mergedContext;
 
     try {
-      const frozenContext = await this.beforeHooks(allHooks, hookContext, options);
+      frozenContext = await this.beforeHooks(allHooks, hookContexts, mergedContext, options);
```
toddbaert: I think this can still be a const declared inside the try?

packages/shared/src/hooks/hook-data.ts
```diff
@@ -0,0 +1,72 @@
...
+   * Clears all values from the hook data store.
+   */
+  clear(): void;
+}
+
+/**
+ * Default implementation of HookData using a Map.
+ * @template TData - A record type that defines the shape of the stored data
+ */
+export class DefaultHookData<TData = Record<string, unknown>> implements HookData<TData> {
```
toddbaert: Personally, I would prefer to call this `MapHookData`. It implements the `HookData` interface by using a map, as the comment says, so I think that justifies such a name. `DefaultHookData` does really say much about how the implementation is actually accomplished.

toddbaert (approved):
Approved with 2 nits.

Testing looks good by the way, I like the timing use-case inclusion.

PR #1225, opened 2025-08-06
feat(web-global-build): impl

zdunecki:
## This PR

- Implements a global (IIFE) build for the web.
- Adds support for global distribution via CDN by including `unpkg` and `jsdelivr` fields in `package.json`.

### Notes

- The global build outputs to `dist/global/index.js`.
- The global bundle is exposed under the global name `OpenFeature`.

### Follow-up Tasks

- Consider adding automated tests or validation for the global bundle.
- If this bundle should be included in a release checklist or CI, link the appropriate issue here.

### How to test

1. Run `npm run build` and ensure `dist/global/index.js` is generated.
2. Serve the file locally or upload to a CDN and verify that `window.OpenFeature` is available in the browser console.

### Motivation
Trying to load OpenFeature as a first dependency before browser will resolve any modules.

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -3,8 +3,8 @@
   "version": "1.5.0",
   "description": "OpenFeature SDK for Web",
   "main": "./dist/cjs/index.js",
-  "unpkg": "dist/global/index.js",
-  "jsdelivr": "dist/global/index.js",
+  "unpkg": "dist/global/index.min.js",
+  "jsdelivr": "dist/global/index.min.js",
   "files": [
     "dist/"
   ],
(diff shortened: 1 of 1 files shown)


lukas-reining:
Hey @zdunecki, thanks for the change, I will have some time tomorrow to review.

beeme1mr (approved):
Hey @zdunecki, thanks for the PR. I don't have experience publishing for unpkg or jsdelivr but it appears that it should "just work" after we publish the next version to NPM. Is that correct?

https://www.jsdelivr.com/documentation#id-publishing-packages

zdunecki:
> Hey @zdunecki, thanks for the PR. I don't have experience publishing for unpkg or jsdelivr but it appears that it should "just work" after we publish the next version to NPM. Is that correct?
> 
> https://www.jsdelivr.com/documentation#id-publishing-packages

As far I know, that's correct.

@lukas-reining @beeme1mr 
I did a small change - added a minified version too. For global (browser) it seems more reasonable. 

Best,

PR #1226, opened 2025-08-07
fix(react): re-evaluate flags on re-render to detect silent provider …

beeme1mr:
## This PR

  - Added `useEffect` that runs on re-render to re-evaluate the flag value
  - Only updates state if the resolved value actually changed (using `isEqual` comparison)
  - Used lazy initialization for `useState` to avoid redundant initial evaluation
  - Added `useCallback` memoization for event handlers
  - Fixed `AbortController` scope issue

### Notes

This resolves a subtle issue where the provider state may update without emitting a change event, leading to confusing results. The `useFlag` hook sets the initial evaluated value in a `useState`. Since this wasn't in a closure, this evaluation happened any time the component using the hook rerendered but the result was essentially ignored. Adding a logging hook shows that the current results but since this evaluation was made in an existing `useState`, the result had no effect.

This resolves a subtle issue where the provider state may update without emitting a change event, leading to stale flag values being displayed.

The `useFlag` hook was evaluating the flag on every re-render (as part of the `useState` initialization), but because `useState` only uses its initial value on the first render, these subsequent evaluations were being discarded. This meant that even though the hook was fetching the correct updated value from the provider on each re-render, it was throwing that value away and continuing to display the stale cached value.

Adding a logging hook would show the correct evaluation happening (proving the provider had the updated value), but the UI would remain stuck with the old value because the `useState` was ignoring the re-evaluated result.

The fix ensures that these re-evaluations on re-render actually update the component's state when the resolved value changes.

The key insight is that the evaluation WAS happening on every re-render (due to how useState works), but React was discarding the result. Your fix makes those evaluations actually matter by checking if the value changed and updating state accordingly.

Original thread: https://cloud-native.slack.com/archives/C06E4DE6S07/p1754508917397519

### How to test

I created a test that reproduced the issue, and it failed. I then implemented the changes and verified that the test passed.

### DIFF
--- a/packages/react/src/evaluation/use-feature-flag.ts
+++ b/packages/react/src/evaluation/use-feature-flag.ts
@@ -8,7 +8,7 @@ import type {
   JsonValue,
 } from '@openfeature/web-sdk';
 import { ProviderEvents, ProviderStatus } from '@openfeature/web-sdk';
-import { useEffect, useRef, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import {
   DEFAULT_OPTIONS,
   isEqual,
@@ -287,8 +287,7 @@ function attachHandlersAndResolve<T extends FlagValue>(
   const client = useOpenFeatureClient();
   const status = useOpenFeatureClientStatus();
   const provider = useOpenFeatureProvider();
-
-  const controller = new AbortController();
+  const isFirstRender = useRef(true);
 
   if (defaultedOptions.suspendUntilReady && status === ProviderStatus.NOT_READY) {
     suspendUntilInitialized(provider, client);
@@ -298,9 +297,22 @@ function attachHandlersAndResolve<T extends FlagValue>(
     suspendUntilReconciled(client);
   }
 
-  const [evaluationDetails, setEvaluationDetails] = useState<EvaluationDetails<T>>(
+  const [evaluationDetails, setEvaluationDetails] = useState<EvaluationDetails<T>>(() =>
     resolver(client).call(client, flagKey, defaultValue, options),
   );
+  
+  // Re-evaluate when dependencies change (handles prop changes like flagKey)
+  useEffect(() => {
+    if (isFirstRender.current) {
+      isFirstRender.current = false;
+      return;
+    }
+    
+    const newDetails = resolver(client).call(client, flagKey, defaultValue, options);
+    if (!isEqual(newDetails.value, evaluationDetails.value)) {
+      setEvaluationDetails(newDetails);
+    }
+  }, [client, flagKey, defaultValue, options, resolver]);
 
   // Maintain a mutable reference to the evaluation details to have a up-to-date reference in the handlers.
   const evaluationDetailsRef = useRef<EvaluationDetails<T>>(evaluationDetails);
@@ -308,7 +320,7 @@ function attachHandlersAndResolve<T extends FlagValue>(
     evaluationDetailsRef.current = evaluationDetails;
   }, [evaluationDetails]);
 
-  const updateEvaluationDetailsCallback = () => {
+  const updateEvaluationDetailsCallback = useCallback(() => {
     const updatedEvaluationDetails = resolver(client).call(client, flagKey, defaultValue, options);
 
     /**
@@ -319,15 +331,19 @@ function attachHandlersAndResolve<T extends FlagValue>(
     if (!isEqual(updatedEvaluationDetails.value, evaluationDetailsRef.current.value)) {
       setEvaluationDetails(updatedEvaluationDetails);
     }
-  };
+  }, [client, flagKey, defaultValue, options, resolver]);
 
-  const configurationChangeCallback: EventHandler<ClientProviderEvents.ConfigurationChanged> = (eventDetails) => {
-    if (shouldEvaluateFlag(flagKey, eventDetails?.flagsChanged)) {
-      updateEvaluationDetailsCallback();
-    }
-  };
+  const configurationChangeCallback = useCallback<EventHandler<ClientProviderEvents.ConfigurationChanged>>(
+    (eventDetails) => {
+      if (shouldEvaluateFlag(flagKey, eventDetails?.flagsChanged)) {
+        updateEvaluationDetailsCallback();
+      }
+    },
+    [flagKey, updateEvaluationDetailsCallback],
+  );
 
   useEffect(() => {
+    const controller = new AbortController();
     if (status === ProviderStatus.NOT_READY) {
       // update when the provider is ready
       client.addHandler(ProviderEvents.Ready, updateEvaluationDetailsCallback, { signal: controller.signal });
@@ -348,7 +364,14 @@ function attachHandlersAndResolve<T extends FlagValue>(
       // cleanup the handlers
       controller.abort();
     };
-  }, []);
+  }, [
+    client,
+    status,
+    defaultedOptions.updateOnContextChanged,
+    defaultedOptions.updateOnConfigurationChanged,
+    updateEvaluationDetailsCallback,
+    configurationChangeCallback,
+  ]);
 
   return evaluationDetails;
 }
--- a/packages/react/test/provider.spec.tsx
+++ b/packages/react/test/provider.spec.tsx
@@ -288,7 +288,7 @@ describe('OpenFeatureProvider', () => {
         { timeout: DELAY * 4 },
       );
 
-      expect(screen.getByText('Will says hi')).toBeInTheDocument();
+      expect(screen.getByText('Will says aloha')).toBeInTheDocument();
     });
   });
 });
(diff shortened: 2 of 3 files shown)


packages/react/test/provider.spec.tsx
```diff
@@ -288,7 +288,7 @@ describe('OpenFeatureProvider', () => {
         { timeout: DELAY * 4 },
       );
 
-      expect(screen.getByText('Will says hi')).toBeInTheDocument();
+      expect(screen.getByText('Will says aloha')).toBeInTheDocument();
```
beeme1mr, line 291: This was a bit confusing at first but I think it makes sense now that I've looked into it. Basically, we're using the global context, which updates the evaluated state. The simulated button was in the same component as the flag being evaluated. Since the resolved value has changed, so does this value.

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -287,8 +287,7 @@ function attachHandlersAndResolve<T extends FlagValue>(
   const client = useOpenFeatureClient();
   const status = useOpenFeatureClientStatus();
   const provider = useOpenFeatureProvider();
-
-  const controller = new AbortController();
+  const isFirstRender = useRef(true);
```
beeme1mr, line 290: I'm using this `useRef` to avoid evaluating twice when the component first renders.

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -319,15 +331,19 @@ function attachHandlersAndResolve<T extends FlagValue>(
...
+    (eventDetails) => {
+      if (shouldEvaluateFlag(flagKey, eventDetails?.flagsChanged)) {
+        updateEvaluationDetailsCallback();
+      }
+    },
+    [flagKey, updateEvaluationDetailsCallback],
+  );
 
   useEffect(() => {
+    const controller = new AbortController();
```
beeme1mr, line 346: We were creating a new abort controller every time the hook ran. 😅

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -298,17 +297,30 @@ function attachHandlersAndResolve<T extends FlagValue>(
...
   }
 
-  const [evaluationDetails, setEvaluationDetails] = useState<EvaluationDetails<T>>(
+  const [evaluationDetails, setEvaluationDetails] = useState<EvaluationDetails<T>>(() =>
     resolver(client).call(client, flagKey, defaultValue, options),
   );
+  
+  // Re-evaluate when dependencies change (handles prop changes like flagKey)
+  useEffect(() => {
+    if (isFirstRender.current) {
```
moredip, line 306: I'm not super familiar with this implementation so take my feedback with a grain of salt, but it seems to me that the risks of this optimization outweigh the benefits. 

Benefit: as I understand it, this is a very small performance improvement from avoiding a double render after a component is first mounted. 

Risk: I can imagine some sort of weird edge case where an initial render uses bad/uninitialized state, then there's a second render that uses the right state, but then no other rendering occurs. In that scenario we'd end up with stale data.

Feels like the risk isn't worth it. Just my 2¢

Also, seems like since we're already comparing newDetails vs evaluationDetails, we're already protected against calling `setEvaluationDetails` unnecessarily. So the only optimization here is avoiding a flag evaluation call.

beeme1mr: Yeah, it's a little micro optimization that I added it right at the end. I could go either way, honestly.

lukas-reining: I do not see the problem, basically the useEffect and the initial value of the useState will be called after each other on the same sender. And after that only the useEffect will trigger for changes so I do not see the risk🤔
But I might not understand it right now.

toddbaert: I'll review this PR tomorrow and consider this.

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -298,17 +297,30 @@ function attachHandlersAndResolve<T extends FlagValue>(
     suspendUntilReconciled(client);
   }
 
-  const [evaluationDetails, setEvaluationDetails] = useState<EvaluationDetails<T>>(
+  const [evaluationDetails, setEvaluationDetails] = useState<EvaluationDetails<T>>(() =>
```
toddbaert, line 300: I don't see a reason for this, am I missing something?

beeme1mr: This prevents us from reevaluating and throwing away the data. Prior to the change, we always ran the evaluation but useState only kept the value from the first run.

toddbaert: So then this is the primary fix?

If so, I think there's a problem with our testing, because removing this doesn't fail any tests.

beeme1mr: It's not the fix but an optimization. I'll add a comment next to the fix.

toddbaert: Ah - OK - I understand now. We are doing this lazily _just_ to save the redundant evaluations, but this _is not_ the primary fix (the primary fix is below, where we update the state if it has changed).

I think my confusion could have been helped by a better comment below (I will make another comment about that). The comment there now actually made me think that use effect didn't have to do with the issue.

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -298,17 +297,30 @@ function attachHandlersAndResolve<T extends FlagValue>(
...
   );
+  
+  // Re-evaluate when dependencies change (handles prop changes like flagKey)
+  useEffect(() => {
+    if (isFirstRender.current) {
+      isFirstRender.current = false;
+      return;
+    }
+    
+    const newDetails = resolver(client).call(client, flagKey, defaultValue, options);
```
beeme1mr, line 311: @toddbaert this is the primary fix

packages/react/src/evaluation/use-feature-flag.ts
```diff
@@ -298,17 +297,30 @@ function attachHandlersAndResolve<T extends FlagValue>(
     suspendUntilReconciled(client);
   }
 
-  const [evaluationDetails, setEvaluationDetails] = useState<EvaluationDetails<T>>(
+  const [evaluationDetails, setEvaluationDetails] = useState<EvaluationDetails<T>>(() =>
     resolver(client).call(client, flagKey, defaultValue, options),
   );
+  
+  // Re-evaluate when dependencies change (handles prop changes like flagKey)
```
toddbaert: I think this comment really should be enhanced to describe the other reason for this `useEffect` (and the primary focus of this PR); to re-render if the evaluation details has changed (this is VERY different from a prop changing, which is all the comment says); this misled me in my review somewhat, so I'm sure it might confuse others later.

packages/react/src/evaluation/use-feature-flag.ts
toddbaert: @beeme1mr @moredip @lukas-reining I think the key here is that there's _probably_ not a lot of reasons that this value would change between the `useState` initialization and the `useEffect` run that wouldn't also trigger our other re-render callbacks below (the provider became ready, or changed state, etc, etc).

Since this is a workaround for providers that don't properly emit change events, it seems to me that we should at least wait for some new render to attempt to evaluate again (meaning we should keep @beeme1mr 's check here) since, there's no reason that the second eval here on initial render might be "more correct" than a first (the same argument could be made for checking a 3rd time, 4th time... etc)

To put it another way, @beeme1mr 's change "rechecks" the the flag at each render - the time between the initial render and the first useEffect run is negligible, so I can't imagine it's really ever worth considering (though maybe I'm missing something).

packages/react/src/evaluation/use-feature-flag.ts
toddbaert: I meant an inline comment. I was able to understand eventually. See: https://github.com/open-feature/js-sdk/pull/1226/files#r2279713217

packages/react/src/evaluation/use-feature-flag.ts
toddbaert:
```suggestion
  // Re-evaluate when dependencies change (handles prop changes like flagKey), or if during a re-render, we have detected a change in the evaluated value
```

toddbaert: https://github.com/open-feature/js-sdk/pull/1226/files#r2279726954

toddbaert (approved):
If you agree with [this](https://github.com/open-feature/js-sdk/pull/1226/files#r2279713217), please feel free to merge. If you don't, than at least one of us is confused about what's happening and I'd ask that you wait. 😅

PS: I did test this in the react demo app, and it had no negative side-effects that I could see.

packages/react/src/evaluation/use-feature-flag.ts
toddbaert: BTW, you could perhaps add a test for this optimization (assert the number of evaluations in the provider in some test) but it's not a blocker for me.

PR #1247, opened 2025-09-15
feat: add feature flag service for angular

lukas-reining:
NOTE: This is still a draft for discussion!

## This PR

Adds a service for evaluating flags in angular services. 
It enables application authors to evaluate flags directly as observables or signal. 

~~I am not 100% sure about returning signals or leaving the `toSignal` to the application autors.
As far as I understand, when used like the following, subscriptions for `toSignal` will still be cleaned up correctly, as the injection context will be captured correctly in that case:~~

```js
@Component({
  template: `
    <div data-testid="value">{{ thumbs()?.value ? '👍' : '👎' }}</div>
    <div data-testid="reason">reason: {{ thumbs()?.reason }}</div>
  `,
  standalone: true,
})
class TestSignalComponent {
  private flagService = inject(FeatureFlagService);
  thumbs = this.flagService.getBooleanDetailsSignal(FLAG_KEY, false);
}
```

Edit: We decided to leave the `toSignal` to the app authors.

@juanparadox please leave feedback.  

I still have to clean up the tests and the code in general, if we decide this is the design we opt for.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2025 OpenFeature Maintainers
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/package-lock.json
+++ b/package-lock.json
@@ -15900,9 +15900,9 @@
       }
     },
     "node_modules/rollup-plugin-dts": {
-      "version": "6.2.1",
-      "resolved": "https://registry.npmjs.org/rollup-plugin-dts/-/rollup-plugin-dts-6.2.1.tgz",
-      "integrity": "sha512-sR3CxYUl7i2CHa0O7bA45mCrgADyAQ0tVtGSqi3yvH28M+eg1+g5d7kQ9hLvEz5dorK3XVsH5L2jwHLQf72DzA==",
+      "version": "6.2.3",
+      "resolved": "https://registry.npmjs.org/rollup-plugin-dts/-/rollup-plugin-dts-6.2.3.tgz",
+      "integrity": "sha512-UgnEsfciXSPpASuOelix7m4DrmyQgiaWBnvI0TM4GxuDh5FkqW8E5hu57bCxXB90VvR1WNfLV80yEDN18UogSA==",
       "dev": true,
       "license": "LGPL-3.0-only",
       "dependencies": {
@@ -15915,7 +15915,7 @@
         "url": "https://github.com/sponsors/Swatinem"
       },
       "optionalDependencies": {
-        "@babel/code-frame": "^7.26.2"
+        "@babel/code-frame": "^7.27.1"
       },
       "peerDependencies": {
         "rollup": "^3.29.4 || ^4",
@@ -19181,7 +19181,7 @@
         "jsdom": "^26.1.0",
         "ng-packagr": "^20.1.0",
         "playwright": "^1.53.2",
-        "rxjs": "^7.8.2",
+        "rxjs": "~7.8.0",
         "tslib": "^2.3.0",
         "typescript": "^5.8.3",
         "vitest": "^3.2.4",
@@ -22034,7 +22034,7 @@
     },
     "packages/react": {
       "name": "@openfeature/react-sdk",
-      "version": "1.0.0",
+      "version": "1.0.1",
       "license": "Apache-2.0",
       "devDependencies": {
         "@openfeature/core": "*",
@@ -22061,7 +22061,7 @@
     },
     "packages/shared": {
       "name": "@openfeature/core",
-      "version": "1.9.0",
+      "version": "1.9.1",
       "license": "Apache-2.0",
       "devDependencies": {}
     },


beeme1mr:
@lukas-reining, what would it look like if you left the `toSignal` up to the author? Your proposal looks good to me but it has been years since I've used Angular.

juanparadox:
> > This would mean that the user can choose to consume the evaluation as an Observable or as a Signal, right? If that is the case, and if it is also the case that Angular wants to continue supporting both Observables and Signals, then I really like the approach of leaving toSignal to the user. This creates more flexibility in how the evaluation results are used/consumed.
> 
> Exactly, I also lean towards only having the observables in the API of the service. Having it in would just be hiding `toSignal` from the app authors.

Okay cool - I'd lean for leaving the `toSignal` out then 👍

lukas-reining:
Okay great!
This is ready to review now, I tried to make it similar to how React works @toddbaert.
Would be happy about a review @beeme1mr @juanparadox @toddbaert :)

packages/angular/projects/angular-sdk/src/lib/internal/is-equal.ts
```diff
@@ -0,0 +1,38 @@
+import { type FlagValue } from '@openfeature/web-sdk';
+
+/**
+ * Deeply compare two values to determine if they are equal.
+ * Supports primitives and serializable objects.
+ * @param {FlagValue} value First value to compare
+ * @param {FlagValue} other Second value to compare
+ * @returns {boolean} True if the values are equal
+ */
+export function isEqual(value: FlagValue, other: FlagValue): boolean {
```
beeme1mr, line 10: Should we move this to core? I could go either way since it's small and is unlikely to change but the same function is also in the React SDK.

packages/angular/projects/angular-sdk/src/lib/feature-flag.service.ts
```diff
@@ -0,0 +1,264 @@
...
+            }
+          },
+          { signal: controller.signal },
+        );
+      }
+
+      // Initial evaluation
+      updateValue();
+      return () => {
+        controller.abort();
```
beeme1mr, line 260: When will this callback fire? Sorry if it's a basic question, I'm not that familiar with Observables. I just want to make sure we're not prematurely unregistering the handlers.

packages/angular/projects/angular-sdk/src/lib/open-feature.module.ts
```diff
@@ -5,7 +5,7 @@ import { EvaluationContext, OpenFeature, Provider } from '@openfeature/web-sdk';
 export type EvaluationContextFactory = () => EvaluationContext;
 
 export interface OpenFeatureConfig {
-  provider: Provider;
+  provider?: Provider;
```
beeme1mr, line 13: Could you add a JSDoc to explain these config options? It would help the developer to understand when to provide a Provider and when it's okay to omit that value. Thanks!

packages/angular/projects/angular-sdk/src/test/test.utils.ts
```diff
@@ -6,11 +6,15 @@ export class TestingProvider extends InMemoryProvider {
     private delay: number,
   ) {
     super(flagConfiguration);
+    if (!delay) {
+      Object.assign(this, { initialize: undefined });
```
beeme1mr, line 11: This is clever. I know why you did this but future me may not. Could you please add a comment explaining why this was necessary?

packages/angular/projects/angular-sdk/README.md
```diff
@@ -282,6 +289,86 @@ This can be used to just render the flag value or details without conditional re
...
+@Component({
+  selector: 'my-component',
+  standalone: true,
+  imports: [AsyncPipe],
+  template: `
+    <div *ngIf="(isFeatureEnabled$ | async)?.value">
+      Feature is enabled! Reason: {{ (isFeatureEnabled$ | async)?.reason }}
+    </div>
+    <div>Theme: {{ (currentTheme$ | async)?.value }}</div>
+    <div>Max items: {{ (maxItems$ | async)?.value }}</div>
```
beeme1mr, line 313: Why do you need the `?`? Won't value always be there once the flag been evaluated?

packages/angular/projects/angular-sdk/src/lib/feature-flag.service.ts
lukas-reining: This has the same effect as returning from a `useEffect`  in React. 
When the observable in unsubscribed, this will be called (so when it is not needed anymore).

packages/angular/projects/angular-sdk/README.md
lukas-reining: Yes you are right :)

packages/angular/projects/angular-sdk/src/lib/internal/is-equal.ts
lukas-reining: Mhh good question, would make sense but I am not sure if it is worth it.

beeme1mr: Yeah, I agree. Let's leave it as is for now.

PR #1251, opened 2025-09-29
fix: angular tests in main

toddbaert:
@lukas-reining this fixes the ongoing test compilation issue we are seeing in main.

The reason you didn't see it locally is because if we have a compiled `dist/` in `../web` (the local version of `@openfeature/web-sdk`) it's resolved, but the main build doesn't have this at the point the tests run. 

I could have fixed this by adding a compile step, but I think it's always better to run the tests against `src/` of our local peers instead of `dist/` (or else things get really weird and you can sometimes not realize you're testing against stale code)

This PR fixes this issue (confirmed by deleting the `../web` dist locally, seeing the same errors, and then applying this fix.


packages/angular/projects/angular-sdk/tsconfig.spec.json
```diff
@@ -6,13 +6,19 @@
...
       "node"
     ],
+    "baseUrl": "../",
     "paths": {
       "angular": [
         "./dist/angular"
-      ]
+      ],
+      "@openfeature/core": [ "../../shared/src" ],
+      "@openfeature/web-sdk": [ "../../web/src" ]
```
toddbaert, line 15: This is the fix.

packages/angular/projects/angular-sdk/tsconfig.spec.json
```diff
@@ -6,13 +6,19 @@
...
     },
     "esModuleInterop": true,
-    "emitDecoratorMetadata": true
+    "emitDecoratorMetadata": true,
+    "noImplicitOverride": false,
+    "noPropertyAccessFromIndexSignature": false,
+    "strict": false
```
toddbaert, line 21: I had to add these, because now the ng-test step compiles some code from the shared packages, and doesn't tolerate some stuff there without these.

packages/web/src/open-feature.ts
```diff
@@ -393,7 +393,7 @@ export class OpenFeatureAPI
         const maybePromise = wrapper.provider.onContextChange(oldContext, newContext);
 
         // only reconcile if the onContextChange method returns a promise
-        if (typeof maybePromise?.then === 'function') {
+        if (typeof (maybePromise as Promise<void>)?.then === 'function') {
```
toddbaert, line 396: This is the only thing the ng-compiler was complaining about I could not ignore - obviously there's no runtime change here, it just makes the compiler happy.

packages/angular/projects/angular-sdk/tsconfig.spec.json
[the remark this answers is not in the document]
toddbaert: This is true, but this is only for test code (`tsconfig.spec.json`); without this, we'd have to make a bunch of changes to `@openfeature/core` and `@openfeature/web` just to satisfy it's some of it's misconceptions.

packages/web/src/open-feature.ts
```diff
@@ -393,7 +393,7 @@ export class OpenFeatureAPI
         const maybePromise = wrapper.provider.onContextChange(oldContext, newContext);
 
         // only reconcile if the onContextChange method returns a promise
-        if (typeof maybePromise?.then === 'function') {
+        if (maybePromise && typeof maybePromise?.then === 'function') {
```
toddbaert, line 396: This is the only thing the ng-compiler was complaining about I could not ignore - obviously there's no real runtime change here, it just makes the compiler happy.

lukas-reining (approved):

PR #1256, opened 2025-09-30
chore: update NPM to ^11.5.1, switch to OIDC NPM Auth

jonathannorris:
## This PR
- updates the NPM version to `^11.5.1` needed for OIDC NPM auth: https://docs.npmjs.com/trusted-publishers/#github-actions-configuration
- Switches from NPM tokens to NPM OIDC auth so we can disable NPM publish tokens.
- updates the base node version in actions to node 22 (current LTS)

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -31,6 +31,7 @@ jobs:
     continue-on-error: true
     permissions:
       id-token: write
+      attestations: write
       contents: read
     strategy:
       matrix:


.github/workflows/release-please.yml
```diff
@@ -27,28 +29,30 @@ jobs:
...
-        run: |
-          npm install -g npm@^10.2.0
-          npm ci
-          npm sbom --sbom-format=cyclonedx --omit=dev --omit=peer --workspace=${{matrix.release}} > bom.json
-      - name: Attach SBOM to artifact
-        env:
-          GITHUB_TOKEN: ${{secrets.RELEASE_PLEASE_ACTION_TOKEN}}
-        run:
-          gh release upload $TAG bom.json
+        run: npm sbom --sbom-format=cyclonedx --omit=dev --omit=peer --workspace=${{ matrix.release }} > sbom.json
```
jonathannorris, line 51: @beeme1mr updated to the same SBOM attestation as we did in MCP: https://github.com/open-feature/mcp/blob/main/.github/workflows/release-please.yml#L56-L63

.github/workflows/audit-pending-releases.yml
```diff
@@ -29,6 +29,6 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 22
```
beeme1mr: Should we add a `.nvmrc` instead of manually updating these values? We do this in a few other repos but apparently not here. Here's an example from the website ([.nvmrc](https://github.com/open-feature/openfeature.dev/blob/main/.nvmrc), [referencing in an action](https://github.com/open-feature/openfeature.dev/blob/main/.github/workflows/pr-checks.yaml#L21))

I can do this as a follow up if you prefer.

.github/workflows/release-please.yml
```diff
@@ -27,28 +29,30 @@ jobs:
     if: ${{ fromJSON(needs.release-please.outputs.paths_released)[0] != null }}
     # Continues with the release process even if SBOM generation fails.
     continue-on-error: true
+    permissions:
+      id-token: write
+      contents: read
```
beeme1mr, line 35: I think we also need `attestations: write`

beeme1mr: https://github.com/actions/attest-sbom

.github/workflows/audit-pending-releases.yml
jonathannorris: yea I use `.nvmrc` in all our repos, would be happy to switch to that. It will kinda set the `development` version of node that you should be using, but we can set the min version lower.

Issue #1236, opened 2025-09-12
[BUG] inconsitent behaviour in server side hookcontext data for evaluation context


aepfli:
### Observed behavior

take a close look at the code, comparing java and js there is a difference in how the hookcontext is build/enriched with evalutation context.

js-code: https://github.com/open-feature/js-sdk/blob/main/packages/server/src/client/internal/open-feature-client.ts#L255-L327

java-code: https://github.com/open-feature/java-sdk/blob/a08ff4d96ccf9236ad882892280b600a05409394/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java#L160-L231

In java the evaluation context is everytime updated, when it was changed (due to hooks, etc). in js it is only set a the beginning with the merged context, but does not update with the frozenContext from the beforehooks.

This leads to the problem that in the after, finally and error hooks the information in the hookcontext is not up to date.

### Expected Behavior

We should solve this in a way that this is backwards compatible and we always have the full context available

MarkoMlakar:
Hello :) 
I have also been looking into this issue and wanted to share my thoughts. I agree with @beeme1mr. By the time after/error/finally hooks run, they already have access to the complete evaluation context.
There are also passing unit tests ([client](https://github.com/open-feature/js-sdk/blob/main/packages/server/test/client.spec.ts#L757-L786), [hooks](https://github.com/open-feature/js-sdk/blob/main/packages/server/test/hooks.spec.ts#L244-L324)) which would likely fail if the merging logic was not working properly.

aepfli:
can we close this issue? @MarkoMlakar

PR #1257, opened 2025-10-01
refactor: simplify evaluation context updates for before hooks

MarkoMlakar:
## This PR
- Remove redundant loop for updating the before hook contexts
- Add simple test case to make sure the same object reference is being updated

### Related Issues

#1236

### Notes
The accumulated evaluation context object is shared with all the hooks (same object reference). This makes the inner for loop for updating the contexts redundant

### How to test
```npm run test```

### DIFF
--- a/packages/server/src/client/internal/open-feature-client.ts
+++ b/packages/server/src/client/internal/open-feature-client.ts
@@ -338,20 +338,16 @@ export class OpenFeatureClient implements Client {
       const hookContextIndex = hooks.length - 1 - index; // reverse index for before hooks
       const hookContext = hookContexts[hookContextIndex];
 
-      // Update the context on the stable hook context object
-      Object.assign(hookContext.context, accumulatedContext);
-
       const hookResult = await hook?.before?.(hookContext, Object.freeze(options.hookHints));
       if (hookResult) {
         accumulatedContext = {
           ...accumulatedContext,
           ...hookResult,
         };
-
-        for (let i = 0; i < hooks.length; i++) {
-          Object.assign(hookContexts[hookContextIndex].context, accumulatedContext);
-        }
       }
+
+      // Update the context on the stable hook context object
+      Object.assign(hookContext.context, accumulatedContext);
     }
 
     // after before hooks, freeze the EvaluationContext.
--- a/packages/server/test/hooks.spec.ts
+++ b/packages/server/test/hooks.spec.ts
@@ -321,6 +321,45 @@ describe('Hooks', () => {
         expect.anything(),
       );
     });
+    it('Should share the same context object reference across all hooks', (done) => {
+      let hook1Context: EvaluationContext;
+      let hook2Context: EvaluationContext;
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by aepfli — APPROVED** [review]
thank you this looks good to me - but i am by far not a js expert, maybe somebody else can have a look at this

File: packages/server/src/client/internal/open-feature-client.ts [hunk 1]
```diff
@@ -338,20 +338,16 @@ export class OpenFeatureClient implements Client {
       const hookContextIndex = hooks.length - 1 - index; // reverse index for before hooks
       const hookContext = hookContexts[hookContextIndex];
 
-      // Update the context on the stable hook context object
-      Object.assign(hookContext.context, accumulatedContext);
```
guidobrei, line 342: If we remove this line, the `accumulatedContext` is not available in the `beforeHook` on the next line.

MarkoMlakar: Thanks @guidobrei you are correct. I have updated the PR

packages/server/src/client/internal/open-feature-client.ts
```diff
@@ -344,10 +344,7 @@ export class OpenFeatureClient implements Client {
       const hookResult = await hook?.before?.(hookContext, Object.freeze(options.hookHints));
       if (hookResult) {
         Object.assign(accumulatedContext, hookResult);
-
-        for (let i = 0; i < hooks.length; i++) {
-          Object.assign(hookContexts[hookContextIndex].context, accumulatedContext);
```
MarkoMlakar, line 349: hookContext.context is a shared object. All the hooks recieve and operate on the same context object, making this for loop redundant

guidobrei: Is it? from the method signature it could be different contexts.

toddbaert: I think his new test basically proves it!

packages/server/test/hooks.spec.ts
```diff
@@ -311,6 +311,45 @@ describe('Hooks', () => {
         expect.anything(),
       );
     });
+    it('Should share the same context object reference across all hooks', (done) => {
+      let hook1Context: EvaluationContext;
+      let hook2Context: EvaluationContext;
+      let hook3Context: EvaluationContext;
+
+      client.getBooleanValue(FLAG_KEY, false, undefined, {
```
toddbaert: Maybe we could also add some API/client level hooks and make sure they also use the same.

lukas-reining: This would be great!

MarkoMlakar: Thanks for the feedback:) I have updated the test case to check context with API, client and invocation hooks in the latest [commit](https://github.com/open-feature/js-sdk/pull/1257/changes/3e69e0ea1624c5e2ec46bab18646022d92bb9628)

PR #1259, opened 2025-10-16
chore: enable release-as=patch for release-please to trigger patch release for all packages

jonathannorris:
Enabling` "release-as": "patch"` in the `release-please-config.json` config per-package will trigger patch releases for each package, allowing us to test NPM OIDC connections with a patch release.

This will be removed after a successful deployment.


beeme1mr (changes requested):
I don't believe this is correct. I think you need to set `release-as` per package and use the version you want released.

toddbaert (approved):

PR #1280, opened 2025-11-05
feat: add timeout with cancellation to requests

leoromanovsky:
## This PR

- adds this new feature

### Related Issues

Fixes #1234523

### DIFF
--- a/packages/server/src/open-feature.ts
+++ b/packages/server/src/open-feature.ts
@@ -10,6 +10,7 @@ import {
 } from '@openfeature/core';
 import type { Client } from './client';
 import { OpenFeatureClient } from './client/internal/open-feature-client';
+import type { ProviderWaitOptions } from './evaluation';
 import { OpenFeatureEventEmitter } from './events';
 import type { Hook } from './hooks';
 import type { Provider} from './provider';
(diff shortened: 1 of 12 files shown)


packages/server/src/open-feature.ts
```diff
@@ -95,13 +106,41 @@ export class OpenFeatureAPI
...
+    if (domain) {
+      // Domain is specified: setProviderAndWait(domain, provider, options?)
+      provider = objectOrUndefined<Provider>(providerOrOptions);
+      options = objectOrUndefined<ProviderWaitOptions>(optionsOrUndefined);
+    } else {
+      // No domain: setProviderAndWait(provider, options?)
+      provider = objectOrUndefined<Provider>(domainOrProvider);
+
+      // Check if providerOrOptions is actually options
+      if (providerOrOptions && !('metadata' in providerOrOptions)) {
```
beeme1mr, line 138: I'd like to find a more reliable way to test if it's a provider.

packages/shared/src/open-feature.ts
```diff
@@ -440,4 +456,42 @@ export abstract class OpenFeatureCommonAPI<
     this._logger.error(`Error during shutdown of provider ${provider.metadata.name}: ${err}`);
     this._logger.error((err as Error)?.stack);
   }
+
+  private async withProviderTimeoutAndCancellation(
```
beeme1mr, line 460: Should the provider continue to attempt to start or should we completely abort? If we want to abort, the error should be provider fatal and we may want to explicitly call the provider shutdown method.

@toddbaert @lukas-reining any thoughts on this?

beeme1mr (commented):
Hey @leoromanovsky, looks good so far. I just left a few thoughts.

toddbaert:
Hey @leoromanovsky - we discussed this in the community meeting today.

I understand the motivation for this, but we've tried to avoid this in the past. The reason is because this is already something the provider can control themselves. Many providers provide options, for example in their constructors, that allow users to specify initialization timeouts. The provider can then simply throw (or otherwise abnormally terminate) to signal an error during their [initialization phase](https://openfeature.dev/specification/sections/providers/#24-initialization), and the SDK will react accordingly by throwing where appropriate, and running error handlers (this is what flagd does, for example). On the other hand, some providers don't require initialization at all, and are ready as soon as they are instantiated, making this option irrelevant for them.

For this reason, we'd like to avoid adding this to the spec or SDKs.

I do think adding abort signals is fine though; that's a different thing than a timeout option, more equivalent to the `context.Context` in the Go SDK.

cc @lukas-reining @jonathannorris @kinyoklion @thomaspoignant

lukas-reining:
> I do think adding abort signals is fine though; that's a different thing than a timeout option, more equivalent to the context.Context in the Go SDK.

How would you add them @toddbaert? Do you see them as part of the spec? 
I would not like to have them in there either. To me, this is very specific for the provider. A provider having this seems totally valid for me, but a timeout too. 


@leoromanovsky I might have failed to find it but do you have an explanation of the use-case of the feature? 
As @toddbaert says, this is already something that can be implemented by the provider using timeouts and also signals if it makes sense for that specific flag source. The SDK handles a failing initialization already. 
Is the motivation that this is missing for a specific provider? Or is there a generic use-case where you need this?

PR #1282, opened 2025-11-13
fix: multi-provider hook context management

### LINKED ISSUE #1268: [BUG] Multi Provider: hookContexts.get(context) fails, but works with context.context
Hi,
I ran into an issue when using the Multi Provider.
These two lines don’t work for me as-is ([link](  https://github.com/open-feature/js-sdk/blob/69230e4cf66e215ea6a7f12651ab9527fbfb79bb/packages/server/src/provider/multi-provider/multi-provider.ts#L143-L144)):
```
const hookContext = this.hookContexts.get(context);
const hookHints = this.hookHints.get(context);
```
But when I change them to this, it works:
```
const hookContext = this.hookContexts.get(context.context);
const hookHints = this.hookHints.get(context.context);
```
Has anyone else experienced this?
I’m using `FlagdProvider` and the `EnvVarProvider` provider.
Thanks!

mdkitzman:
## This PR
Fixing an issue with the MultiProvider where hook contexts and hints were being lost due to copies of the context data being created in the OpenFeature sdk evaluation.

Since key evaluation of Maps using objects is done by reference, the lookup of the context during evaluation was failing, leading to errors.

- adds this new feature

### Related Issues

Fixes #1268

### DIFF
--- a/packages/server/src/provider/multi-provider/multi-provider.ts
+++ b/packages/server/src/provider/multi-provider/multi-provider.ts
@@ -30,8 +30,9 @@ export class MultiProvider implements Provider {
 
   public readonly events = new OpenFeatureEventEmitter();
 
-  private hookContents: Array<[HookContext, HookHints]> = [];
-  
+  private hookContexts: WeakMap<EvaluationContext, HookContext> = new WeakMap<EvaluationContext, HookContext>();
+  private hookHints: WeakMap<EvaluationContext, HookHints> = new WeakMap<EvaluationContext, HookHints>();
+
   metadata: ProviderMetadata;
 
   providerEntries: RegisteredProvider[] = [];
@@ -139,8 +140,9 @@ export class MultiProvider implements Provider {
     defaultValue: T,
     context: EvaluationContext,
   ): Promise<ResolutionDetails<T>> {
-    const [hookContext, hookHints] = this.hookContents.shift() ?? [];
-    
+    const hookContext = this.hookContexts.get(context);
+    const hookHints = this.hookHints.get(context);
+
     if (!hookContext || !hookHints) {
       throw new GeneralError('Hook context not available for evaluation');
     }
@@ -297,7 +299,8 @@ export class MultiProvider implements Provider {
     return [
       {
         before: async (hookContext: BeforeHookContext, hints: HookHints): Promise<EvaluationContext> => {
-          this.hookContents.push([hookContext, hints ?? {}]);
+          this.hookContexts.set(hookContext.context, hookContext);
+          this.hookHints.set(hookContext.context, hints ?? {});
           return hookContext.context;
         },
       },
--- a/packages/server/src/client/internal/open-feature-client.ts
+++ b/packages/server/src/client/internal/open-feature-client.ts
@@ -332,7 +332,7 @@ export class OpenFeatureClient implements Client {
     mergedContext: EvaluationContext,
     options: FlagEvaluationOptions,
   ) {
-    let accumulatedContext = mergedContext;
+    const accumulatedContext = mergedContext;
 
     for (const [index, hook] of hooks.entries()) {
       const hookContextIndex = hooks.length - 1 - index; // reverse index for before hooks
@@ -343,10 +343,7 @@ export class OpenFeatureClient implements Client {
 
       const hookResult = await hook?.before?.(hookContext, Object.freeze(options.hookHints));
       if (hookResult) {
-        accumulatedContext = {
-          ...accumulatedContext,
-          ...hookResult,
-        };
+        Object.assign(accumulatedContext, hookResult);
 
         for (let i = 0; i < hooks.length; i++) {
           Object.assign(hookContexts[hookContextIndex].context, accumulatedContext);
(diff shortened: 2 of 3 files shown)


packages/server/src/provider/multi-provider/multi-provider.ts
```diff
@@ -30,9 +30,8 @@ export class MultiProvider implements Provider {
 
   public readonly events = new OpenFeatureEventEmitter();
 
-  private hookContexts: WeakMap<EvaluationContext, HookContext> = new WeakMap<EvaluationContext, HookContext>();
-  private hookHints: WeakMap<EvaluationContext, HookHints> = new WeakMap<EvaluationContext, HookHints>();
-
+  private hookContents: Array<[HookContext, HookHints]> = [];
```
jonathannorris: from reviewing the code, I believe the switch from the `WeakMap` to `Array` should be ok. However, this does mean that we are now keeping strong references to the `HookContext`, whereas we previously kept weak references to `EvaluationContext`. @toddbaert / @lukas-reining any thoughts on this?

Here is an AI analysis of the change from `WeakMap`: 

## Memory Analysis

**WeakMap approach:**
- Keys (`EvaluationContext`) are weakly held — auto-removed when GC'd
- Values (`HookContext`, `HookHints`) are strongly held — kept until manually removed or the WeakMap is cleared

**Array approach:**
- `HookContext` and `HookHints` are strongly held in the array
- Removed immediately when `shift()` is called (line 142, first line of `flagResolutionProxy`)

## Memory Impact

1. Strong references: Both approaches strongly hold `HookContext` and `HookHints`; the difference is minimal.
2. Cleanup: The array removes items immediately via `shift()`, so they're released right after use.
3. Edge case: If `before` runs but `flagResolutionProxy` never runs (e.g., client error), an item could remain. This is unlikely since they're in the same async chain.

## Verdict

No significant memory concern. The array approach:
- Removes items immediately via `shift()` at the start of `flagResolutionProxy`
- Doesn't accumulate entries in normal operation
- Has similar memory characteristics since both hold strong references to the hook data

The WeakMap's auto-cleanup on key GC isn't needed here because items are consumed immediately. The array approach is simpler and has equivalent memory behavior for this use case.

beeme1mr: It's probably better to avoid a weakmap if it isn't required.

mdkitzman:
> Could a test be added to verify the issue has been resolved?

Yeah, let me take a look at adding a test for this.

packages/server/src/provider/multi-provider/multi-provider.ts
mdkitzman: After looking at this and thinking about how to test it, I realized that perhaps this wasn't an issue with the use of the WeakMap in the MultiProvider, but rather in how the Open Feature SDK builds the context for evaluation of the hooks.

In the OpenFeature client, it attempts to accumulate the hook contexts into the `accumulatedContext`, but for each hook evaluation, it is re-assigning it to a new object
https://github.com/open-feature/js-sdk/blob/main/packages/server/src/client/internal/open-feature-client.ts#L346-L349

and the [comment where this method is invoked](https://github.com/open-feature/js-sdk/blob/main/packages/server/src/client/internal/open-feature-client.ts#L280-L299) seems to indicate that the expectation is that providers can use a WeakMap to keep a reference to the context.

I think I might take a slightly different approach and change the OpenFeature Client code.

packages/server/src/provider/multi-provider/multi-provider.ts
```diff
@@ -140,9 +139,8 @@ export class MultiProvider implements Provider {
     defaultValue: T,
     context: EvaluationContext,
   ): Promise<ResolutionDetails<T>> {
-    const hookContext = this.hookContexts.get(context);
-    const hookHints = this.hookHints.get(context);
-
+    const [hookContext, hookHints] = this.hookContents.shift() ?? [];
```
lukas-reining: I think this could be a problem. 
The contents are pushed async to `hookContents` and then read async. 
For one evaluation this should be okay when providers are evaluated sequentially. 
But if the providers are evaluated in `parallel` (concurrently), or if two evaluations are done in parallel, I think `hookContents` could get out of order and we have a race condition. 
Or am I overlooking anything @beeme1mr @jonathannorris @mdkitzman?

packages/server/src/provider/multi-provider/multi-provider.ts
lukas-reining: I think the weak map should not be the problem. 
But there might be another problem: https://github.com/open-feature/js-sdk/pull/1282/files#r2538551399

mdkitzman: I think I might have a simpler approach where this would be addressed in the open feature client code, and the multi provider can continue to utilize the weak map to store and retrieve these references.  At that point, I think we would eliminate any concern about a race condition.

mdkitzman: I think I have narrowed the issue down to how the hook context is accumulated prior to evaluation in the `open-feature-client` combined with TypeScript transpilation.

I was unable to successfully author a test to replicate the issue with the existing Jest test suite, and decided to try and replicate my test in another project using the transpiled library.  Turns out the issue exists with the transpiled code and not when executing the typescript code with Jest.

mdkitzman: I was able to replicate the issue in typescript, so (thankfully) it wasn't a transpilation issue

beeme1mr (approved):
Thanks!

PR #1287, opened 2025-11-21
fix: compare full EvaluationDetails to prevent stale data

### LINKED ISSUE #1286: [BUG] useFlag hooks display stale EvaluationDetails when flag value unchanged but metadata differs
### Observed behavior

When a flag is re-evaluated and the value remains the same but other EvaluationDetails fields change (e.g., reason, errorCode, variant, flagMetadata), components using useFlag or related hooks continue to display the old EvaluationDetails instead of the updated metadata.

### Expected Behavior

Components should re-render and display updated EvaluationDetails whenever any field changes, not just the value.

### Steps to reproduce

1. Set up a React component using useBooleanFlagDetails (or similar hook)
2. Mount the component while the provider is NOT_READY (flag evaluates with reason: "ERROR", errorCode: "PROVIDER_NOT_READY")
3. Wait for the provider to become READY (same flag value but now reason: "STATIC", no error code)
4. Observe that the component's EvaluationDetails still shows reason: "ERROR" instead of reason: "STATIC"

beeme1mr:
## This PR

- updated the `isEqual` function to compare any input
- updated the compare to evaluate the whole EvaluationDetails object instead of just the value 

### Related Issues

Fixes #1286

### Notes

This may trigger slightly more re-renders but I feel that that's better than return stale EvaluationDetails data which has proven to be confusing to folks attempting to debug unexpected evaluations.

### DIFF
--- a/packages/react/src/internal/is-equal.ts
+++ b/packages/react/src/internal/is-equal.ts
@@ -1,6 +1,9 @@
 /**
  * Deeply compare two values to determine if they are equal.
  * Supports primitives and serializable objects.
+ * 
+ * Note: Does not handle Date, RegExp, Map, Set, or circular references.
+ * Suitable for comparing EvaluationDetails and other JSON-serializable data.
  * @param {unknown} value First value to compare
  * @param {unknown} other Second value to compare
  * @returns {boolean} True if the values are equal


packages/react/test/evaluation.spec.tsx
```diff
@@ -585,6 +585,79 @@ describe('evaluation', () => {
       return new TestingProvider(CONFIG, DELAY); // delay init by 100ms
     };
 
+    describe('provider ready event updates', () => {
+      it('should update EvaluationDetails when provider becomes ready (including reason)', async () => {
```
beeme1mr, line 589: This test failed before I updated the compare.

beeme1mr:
> However, I've identified a potential issue in the updated `isEqual` function where it doesn't distinguish between arrays and objects with numeric keys, which could lead to missed updates in some edge cases. I've left a suggestion to make it more robust.

I think these acceptable limitations.

packages/react/src/internal/is-equal.ts
```diff
@@ -1,13 +1,14 @@
-import { type FlagValue } from '@openfeature/web-sdk';
-
 /**
  * Deeply compare two values to determine if they are equal.
  * Supports primitives and serializable objects.
- * @param {FlagValue} value First value to compare
- * @param {FlagValue} other Second value to compare
+ * 
+ * Note: Does not handle Date, RegExp, Map, Set, or circular references.
+ * Suitable for comparing EvaluationDetails and other JSON-serializable data.
```
toddbaert, line 6: That's a good callout.

PR #1291, opened 2025-11-21
fix: in-line docs, un-needed AbortController creation

toddbaert:
This PR:

- clarifies some doc around `useWhenProviderReady` (it really only works with READY)
- moves AbortController instantiation so we don't create it every render if we don't need to


lukas-reining (approved):
That makes much sense!

PR #1300, opened 2025-11-25
feat: allow validating context to skip initialization + context changes

### LINKED ISSUE #1279: [FEATURE] Ability to suspend until provider context ready
### Requirements

We want to be able to use the React hooks to suspend the app until we've loaded an async context and the provider has evaluated flags with that context (loaded from a hook and passed in via useContextMutator).

If we do not set a provider immediately, the default provider in the SDK (noop) does not trigger the suspend, and if we're in a non-default domain and a provider has been set on the default domain, that will also not suspend.

So, we have to set the provider immediately, but then it will try to evaluate the flags without the context as we don't have that yet, and so won't remain suspended, instead becoming ready with the fallback values for any flag hooks.

Our current workaround is to wrap a provider with logic that blocks the initialization until we call the context change, at which point we pass that context back to the paused initialization and allow it to complete before returning from the context change (without passing the context change to the underlying provider).

This allows for the provider to sit in the NOT_READY state until we pass in a context, at which point it flips into RECONCILING as we have an async context change method (waiting for the async initialization to complete), and then to READY.

```ts
// Wrap an OpenFeature provider to ensure it waits for the full context to be available before initializing.
// This allows for us to use the suspend functionality of OpenFeature to delay rendering...
//  ...until our flags have been evaluated with the full context, rather than using defaults.
function createContextualProvider<
  T extends new (...args: any[]) => Provider, // eslint-disable-line @typescript-eslint/no-explicit-any
>(ProviderClass: T): new (...args: ConstructorParameters<T>) => Provider {
  return class ContextualProvider extends ProviderClass implements Provider {
    private contextPromise: Promise<EvaluationContext> | undefined = undefined;
    private contextResolve: ((context: EvaluationContext) => void) | undefined = undefined;

    private initializePromise: Promise<void> | undefined = undefined;
    private initializeResolve: (() => void) | undefined = undefined;

    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    constructor(...args: any[]) {
      super(...args);
    }

    async initialize(context?: EvaluationContext) {
      // If called a second time for some reason, wait until the initial initialization has completed
      if (this.contextPromise) {
        await this.contextPromise;
        return super.initialize?.(context);
      }

      // Otherwise, disregard any initial context passed in, and wait for the full context to be provided via `onContextChange`
      this.contextPromise = new Promise<EvaluationContext>((resolve) => {
        this.contextResolve = resolve;
      });
      this.initializePromise = new Promise<void>((resolve) => {
        this.initializeResolve = resolve;
      });
      const data = await this.contextPromise;

      // With the full context provided, initialize the provider using it, not the initial context passed in
      await super.initialize?.(data);
      this.initializeResolve?.();
    }

    onContextChange(oldContext: EvaluationContext, newContext: EvaluationContext) {
      // If we've already done the initial wait for a full context, pass through to base implementation
      if (!this.contextResolve) {
        return super.onContextChange?.(oldContext, newContext);
      }

      // Otherwise, allow the initialization to proceed with the full context we've now got, and do not call the base implementation
      // But, leave the promise around in case initialize is called again for some reason, so we know not to wait again
      this.contextResolve(newContext);
      this.contextResolve = undefined;

      // Wait until initialization has completed before considering the context change complete
      // Returning a promise here puts OpenFeature into RECONCILING and ensures it waits
      return Promise.resolve(this.initializePromise).then(() => {
        this.initializePromise = undefined;
        this.initializeResolve = undefined;
      });
    }

    async onClose() {
      this.contextPromise = undefined;
      this.contextResolve = undefined;
      this.initializePromise = undefined;
      this.initializeResolve = undefined;

      return super.onClose?.();
    }
  };
}
```

However, as you might be able to see, this is a bit janky and it would definitely be nice for a solution to this in some form to exist in the SDK itself.

MattIPv4:
## This PR

Adds a new options argument when calling `setProvider` and `setProviderAndWait` in the web SDK, allowing for a context validation method to be provided. This method can cause the initialization of the provider to be skipped when setting a new provider, leaving it NOT_READY, until a future valid context change is made. It can also cause context change reconciliation to be skipped.

### Related Issues

Resolves #1279

### Notes

While returning false during context validation will prevent a provider from reconciling context changes, the updated context is still stored and persisted.

### Follow-up Tasks

N/A

### How to test

See test suite :)

### DIFF
--- a/packages/shared/src/open-feature.ts
+++ b/packages/shared/src/open-feature.ts
@@ -228,13 +228,15 @@ export abstract class OpenFeatureCommonAPI<
   abstract setProviderAndWait(
     clientOrProvider?: string | P,
     providerContextOrUndefined?: P | EvaluationContext,
-    contextOrUndefined?: EvaluationContext,
+    contextOptionsOrUndefined?: EvaluationContext | Record<never, never>,
+    optionsOrUndefined?: Record<never, never>,
   ): Promise<void>;
 
   abstract setProvider(
     clientOrProvider?: string | P,
     providerContextOrUndefined?: P | EvaluationContext,
-    contextOrUndefined?: EvaluationContext,
+    contextOptionsOrUndefined?: EvaluationContext | Record<never, never>,
+    optionsOrUndefined?: Record<never, never>,
   ): this;
 
   protected initializeProviderForDomain(
@@ -294,7 +296,11 @@ export abstract class OpenFeatureCommonAPI<
       });
   }
 
-  protected setAwaitableProvider(domainOrProvider?: string | P, providerOrUndefined?: P): Promise<void> | void {
+  protected setAwaitableProvider(
+    domainOrProvider?: string | P,
+    providerOrUndefined?: P,
+    skipInitialization = false,
+  ): Promise<void> | void {
     const domain = stringOrUndefined(domainOrProvider);
     const provider = objectOrUndefined<P>(domainOrProvider) ?? objectOrUndefined<P>(providerOrUndefined);
 
@@ -327,17 +333,19 @@ export abstract class OpenFeatureCommonAPI<
       this._statusEnumType,
     );
 
-    // initialize the provider if it's not already registered and it implements "initialize"
-    if (!this.allProviders.includes(provider)) {
-      initializationPromise = this.initializeProviderForDomain(wrappedProvider, domain);
-    }
+    if (!skipInitialization) {
+      // initialize the provider if it's not already registered and it implements "initialize"
+      if (!this.allProviders.includes(provider)) {
+        initializationPromise = this.initializeProviderForDomain(wrappedProvider, domain);
+      }
 
-    if (!initializationPromise) {
-      wrappedProvider.status = this._statusEnumType.READY;
-      emitters.forEach((emitter) => {
-        emitter?.emit(AllProviderEvents.Ready, { clientName: domain, domain, providerName });
-      });
-      this._apiEmitter?.emit(AllProviderEvents.Ready, { clientName: domain, domain, providerName });
+      if (!initializationPromise) {
+        wrappedProvider.status = this._statusEnumType.READY;
+        emitters.forEach((emitter) => {
+          emitter?.emit(AllProviderEvents.Ready, { clientName: domain, domain, providerName });
+        });
+        this._apiEmitter?.emit(AllProviderEvents.Ready, { clientName: domain, domain, providerName });
+      }
     }
 
     if (domain) {
--- a/packages/shared/src/open-feature.ts
+++ b/packages/shared/src/open-feature.ts
@@ -27,6 +27,8 @@ type AnyProviderStatus = ClientProviderStatus | ServerProviderStatus;
  */
 export class ProviderWrapper<P extends CommonProvider<AnyProviderStatus>, S extends AnyProviderStatus> {
   private _pendingContextChanges = 0;
+  private _initializing = false;
+  private _initialized = false;
 
   constructor(
     private _provider: P,
@@ -49,6 +51,8 @@ export class ProviderWrapper<P extends CommonProvider<AnyProviderStatus>, S exte
         this._status = _statusEnumType.ERROR as S;
       }
     });
+
+    this._initialized = !(typeof _provider.initialize === 'function');
   }
 
   get provider(): P {
@@ -65,6 +69,22 @@ export class ProviderWrapper<P extends CommonProvider<AnyProviderStatus>, S exte
 
   set status(status: S) {
     this._status = status;
+  }
+
+  get initializing() {
+    return this._initializing;
+  }
+
+  set initializing(initializing: boolean) {
+    this._initializing = initializing;
+  }
+
+  get initialized() {
+    return this._initialized;
+  }
+
+  set initialized(initialized: boolean) {
+    this._initialized = initialized;
   }
 
   get allContextChangesSettled() {
@@ -221,10 +241,11 @@ export abstract class OpenFeatureCommonAPI<
     wrapper: ProviderWrapper<P, AnyProviderStatus>,
     domain?: string,
   ): Promise<void> | void {
-    if (typeof wrapper.provider.initialize !== 'function') {
+    if (typeof wrapper.provider.initialize !== 'function' || wrapper.initializing || wrapper.initialized) {
       return;
     }
 
+    wrapper.initializing = true;
     return wrapper.provider
       .initialize(domain ? (this._domainScopedContext.get(domain) ?? this._context) : this._context)
       .then(() => {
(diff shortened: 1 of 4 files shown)


packages/web/src/open-feature.ts
```diff
@@ -185,7 +268,28 @@ export class OpenFeatureAPI
       }
     }
 
-    const maybePromise = this.setAwaitableProvider(domain, provider);
+    const maybePromise = this.setAwaitableProvider(domain, provider, skipInitialization);
+
+    // If there was a validation error with the context, move the newly created provider to ERROR state.
+    // We know we've skipped initialization if this happens, so no need to worry about the promise changing the state later.
```
lukas-reining, line 274: The spec says [[1]](https://openfeature.dev/specification/sections/flag-evaluation#requirement-174): 

```
The client's provider status accessor MUST indicate ERROR if the initialize function of the associated provider terminates abnormally.
```

The non-normative part says: 

```
If the provider has failed to initialize, the provider status should indicate the provider is in an error state.
```

So I think it is acceptable to set the state to `ERROR` if the initialize has not been run at all due to invalid context. But I am not 100% sure @toddbaert.

packages/web/src/open-feature.ts
```diff
@@ -380,6 +484,26 @@ export class OpenFeatureAPI
...
     try {
+      // validate the context to decide if we should run the context change handler.
+      // notably, the stored context will still be updated by this point, this just skips reconciliation.
+      const options = this.getProviderOptions(domain);
+      if (typeof options.validateContext === 'function') {
+        if (!options.validateContext(newContext)) {
+          this._logger.debug(
+            `Skipping context change for domain '${domain ?? 'default'}' due to validateContext returning false.`,
+          );
+          return;
```
lukas-reining, line 495: Returning here makes `provider.onContextChange` not being called. 
While this makes sense if the context is invalid, I think this is not conformant to the spec. 
The spec says [[1]](https://openfeature.dev/specification/sections/evaluation-context#conditional-requirement-3241): 

```
> When the global evaluation context is set, the on context changed function MUST run.

The SDK implementation must run the on context changed function on all registered providers that use the global evaluation context whenever it is mutated.
```

With this implementation, the context is still mutated [[2]](https://github.com/open-feature/js-sdk/pull/1300/changes#diff-7c5a921fafc79ca3f282d3fa80a5981dad0e7774577a631dcc22e7e8f84c1fd6R343) but the handler is not run.
While this is logical this is unexpected regarding the spec.

My feeling is we would need to add it to the spec before implementing it this way but I could be on the wrong track @open-feature/technical-steering-committee

lukas-reining: Not mutating the context would be a problem too. 
In that case it could happen that `validateContext` is set for the default domain. If there was a context set on the default domain which is invalid for the default domain, that is also expected to be used by a domain scoped provider, the `validateContext` of the default context could determine the context to be invalid while it is valid for domain scoped provider.
So not mutating would also not be an option from my point of view. 

Having that said, I still think we would need to adapt the spec for this feature to be compliant.

packages/shared/src/open-feature.ts
```diff
@@ -49,6 +51,8 @@ export class ProviderWrapper<P extends CommonProvider<AnyProviderStatus>, S exte
         this._status = _statusEnumType.ERROR as S;
       }
     });
+
+    this._initialized = !(typeof _provider.initialize === 'function');
```
lukas-reining:
```suggestion
    this._initialized = typeof _provider.initialize !== 'function';
```

This should be okay and more simple.

lukas-reining (commented):
Sorry for the long delay @MattIPv4!

I really like the idea and I think it solves an existing problem for the static-context paradigm.
My main concern is that the change does not conform to the current spec in my eyes. I might be overlooking something there. Interested in the opinion of the rest of the @open-feature/technical-steering-committee.

If my concern is correct, I would favor to make this part of the spec. 
Again, I think this solves a real problem and could be a good addition to the spec.

packages/web/src/open-feature.ts
MattIPv4: I think setting the state to `ERROR` if the context has been determined to be invalid would break the purpose of this PR, as that would not allow you to suspend until the provider is ready?

packages/web/src/open-feature.ts
MattIPv4: Hmm, yes, I suspect this may need a spec change then. I don't see a great way to do this that would be compliant with what you've outlined in the spec currently.

packages/web/src/open-feature.ts
lukas-reining: Oh sure, I was mislead by 

```
If there was a validation error with the context, move the newly created provider to ERROR state.
```

packages/web/src/open-feature.ts
lukas-reining: Okay let's wait for feedback of the rest of the TC.

packages/web/src/open-feature.ts
MattIPv4: Ah I see -- the validation throws an error, it'll move to the ERROR state. If the validation successfully returns `false`, initialization is just skipped.

toddbaert: > Ah I see -- the validation throws an error, it'll move to the ERROR state. If the validation successfully returns `false`, initialization is just skipped.

That makes sense to me.

packages/web/src/open-feature.ts
toddbaert: > the default context could determine the context to be invalid while it is valid for domain scoped provider.

I think this is a key problem.

With respect to multiple domains, we need to decide:

- if a validation func returns false for only a subset of domains, what happens?
  - probably it's best for this subset to remain NOT_READY, right?
- if a validation func throws for only a subset of domains, should `setProviderAndWait` throw? consistent with initialization errors?
  - this means some providers will be READY and some will be ERROR
  - the other option is to consider every provider in ERROR state here, but that's not consistent with the first point above

As @lukas-reining says, this seems complex enough to require specification.

And as always, the question remains... is it better to simply avoid this addition, and push this responsibility down to providers. I know it's not ideal, but there's a complexity trade-off here as we can see.

MattIPv4: If a domain has a validation function, it also has its own context, so that does reduce the surface area we need to think about here. So, I think all we need to worry about is what happens when the default domain has a validation function, and another domain is reliant on the default domain context...

I think what the current code does there is maybe logical, in that the validation function just runs against the default domain, and the default context is only updated if the validation succeeds -- other domains have no "insight" into this, and only see the context change when the default domain decides it has a valid context change to process?

I do agree this'll need a spec change though, and I would ask for help on that as I'm not sure I know the intricacies well enough to propose a spec change for this.

I suppose the alternative would be to leave this to providers, though from experience (this is what we do currently, which prompted this PR), there isn't a particularly nice way to do this beyond essentially having initialize be a Promise that hangs for a super long time.

PR #1301, opened 2025-11-26
feat: allow method to be passed into setContext hook

MattIPv4:
## This PR

Allows for `setContext` returned by the `useContextMutator` hook to be passed a method that returns the new context, with the current (previous) context passed into it. This aligns the signature of `setContext` to the setter of a regular `useState` hook, making it easier to conditionally update the context without needing to keep track of the domain separately from the React context.

### DIFF
--- a/packages/react/src/context/use-context-mutator.ts
+++ b/packages/react/src/context/use-context-mutator.ts
@@ -18,10 +18,10 @@ export type ContextMutation = {
    * Context-aware function to set the desired context (see: {@link ContextMutationOptions} for details).
    * There's generally no need to await the result of this function; flag evaluation hooks will re-render when the context is updated.
    * This promise never rejects.
-   * @param updatedContext
+   * @param updatedContext New context object or method to generate it from the current context
    * @returns Promise for awaiting the context update
    */
-  setContext: (updatedContext: EvaluationContext | ((previousContext: EvaluationContext) => EvaluationContext)) => Promise<void>;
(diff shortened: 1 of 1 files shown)


packages/react/src/context/use-context-mutator.ts
```diff
@@ -34,14 +34,18 @@ export function useContextMutator(options: ContextMutationOptions = { defaultCon
...
     const previousContext = useRef<null | EvaluationContext>(null);
 
-    const setContext = useCallback(async (updatedContext: EvaluationContext) => {
-        if (previousContext.current !== updatedContext) {
+    const setContext = useCallback(async (updatedContext: EvaluationContext | ((previousContext: EvaluationContext) => EvaluationContext)): Promise<void> => {
+        const resolvedContext = typeof updatedContext === 'function'
+            ? updatedContext(OpenFeature.getContext(options?.defaultContext ? undefined : domain))
+            : updatedContext;
+
+        if (previousContext.current !== resolvedContext) {
```
MattIPv4, line 42: I wasn't quite sure what the `previousContext` ref was doing here, as it doesn't seem to actually maintain itself against any external changes, perhaps it's there to debounce changes?

toddbaert: Ya I believe this is just to debounce non-changes.

toddbaert (approved):
Nice idea!

Issue #1218, opened 2025-07-15
[BUG] Memory leak after destroying angular application


MillerSvt:
### Observed behavior

We have multiple angular applications per page, and each have their own feature flags. We use `web-sdk`, and implement `RestClient` with using Angular `HttpClient`. Then we create OpenFeatureClient like this:

```typescript
        {
            provide: dcoFeatureFlagClient,
            useFactory: async (): Promise<Client> => {
                const restClient = new OpenFeatureRestClient(url); // here is HttpClient inside
                const provider = new FeatureToggleProvider(restClient);
                const context = await contextFactory();

                return OpenFeature.setProvider(provider).getClient(context);
            },
        },
```

This creates some unexpected behavior:
1. All Angular applications, uses the same HttpClient instance (from last created application).
2. HttpClient and all its dependencies does not cleaned by GC, and still persists in memory heap dump until call `globalThis[Symbol(@openfeature/js-sdk/api)] = null`.
3. After destroying last created application, others cannot retrieve any feature flags, because HttpClient from last created application, that used by all applications is destroyed, and it throws `NG0205` error. 

### Expected Behavior

DO NOT create global OpenFeature instance in globalThis object. Add ability to create separate instances of OpenFeature API for each applications.

lukas-reining:
Thanks for opening the issue @MillerSvt!
As @beeme1mr says, the design of the SDK is to never let users create the API instances. 
Instead, the global instance is meant to manage the lifecycle of the clients, which can be distinguished by the usage of domains. 
Our Angular SDK also uses this approach to manage feature flagging over [custom directives](https://openfeature.dev/docs/reference/technologies/client/web/angular).

> We have multiple angular applications per page.

Are you doing this as micro frontends? Can you share which technology you uses? Something like [single-spa](https://single-spa.js.org/), iframes,[ @angular/elements](https://angular.dev/guide/elements) or [module federation](https://webpack.js.org/concepts/module-federation/)?

If so, and the apps have independent lifecycles, I understand that using non-domain scoped clients can lead to problems. 
In that case you should have a domain scoped client for each app.
Also, in that case you should not call `OpenFeature.shutdown` before all applications have been shut down.

PR #1303, opened 2025-11-28
feat: allow non-singleton access to OpenFeatureAPI

### LINKED ISSUE #1308: [FEATURE] Provide option to opt-out of OpenFeature API singleton
### Requirements

The OpenFeature specification requires (should) the `OpenFeature API` to be a singleton [[1]](https://openfeature.dev/specification/sections/flag-evaluation#requirement-111)
This design works good for most use cases and makes using OpenFeature very simple in most languages and frameworks. 
In the past there have been multiple cases where application authors asked for an option to get an instance of the  API that is not a singleton. 

Recently we had @MattIPv4 mentioning an issue using the singleton here in Slack [[2]](https://cloud-native.slack.com/archives/C06E4DE6S07/p1764115826255009) 

```
Just been bitten by OpenFeature being a global singleton -- we have a plugin-based architecture for our frontend, and we use OpenFeature in the core as well as in some plugins. Because it is a global singleton, and the core loads first, the version that the core is using is forced upon all the plugins even if they have a different version installed themselves. 
....
Not only was the core version winning out because it loaded first into the global singleton, but the library from core that was loading it is externalised in our plugin builds and baked into our docker image, so even overriding the core's dependency locally didn't work.
```

Also there was an issue mentioned in #1218, where an Angular App was decomposed into multiple Angular Apps. In this scenario, initialization and shutdown of the OpenFeature SDK could not be coupled to the lifecycle of the individual apps due to the shared SDK. 

In todays community meeting Leo and Aaron from Datadog (Eppo) shared they also had cases they could not use the singleton. 

We should clarify if all the cases mentioned here can be solved adequately with the current "singleton only" approach or if there should be a mechanism to get a non-singleton instances in cases where it does not work differently.

MattIPv4:
## This PR

Allows for folks who do not wish to interact with the global singleton instance of the OpenFeatureAPI in the web-sdk to instead create isolated non-singleton instances of the OpenFeatureAPI that they'll need to pass around themselves. This should be useful for micro-frontend environments where there is a risk for different versions of the SDK being loaded by micro-frontends and colliding in `window` due to the global singleton mechanism.

### Related Issues

Resolves #1308

cc https://github.com/open-feature/spec/issues/359

### Notes

N/A

### Follow-up Tasks

N/A

### How to test

tbd.

### DIFF
--- a/packages/react/src/context/use-context-mutator.ts
+++ b/packages/react/src/context/use-context-mutator.ts
@@ -35,6 +35,7 @@ export type ContextMutation = {
 export function useContextMutator(options: ContextMutationOptions = { defaultContext: false }): ContextMutation {
   const { client } = useContext(Context) || {};
   const domain = client?.metadata.domain;
+  const sdk = client?.metadata.sdk || OpenFeature;
 
   // TODO: Replace this warning with a thrown error in a future major release,
   //       to match the behavior of `useOpenFeatureProvider` + `useOpenFeatureClient`,
@@ -60,15 +61,14 @@ export function useContextMutator(options: ContextMutationOptions = { defaultCon
     async (
       updatedContext: EvaluationContext | ((currentContext: EvaluationContext) => EvaluationContext),
     ): Promise<void> => {
-      // TODO: Needs to handle `openfeature` option like OpenFeatureProvider
-      const previousContext = OpenFeature.getContext(options?.defaultContext ? undefined : domain);
+      const previousContext = sdk.getContext(options?.defaultContext ? undefined : domain);
       const resolvedContext = typeof updatedContext === 'function' ? updatedContext(previousContext) : updatedContext;
 
       if (previousContext !== resolvedContext) {
         if (!domain || options?.defaultContext) {
-          await OpenFeature.setContext(resolvedContext);
+          await sdk.setContext(resolvedContext);
         } else {
-          await OpenFeature.setContext(domain, resolvedContext);
+          await sdk.setContext(domain, resolvedContext);
         }
       }
     },
--- a/packages/react/src/context/use-context-mutator.ts
+++ b/packages/react/src/context/use-context-mutator.ts
@@ -60,6 +60,7 @@ export function useContextMutator(options: ContextMutationOptions = { defaultCon
     async (
       updatedContext: EvaluationContext | ((currentContext: EvaluationContext) => EvaluationContext),
     ): Promise<void> => {
+      // TODO: Needs to handle `isolated` option like OpenFeatureProvider
       const previousContext = OpenFeature.getContext(options?.defaultContext ? undefined : domain);
       const resolvedContext = typeof updatedContext === 'function' ? updatedContext(previousContext) : updatedContext;
 
--- a/packages/react/src/context/use-context-mutator.ts
+++ b/packages/react/src/context/use-context-mutator.ts
@@ -60,7 +60,7 @@ export function useContextMutator(options: ContextMutationOptions = { defaultCon
     async (
       updatedContext: EvaluationContext | ((currentContext: EvaluationContext) => EvaluationContext),
     ): Promise<void> => {
-      // TODO: Needs to handle `isolated` option like OpenFeatureProvider
+      // TODO: Needs to handle `openfeature` option like OpenFeatureProvider
       const previousContext = OpenFeature.getContext(options?.defaultContext ? undefined : domain);
       const resolvedContext = typeof updatedContext === 'function' ? updatedContext(previousContext) : updatedContext;
 
--- a/packages/web/src/open-feature.ts
+++ b/packages/web/src/open-feature.ts
@@ -362,7 +362,7 @@ export class OpenFeatureAPI
       (domain?: string) => this.getContext(domain),
       () => this.getHooks(),
       () => this._logger,
-      { domain, version },
+      { domain, version, sdk: this },
     );
   }
 
--- a/packages/web/src/open-feature.ts
+++ b/packages/web/src/open-feature.ts
@@ -19,6 +19,7 @@ type DomainRecord = {
 };
 
 const _globalThis = globalThis as OpenFeatureGlobal;
+const _localThis = {} as OpenFeatureGlobal;
 
 export class OpenFeatureAPI
   extends OpenFeatureCommonAPI<ClientProviderStatus, Provider, Hook>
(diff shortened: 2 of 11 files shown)


MattIPv4:
Hey!

> **1. Higher Level SDKs and integrations** Several mechanisms are relying on the singleton. E.g the React and Angular SDK are using the singleton today. A micro frontend architecture using the higher level React and Angular SDKs (ignoring all other possible problems with the SDKs and micro frontends) would not be benefitting from this today. And it could become hard to communicate the intended use and possible problems with these SDKs using micro frontends. In both SDKs, I actually think this could be solved by updating them to use the new `package local` singleton as users of these SDKs typically do not have to interact with the singleton anyways.

To avoid this being a breaking change, I don't want to switch the higher-level SDKs over to using the isolated instance by default. I can see there being a world in which users are wanting the global singleton behaviour there, even in a micro-frontend -- you might want the provider set by the core of the frontend to be accessible by micro-frontends using the higher-level SDKs.

I think my suggestion would be to add an `isolated` flag to the higher-level SDK providers (I'm only familiar with the React one, so do correct me if that doesn't make sense for them all), so that folks can opt into the isolated behavior if they want, similar to how you'd opt into it if you were just using the base web SDK.

> **2. Provider lifecycle** Allowing more than one instance of the API, we should be careful about the provider lifecycles. We should never have a provider instance get registered to more than one instance of the OpenFeature API. If we had, the static context and the initialization and shutdown make the provider state inconsistent, always resulting in a `last write wins` situation. A solution to handle this could be to add a method to the provider class to `lock` the provider. When one SDK would run `setProvider`, it could check the lock and either call `setLocked` or fail if the provider has already been locked by another instance of the API. As a provider may be bound to the same SDK "multiple times", we could save the API object identity or generate a ID per instance which would be checked in the lock.

Absolutely agree with this, adding a locking mechanism so that a provider can't be reused makes sense. Though, unless I've missed something, this is already a problem today, as you can add a provider to multiple domains? Do we want to continue allowing this (in which case we'll want to track a ref to the API instance), or prevent that as part of the new locking mechanism (in which case we can just track a lock bool)?

> **3. API clarity** I am not 100% sure about having the `package local singleton`. Having the `package local singleton` as a getter on the global singleton could make understanding the consequences of using either pretty complicated. My feeling is, that is would be better to allow for the singleton as it is today and a way to create a new "non-singleton" API. This would give the responsibility of making the API available throughout the code to the app author in the case they can not use the window singleton. They could either create a singleton theirselves or use something like dependency injection if they want. To me this would make the two ways of using the API more clear to app authors.

I'm not sure I follow why you wouldn't want a package-local singleton? If you import this in multiple files, you'd want the same instance to interact with? Or are you suggesting you'd prefer folks to manage this fully themselves and export/import a single instance around their app on their own? I don't really see any benefit to the latter; it seems like a worse DX for no benefit (once you've got package-level isolation to solve the SDK versioning issues, domains give you any additional isolation you'd need, I think)?

lukas-reining:
> To avoid this being a breaking change, I don't want to switch the higher-level SDKs over to using the isolated instance by default. I can see there being a world in which users are wanting the global singleton behaviour there, even in a micro-frontend -- you might want the provider set by the core of the frontend to be accessible by micro-frontends using the higher-level SDKs.
>
> I think my suggestion would be to add an isolated flag to the higher-level SDK providers (I'm only familiar with the React one, so do correct me if that doesn't make sense for them all), so that folks can opt into the isolated behavior if they want, similar to how you'd opt into it if you were just using the base web SDK.

That's a valid point. I think this would work

lukas-reining:
> I'm not sure I follow why you wouldn't want a package-local singleton? If you import this in multiple files, you'd want the same instance to interact with? Or are you suggesting you'd prefer folks to manage this fully themselves and export/import a single instance around their app on their own? I don't really see any benefit to the latter; it seems like a worse DX for no benefit (once you've got package-level isolation to solve the SDK versioning issues, domains give you any additional isolation you'd need, I think)?

> Or are you suggesting you'd prefer folks to manage this fully themselves and export/import a single instance around their app on their own?

Yes I do. But I am not 100% sure if we want it or if the package-local singleton solves enough of an issue while we may not want to handler others for now.
Issues like #1218 suggest that there are cases with only one version and instance of the SDK package where it could be necessary to have two API instances.
For that I want to be sure that we know before if we want to consider and solve these cases or not. 

Thinking about it, I might be leaning towards agreeing with the package-local singleton and domains beeing sufficient, 
even though does not handle the case described in #1218.

MattIPv4:
> Binding to multiple domains is not a problem. The SDK tracks if a provider is still used in another domain or the default domain. Only if a provider is not registered for any domain anymore, it is `shutdown`. And for setting the context it is not a problem as the static context is the same for all domains as it is only set on the SDK.
> 
> There really must be two API instances for this to be a problem.

Ah good to know, I hadn't poked around closely enough at how this works today!



> Good question, I think we can solve this in another issue. I will open one up for this explicitly.
> If we provide a non-singleton way for other languages, the locking should even become part of the API.
> But I would wait for the result of this discussion.

That sounds good to me -- leave it as a known gap in this implementation for now (given this is very explicitly opt-in as a whole, I'd hope folks aren't going to mix and match both), and then address it separately down the road 👍

packages/web/src/open-feature.ts
```diff
@@ -421,8 +426,58 @@ export class OpenFeatureAPI
...
+   */
+  getIsolated: () => OpenFeatureAPI;
+}
+
+const createOpenFeatureAPI = (): OpenFeatureAPIWithIsolated =>
+  Object.assign(OpenFeatureAPI.getInstance(), {
+    getIsolated() {
+      return OpenFeatureAPI.getInstance(false);
+    },
+  });
```
MattIPv4, line 477: Doing this via `Object.assign`, rather than a method on the class, so that you can do `OpenFeature.getIsolated()` but not `OpenFeature.getIsolated().getIsolated()`.

packages/react/src/context/use-context-mutator.ts
```diff
@@ -39,6 +39,7 @@ export function useContextMutator(options: ContextMutationOptions = { defaultCon
     async (
       updatedContext: EvaluationContext | ((currentContext: EvaluationContext) => EvaluationContext),
     ): Promise<void> => {
+      // TODO: Needs to handle `openfeature` option like OpenFeatureProvider
```
MattIPv4: Once https://github.com/open-feature/js-sdk/pull/1317 is in, I suspect I can apply a similar approach to that here, where the client tracks a reference to the OpenFeatureAPI instance it came from, and we can use that here to infer the correct instance

MattIPv4:
With #1317 landed, pushed up a change to follow a similar pattern that allows for the React SDK to properly utilise an isolated SDK instance if desired.

PR #1311, opened 2025-12-09
ci: enable Prettier formatting

### LINKED ISSUE #1302: Prettier installed as devDependency but not run
👋 While I've been working on a couple of PRs I noticed that Prettier is installed as a dev dependency and there is a minimal Prettier config in the repository, but Prettier never seems to actually be run, nor does there seem to be any CI check for it.

I do see that `prettier` is in the ESLint `extends`, but as far as I can tell that only disbles rules that'd interfere with Prettier being run via the CLI -- you'd need to have `eslint-plugin-prettier` installed (which would then be `plugin:prettier/recommended` in the `extends`) to actually run Prettier as part of ESLint, I think.

Am I missing something, and it is working, just not the way I expect

If not, is it desired to have it working? Via the ESLint plugin, or separately with the Prettier CLI?

MattIPv4:
## This PR

Enables Prettier for the repository, adding scripts to the root package for checking + applying formatting. Adds a new CI job to check formatting on PRs (and also moves the lint to there, as I don't think lint needs to run across the Node.js version matrix).

I've also added husky + lint-staged to automatically run Prettier against files being committed locally, but happy to pull that back out if folks would prefer not to have that.

### Related Issues

Resolves #1302 

### Notes

I'd recommend reviewing the commits individually, except the last commit, which was just the result of actually running Prettier.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,8 +1,8 @@
 {
   "packages/nest": "0.2.5",
-  "packages/react": "1.0.2",
-  "packages/web": "1.7.1",
-  "packages/server": "1.20.0",
+  "packages/react": "1.1.0",
+  "packages/web": "1.7.2",
+  "packages/server": "1.20.1",
   "packages/shared": "1.9.1",
   "packages/angular/projects/angular-sdk": "0.0.19"
 }
--- a/packages/react/CHANGELOG.md
+++ b/packages/react/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## [1.1.0](https://github.com/open-feature/js-sdk/compare/react-sdk-v1.0.2...react-sdk-v1.1.0) (2025-12-09)
+
+
+### ✨ New Features
+
+* allow method to be passed into setContext hook ([#1301](https://github.com/open-feature/js-sdk/issues/1301)) ([279f9bf](https://github.com/open-feature/js-sdk/commit/279f9bf624d4e8a2171522dea49871173ea6ccd3))
+
+
+### 🐛 Bug Fixes
+
+* await context change in useContextMutator setContext ([#1306](https://github.com/open-feature/js-sdk/issues/1306)) ([7ef721a](https://github.com/open-feature/js-sdk/commit/7ef721aa9ba4c70497759a4de462715d1b32b240))
+* compare full EvaluationDetails to prevent stale data ([#1287](https://github.com/open-feature/js-sdk/issues/1287)) ([8133a4f](https://github.com/open-feature/js-sdk/commit/8133a4f5c3d9f61ad30c84f532e01df8c7785a4f))
+* in-line docs, un-needed AbortController creation ([#1291](https://github.com/open-feature/js-sdk/issues/1291)) ([9b05be9](https://github.com/open-feature/js-sdk/commit/9b05be9a045232b36dec008611981bd721694abe))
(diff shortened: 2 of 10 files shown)


.github/workflows/pr-checks.yaml
```diff
@@ -11,7 +11,7 @@ on:
   merge_group:
 
 jobs:
-  build-test-lint:
```
beeme1mr, line 14: Reminder to myself to check if this will impact the repo settings.

.github/workflows/pr-checks.yaml
```diff
@@ -34,15 +34,31 @@ jobs:
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: ".nvmrc"
+          cache: "npm"
+
+      - name: Install
+        run: npm ci
+
+      - name: Format
+        run: npm run format
```
beeme1mr, line 57: Should we fail the linting step if this makes a change? It seems like we could be masking linting issues if we run the formatter here.

I noticed that this PR added Husky, so this may not be much of a concern.

packages/angular/projects/angular-sdk/CHANGELOG.md
beeme1mr, line 1: Could the change logs be ignored by Prettier? These are auto-generated by Release Please.

beeme1mr (approved):

.github/workflows/pr-checks.yaml
beeme1mr: We're using `e2e` as the required status check.

.github/workflows/pr-checks.yaml
MattIPv4: `npm run format` _checks_ the formatting and fails if there is an issue -- `npm run format:fix` would be what actually _writes_ changes. I followed the same pattern as `lint`/`lint:fix` for consistency.

packages/angular/projects/angular-sdk/CHANGELOG.md
MattIPv4: Sure! All `CHANGELOG.md` files?

beeme1mr: Yes please

.github/workflows/pr-checks.yaml
beeme1mr: Cool, thanks for clarifying.

packages/angular/projects/angular-sdk/CHANGELOG.md
MattIPv4: Done :)

lukas-reining (approved):

package.json
```diff
@@ -15,10 +15,13 @@
     "e2e": "npm run e2e-server && npm run e2e-web",
     "lint": "npm run lint --workspace=packages/shared --workspace=packages/server --workspace=packages/web --workspace=packages/react --workspace=packages/angular --workspace=packages/nest",
     "lint:fix": "npm run lint:fix --workspace=packages/shared --workspace=packages/server --workspace=packages/web --workspace=packages/react --workspace=packages/angular --workspace=packages/nest",
+    "format": "prettier --check .",
+    "format:fix": "prettier --write .",
```
lukas-reining, line 19: My personal goto is doing it the opposite. 
Having `format` and `format:check`, to me that is a bit more intuitive for development. 
But both is okay to me.

MattIPv4: I tend to flip-flop between the two -- I did it this way only to be consistent with the existing `lint` and `lint:fix`.

> to me that is a bit more intuitive for development.

In theory you won't need to think about these ever, husky/lint-staged will do it automatically when you commit!

lukas-reining: Oh that consistency argument makes sense. I overlooked that :)

PR #1316, opened 2025-12-11
feat: Angular 21 support

### LINKED ISSUE #1314: [FEATURE] Support Angular 21
### Requirements

Angular 21 has been released a month ago. We would like to upgrade but peer deps prevent it.

CharlesEDF:
## This PR
Adds Angular 21 support

### Related Issues
Fixes #1314 

### Notes
@angular-eslint has also been updated to v21


packages/angular/tsconfig.json
```diff
@@ -19,13 +19,12 @@
     "sourceMap": true,
     "declaration": false,
     "experimentalDecorators": true,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
beeme1mr, line 22: I don't think this will cause any issues, and it seems correct to me. However, it may be a bit risky. Is there a reason to make the change in this PR? Perhaps it should be in a separate PR so that it shows up as a change in the change log.

CharlesEDF: Thanks for the feedback!

This specific change was introduced automatically by the Angular migration schematics when I ran `ng update @angular/core@21 @angular/cli@21` (along with the addition of Vitest and other updates). It wasn't a manual modification on my part.

I understand your concern about the risk. I can certainly revert this part of the change and remove it from this PR if you prefer that it not be merged now.

Let me know what you'd like me to do!

lukas-reining: To me this looks fine. Especially as this is mainly relevant for the build. 
Let me verify it tomorrow. 
After that we can merge it.

lukas-reining (approved):
Just had a look and compared the build outputs between the old and current version.
They are compatible and only changed in a non-breaking way. 
The way exports and imports are structured kept the same (as expected).

We can merge this, thank you again @CharlesEDF!

PR #1317, opened 2025-12-11
fix: infer domain from client for useContextMutator setContext

MattIPv4:
## This PR

As identified when chatting in https://github.com/open-feature/js-sdk/pull/1303#issuecomment-3627349190, if you pass a `Client` into `<OpenFeatureProvider/>`, then `useContextMutator#setContext` will always see `domain` as `undefined` and so will always mutate the default domain even if the client is not for that domain -- instead, we should rely on the domain that the client tells us it is for!

I've also updated the tests generally for `useContextMutator` slightly, as they were sometimes reusing contexts set by other tests, leading to unexpected failures when I added this new test.

### Notes

This, and the provider hook, were the only two things relying on the `domain` property being exposed in the internal context, so I've removed it to hopefully prevent any future logic from running into this same trap.

Open question that I've left as a TODO, should the hook throw an error if someone tries to use the hook outside of `<OpenFeatureProvider/>` (if they've not set `defaultContext: true`)? Though I fear that would make this a breaking change (though arguably it is still a fix as being able to use this outside a hook today feels like a bug, not an intentional design decision)?

### DIFF
--- a/packages/react/src/context/use-context-mutator.ts
+++ b/packages/react/src/context/use-context-mutator.ts
@@ -1,4 +1,4 @@
-import { useCallback, useContext } from 'react';
+import { useCallback, useContext, useEffect, useState } from 'react';
 import type { EvaluationContext } from '@openfeature/web-sdk';
 import { OpenFeature } from '@openfeature/web-sdk';
 import { Context } from '../internal';
@@ -36,8 +36,25 @@ export function useContextMutator(options: ContextMutationOptions = { defaultCon
   const { client } = useContext(Context) || {};
   const domain = client?.metadata.domain;
 
-  // TODO: If `defaultContext` is `false`, and we don't have a `client`,
-  //       should we throw an error (that we're not inside `<OpenFeatureProvider/>`)?
+  // TODO: Replace this warning with a thrown error in a future major release,
+  //       to match the behavior of `useOpenFeatureProvider` + `useOpenFeatureClient`,
+  //       when `defaultContext` isn't explicitly set to true.
+  const [warned, setWarned] = useState(false);
+  useEffect(() => {
+    if (options.defaultContext || domain) {
+      if (warned) {
+        setWarned(false);
+      }
+      return;
+    }
+
+    if (!warned) {
+      console.warn(
+        '[useContextMutator] No domain available from OpenFeature context; are you using <OpenFeatureProvider/>? setContext will mutate the default context, as if `defaultContext: true` were set. This may result in a thrown error in the future.',
+      );
+      setWarned(true);
+    }
+  }, [warned]);
 
   const setContext = useCallback(
     async (


packages/react/src/context/use-context-mutator.ts
```diff
@@ -33,7 +33,11 @@ export type ContextMutation = {
  * @returns {ContextMutation} context-aware function(s) to mutate evaluation context
  */
 export function useContextMutator(options: ContextMutationOptions = { defaultContext: false }): ContextMutation {
-  const { domain } = useContext(Context) || {};
+  const { client } = useContext(Context) || {};
+  const domain = client?.metadata.domain;
+
+  // TODO: If `defaultContext` is `false`, and we don't have a `client`,
```
lukas-reining: Mh good question.

> [...] (though arguably it is still a fix as being able to use this outside a hook today feels like a bug, not an intentional design decision)?

This is true. On the other hand, I think we still need to assume that this has happened. 
Maybe we could start by logging a warning if that happens? Later we could throw then. Even thought you are right, I think we should be conservative about throwing to remove unintended usage that is already out. 

@toddbaert what do you think?

lukas-reining (commented):
The change looks good, thanks @MattIPv4!
Left my opinion regarding the breaking change.

packages/react/test/provider.spec.tsx
```diff
@@ -196,8 +196,8 @@ describe('OpenFeatureProvider', () => {
     };
 
     it('should update context when a domain is set', async () => {
-      const DOMAIN = 'mutate-context-tests';
-      OpenFeature.setProvider(DOMAIN, suspendingProvider());
+      const DOMAIN = 'mutate-context-domain';
+      OpenFeature.setProvider(DOMAIN, suspendingProvider(), {});
```
beeme1mr: Why was this change necessary?

packages/react/src/context/use-context-mutator.ts
beeme1mr: We do throw if you use a hook without the provider context. However, I tend to agree with @lukas-reining suggestion that logging may be the better approach here. If the React SDK has easy access to the safe logger, I'd recommend going that route.

packages/react/test/provider.spec.tsx
MattIPv4: Did this for all of the tests here to ensure they're working with fresh contexts for each test, rather than potentially colliding into each other with an indeterminate global state in the test suite (gotta love singletons).

MattIPv4: Would you prefer I do a beforeEach with OpenFeature.clearContexts() instead?

packages/react/src/context/use-context-mutator.ts
MattIPv4: Hm, I don't see any logger usage in the React source. Where would the logger be coming from here, is there a global one?

beeme1mr: It's in core but I wasn't sure it was used in the React SDK yet.

https://github.com/open-feature/js-sdk/tree/main/packages/shared/src/logger

It may not be necessary if we don't allow users to use their own logger implementation. In this case, it's probably fine to just use `console.warn`.

packages/react/test/provider.spec.tsx
beeme1mr: I would be slightly cleaner in my opinion but I'm good either way.

packages/react/src/context/use-context-mutator.ts
MattIPv4: Added a regular `console.warn` with a state so it doesn't log repeatedly 👍

beeme1mr (approved):

Issue #1321, opened 2025-12-18
[FEATURE] Add suspense/loading support in "Declarative" (component-based) API


toddbaert:
### Requirements

With https://github.com/open-feature/js-sdk/pull/1164, we added a new component-based API. This doesn't yet support any kind of "loading" or suspense until the provider is [ready](https://openfeature.dev/docs/reference/concepts/events#provider_ready), or while context is being [reconciled](https://openfeature.dev/docs/reference/concepts/events#provider_reconciling-static-contextclient-side-only).

We should support this. There's some design choices to be made here. Consider how to give users control, as well as reasonable defaults, and also how to be consistent with similar configuration options elsewhere.

I see 2 possibilities (both may be valuable):

- add a `suspend` option as we have in our hooks: https://github.com/open-feature/js-sdk/blob/e3c9d79474bf6d6e4a7ef122e6617d25a65f6ade/packages/react/README.md?plain=1#L246
  - we probably want to respect the settings at the context-provider level as well, similar to the way our hooks do.
- add a "loading" template option (I'm less sure about this, since suspense is essentially a native means of doing this)
  - this would be displayed during provider reconciliation, or while the provider is not ready
  - try to avoid duplication with existing suspense usage
  - again, the best option may be to simply not include this, but I'm not sure

toddbaert:
FYI @weyert @marcozabel

marcozabel:
In the [mentioned PR](https://github.com/open-feature/js-sdk/pull/1322) I’ve updated the declarative component to provide the same level of control as the hook. Instead of only offering a `suspend` option, the entire evaluation config is now passed as a prop.

PR #1322, opened 2025-12-19
feat: Add evaluation options to FeatureFlag component

marcozabel:
## This PR
This PR is adding evaluation options to the FeatureFlag component to allow the use of suspend options.

### Related Issues

https://github.com/open-feature/js-sdk/issues/1321

### DIFF
--- a/packages/react/src/declarative/FeatureFlag.tsx
+++ b/packages/react/src/declarative/FeatureFlag.tsx
@@ -97,7 +97,7 @@ export function FeatureFlag<T extends FlagValue = FlagValue>({
 }: FeatureFlagComponentProps<T>): React.ReactElement | null {
   const details = useFlag(flagKey, defaultValue, {
     updateOnContextChanged: true,
-    ...evaluationOptions
+    ...evaluationOptions,
   });
 
   // If the flag evaluation failed, we render the fallback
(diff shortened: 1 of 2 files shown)


packages/react/src/declarative/FeatureFlag.tsx
```diff
@@ -86,10 +92,12 @@ export function FeatureFlag<T extends FlagValue = FlagValue>({
   predicate,
   defaultValue,
   children,
+  evaluationOptions = {},
   fallback = null,
 }: FeatureFlagComponentProps<T>): React.ReactElement | null {
   const details = useFlag(flagKey, defaultValue, {
     updateOnContextChanged: true,
```
beeme1mr, line 99: We should `omit` the `updateOnContextChanged` property from `ReactFlagEvaluationOptions` if it's not configurable?

Is there a reason why it isn't configurable? Perhaps it could default to true but be overridden via the `ReactFlagEvaluationOptions`

marcozabel: It is configurable, because the `evaluationOptions` are spread after the `updateOnContextChanged` it can be overridden.

beeme1mr: Yes, you're right. Sorry... too early I guess.

PR #1325, opened 2026-01-02
feat: Add isolated (non-singleton) OpenFeature API instances

jonathannorris:
## Summary

- Adds `createIsolatedOpenFeatureAPI()` function accessible via `/isolated` sub-path export for both web and server SDKs
- Enables micro-frontend architectures and testing scenarios where isolated OpenFeature instances are needed
- Updates React SDK's `OpenFeatureProvider` to accept an optional `openfeature` prop for using isolated instances

## Motivation

Addresses the singleton limitations discussed in https://github.com/open-feature/spec/issues/359. The global singleton can cause issues in:
- Micro-frontend architectures where different parts of an app need isolated providers
- Tests that mutate the singleton and trample on each other
- DI frameworks that don't work well with externally-managed singletons

## Usage

```typescript
// Web SDK
import { createIsolatedOpenFeatureAPI } from '@openfeature/web-sdk/isolated';

const MyOpenFeature = createIsolatedOpenFeatureAPI();
MyOpenFeature.setProvider(myProvider);
const client = MyOpenFeature.getClient();

// React SDK
<OpenFeatureProvider openfeature={MyOpenFeature}>
  {children}
</OpenFeatureProvider>

// Server SDK
import { createIsolatedOpenFeatureAPI } from '@openfeature/server-sdk/isolated';
```

## Implementation

- Extracts core logic into `OpenFeatureAPIBase` class (~380 lines for web, ~220 lines for server)
- `OpenFeatureAPI` becomes a thin singleton wrapper (~35 lines)
- Isolated instances are only accessible via the `/isolated` sub-path - **no changes to `OpenFeatureAPI` class interface**
- Each isolated instance has its own providers, context, hooks, and event handlers

## Changes

| Package | Files Added | Files Modified |
|---------|-------------|----------------|
| `@openfeature/web-sdk` | 4 | 3 |
| `@openfeature/server-sdk` | 4 | 3 |
| `@openfeature/react-sdk` | 0 | 1 |


packages/react/src/provider/provider.tsx
```diff
@@ -31,8 +54,11 @@ type ProviderProps = {
  * @param {ProviderProps} properties props for the context provider
  * @returns {OpenFeatureProvider} context provider
  */
-export function OpenFeatureProvider({ client, domain, children, ...options }: ProviderProps) {
-  const stableClient = React.useMemo(() => client || OpenFeature.getClient(domain), [client, domain]);
+export function OpenFeatureProvider({ client, domain, openfeature, children, ...options }: ProviderProps) {
```
jonathannorris, line 58: I will see if there is a way to have a separate React package like the other SDKs.

packages/web/src/isolated.ts
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * An isolated (non-singleton) OpenFeature API instance.
+ * This class is NOT exported from the main entry point.
+ * @internal
+ */
+class OpenFeatureIsolatedAPIImpl extends OpenFeatureAPIBase {
+  constructor() {
+    super();
+  }
+}
```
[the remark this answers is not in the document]
jonathannorris: The constructor is required because `OpenFeatureAPIBase` has a protected constructor. Without an explicit constructor in the subclass, TypeScript will try to use a default public constructor which cannot access the protected parent constructor.

packages/server/src/isolated.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * An isolated (non-singleton) OpenFeature API instance.
+ * This class is NOT exported from the main entry point.
+ * @internal
+ */
+class OpenFeatureIsolatedAPIImpl extends OpenFeatureAPIBase {
+  constructor() {
+    super();
+  }
+}
```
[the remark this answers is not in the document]
jonathannorris: The constructor is required because `OpenFeatureAPIBase` has a protected constructor. Without an explicit constructor in the subclass, TypeScript will try to use a default public constructor which cannot access the protected parent constructor.

packages/react/src/provider/provider.tsx
```diff
@@ -11,6 +11,28 @@ type ClientOrDomain =
        * @see OpenFeature.setProvider() and overloads.
        */
       domain?: string;
+      /**
+       * An isolated OpenFeature API instance to use instead of the global singleton.
```
moredip, line 16: Am I right that this `openfeature` prop should ONLY be used when someone is doing the advanced isolated stuff? If so I think we should add that clearly in the docstring - "WARNING: this is only required for special cases where an isolated instance is needed. If you don't know what that means, you should not be using this prop".

We could also call it `isolatedOpenFeature` to make it clearer that it's a special-case thing, not a generally-used thing.

packages/server/src/isolated.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * @module @openfeature/server-sdk/isolated
+ * Provides non-singleton OpenFeature API instances for testing and multi-tenant scenarios.
+ * WARNING: This module provides non-singleton instances that do NOT share state
```
moredip, line 4: I think this warning could be a little stronger/clearer

packages/server/src/open-feature.ts
```diff
@@ -21,26 +8,16 @@ type OpenFeatureGlobal = {
...
-  protected _createEventEmitter = () => new OpenFeatureEventEmitter();
-
-  private _transactionContextPropagator: TransactionContextPropagator = NOOP_TRANSACTION_CONTEXT_PROPAGATOR;
-
+/**
+ * The OpenFeatureAPI is the entry point for the OpenFeature SDK.
+ * This is a singleton class that provides access to the global OpenFeature API instance.
+ *
+ * For isolated (non-singleton) instances, use the `createIsolatedOpenFeatureAPI` function
+ * from `@openfeature/server-sdk/isolated`.
```
moredip, line 16:
```suggestion
 * from `@openfeature/server-sdk/isolated` (this is an advanced use case; only use it if you know you need it)
```

moredip:
If we do add this feature we should also add a writeup to the docs, and link to it from the docstring here. Just a quick summary of what the isolated use case is, and an extolation to only use it if you know what you're doing.

PR #1328, opened 2026-01-15
feat: nestjs async factory

imranismail:
Adds `forRootAsync()` method to enable asynchronous configuration of the OpenFeature NestJS module using dependency injection. This allows OpenFeature to be configured with values from other providers (e.g., `ConfigService`) that are only available at runtime.

### Changes

**Implementation:**
- Migrated to NestJS `ConfigurableModuleBuilder` pattern for standardized async module support
- Added `forRootAsync()` method that supports `useFactory`, `useClass`, and `useExisting` patterns
- Refactored `forRoot()` to use the same internal `buildModule()` method for consistency
- Changed provider initialization to use `setProviderAndWait()` instead of synchronous `setProvider()` for proper async handling
- Moved OpenFeature SDK initialization into a factory provider to ensure it runs after module bootstrap

**New Features:**
- `forRootAsync()` with full DI support (useFactory, inject, imports, etc.)
- `domains` configuration in extras for domain-scoped client injection, unfortunately this is required to be able to build the client token for domains, since dependency tree needs to be determined synchronously at "build" time.
- `isGlobal` configuration option (defaults to `true`)
- Async provider initialization with proper dependency ordering

### Migration Guide

**Before (still works):**
```typescript
OpenFeatureModule.forRoot({
  defaultProvider: myProvider,
  useGlobalInterceptor: true,
})
```

**After (with async config):**
```typescript
OpenFeatureModule.forRootAsync({
  imports: [ConfigModule],
  useFactory: async (config: ConfigService) => ({
    defaultProvider: new Provider({
      apiKey: config.get<string>('PROVIDER_API_KEY'),
    }),
    providers: {
      'domain-scoped-client': new Provider({
        apiKey: config.get<string>('DOMAIN_A_API_KEY'),
      }),
    }
  }),
  inject: [ConfigService],
  isGlobal: true,
  useGlobalInterceptor: true,
  domains: ['domain-scoped-client'], // if using domain-scoped clients
})
```

### Breaking Changes
**Minor:** The `useGlobalInterceptor` option moved from the main options to extras. However, this is backward compatible because the ConfigurableModuleBuilder merges extras into the options type.


imranismail:
@lukas-reining I've changed the implementation slightly, also ensure existing test passes.

I'm a bit concerned about having to specify the domains again in the extras, happy to hear some ideas if you have.

lukas-reining (commented):
Hey @imranismail sorry that it took so long to review. 
I really like the change but the `domain` field is a problem to me that I would like to avoid. Having it would open a huge inconvenience that I would like to avoid. 
This would not be a problem if we implement the Module as a class implementing the register method. 

I totally see why you opted for the builder, especially as the docs state the following. 

```
As manually creating highly configurable, dynamic modules that expose async methods (registerAsync, forRootAsync, etc.) is quite complicated, especially for newcomers, Nest exposes the ConfigurableModuleBuilder class that facilitates this process and lets you construct a module "blueprint" in just a few lines of code.
```

But the problem accessing the options was exactly why we opted for the implementation of the module as a class implementing the `forRoot` method which is able to receive all options.
I would like to keep it this way, meaning we could have an internal method like `initializeOpenFeature` that `forRoot` and `forRootAsync` can both call. 
Does that make sense to you?

If so please change it to work this way. 
I am more than happy to assist on it if you wish so.

lukas-reining:
Hey @imranismail are you still interested in getting this in? :)

Issue #1329, opened 2026-01-15
Roadmap to v1.0.0 and Production Readiness for the Angular package


CharlesEDF:
Hello OpenFeature team,

First of all, thank you for the work on this Angular SDK. We are currently integrating @openfeature/angular-sdk into our enterprise applications.

I noticed that the package is currently on version 0.0.x. Since Semantic Versioning implies that major version 0 is for initial development and that the public API should not be considered stable, this raises a few questions regarding our internal stability policies and tooling configuration (specifically Renovate, which treats 0.x updates with more caution regarding breaking changes).

Could you shed some light on the current status of this SDK?
1. Production Readiness: Despite the version number, is the current API considered stable enough for production use?
2. Roadmap to v1.0.0: What are the missing pieces or blockers preventing a 1.0.0 release? We'd be happy to help !

Understanding the path to v1 would help us better manage our dependencies and ensure the stability of our Angular apps.

Thanks again for your help!

toddbaert:
Hey @CharlesEDF !

The Angular SDK is sub 1.0 now really simply because it's new relative to the other SDKs, and we wanted to see some adoption before we stabilized it. Personally, I'm not aware of any planned changes or issues that would force us to break contracts at this point, so moving to a 1.0 seems reasonable to me.

@lukas-reining Besides the obvious (updating the versioning) is there anything that you consider important to do before a 1.0? Any API changes you are concerned about? I think our downloads are usage are at this point sufficient to consider stabilizing the API.

cc @jonathannorris @beeme1mr @thomaspoignant @kinyoklion

PR #1333, opened 2026-01-20
chore: prompt release for angular 1.0.0

lukas-reining:
Releases the Angular SDK as 1.0.0 closing #1329


toddbaert (approved):
Approving but I changed the title since this won't actually release, but just make the next release please PR a 1.0.

Like I said before, you might have to do "something" in the file path to prompt it, not sure if release-as will force a release with no changes.

PR #1342, opened 2026-02-15
feat: lock provider binding to a single API instance

lukas-reining:
## This PR
Once we implement non-singleton OpenFeature APIs as discussed in https://github.com/open-feature/spec/issues/359, we need to make sure a single provider is only ever assigned to one API instance at a time. This is necessary due to the API instance managing the lifecycle of the provider. 

This PR drafts a mechanism to track and enforce the association between provider instances and `OpenFeatureAPI` instances, preventing a provider from being registered with multiple APIs at the same time. 
It does this by using a symbol to "bind" a provider to an API instance and ensures proper cleanup (unbinding) when providers are replaced or the API is shut down.

This change is transparent to the provider implementations. 

### Related Issues

https://github.com/open-feature/spec/issues/359

### DIFF
--- a/packages/shared/src/open-feature.ts
+++ b/packages/shared/src/open-feature.ts
@@ -22,6 +22,13 @@ import type { Paradigm } from './types';
 type AnyProviderStatus = ClientProviderStatus | ServerProviderStatus;
 
 /**
+ * Well-known Symbol used to track which {@link OpenFeatureCommonAPI} instance a provider is currently bound to.
+ * The symbol is shared across module boundaries and globally unique in the runtime.
+ * @internal
+ */
+export const BOUND_API_KEY = Symbol.for('@openfeature/js-sdk/provider/bound-api');
+
+/**
  * A provider and its current status.
  * For internal use only.
  */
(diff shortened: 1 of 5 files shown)


packages/shared/src/open-feature.ts
```diff
@@ -21,6 +21,13 @@ import type { Paradigm } from './types';
 
 type AnyProviderStatus = ClientProviderStatus | ServerProviderStatus;
 
+/**
+ * Well-known Symbol used to track which {@link OpenFeatureCommonAPI} instance a provider is currently bound to.
+ * The symbol is shared across module boundaries and globally unique in the runtime.
+ * @internal
```
lukas-reining, line 27: Do we have to move this to an extra internal file as we do with `OpenFeatureClient` implementations? 
This is only exported for the tests.

packages/shared/src/open-feature.ts
```diff
@@ -217,6 +224,21 @@ export abstract class OpenFeatureCommonAPI<
...
   ): this;
 
+  private bindProvider(provider: P): void {
+    Object.defineProperty(provider, BOUND_API_KEY, {
+      value: this,
+      configurable: true,
+      enumerable: false,
+      writable: false,
+    });
+  }
```
lukas-reining, line 234: This way is a transparent implementation to the providers. 
We can not directly add it to the provider interface "non-breaking" as it only is an interface.

Issue #799, opened 2024-01-31
List of breaking 2.0 changes


toddbaert:
Please use this issue to compile a list of small breaking changes for 2.0. These should not require much to absorb should generally be small typing/naming changes.

List so far:

- [ ] make it impossible to return an error resolution from providers (force them to throw if something goes wrong)
- [ ] removing setProviderAndWait and make setProvider async
- [ ] consider renaming `Client` interface to something less likely to clash (I'm not sure about this one, but @luizgribeiro wanted to consider it)
- [ ]

luizgribeiro:
> consider renaming Client interface to something less likely to clash (I'm not sure about this one, but @luizgribeiro wanted to consider it)

Giving some context on this, I was working on a project that already had an SFTP dependency that had a exported `Client`.
The idea here is not to be a breaking change, but provide an alias to OF `Client` (maybe  `OFClient`) in order to prevent this kind of clash. It wouldn't be necessary to deprecate Client though.

beeme1mr:
> Giving some context on this, I was working on a project that already had an SFTP dependency that had a exported `Client`.
> The idea here is not to be a breaking change, but provide an alias to OF `Client` (maybe `OFClient`) in order to prevent this kind of clash. It wouldn't be necessary to deprecate Client though.

That sounds like a good idea to me. A ton of libraries export clients and it can even cause issues when IDEs import from the wrong one if you're not careful.

lukas-reining:
I a meeting we decided to not do the 2.0 release yet.
The crossed out points will be considered as non breaking and done before a 2.0 release.

MattIPv4:
Assuming #1356 lands as-is, removing the deprecated `InMemoryProvider` and renaming `TypedInMemoryProvider` back to `InMemoryProvider` will be a good candidate for the 2.0 release.

PR #1344, opened 2026-02-16
feat!: Update `setProvider` to be async and remove `setProviderAndWait` #695

### LINKED ISSUE #695: [FEATURE] Update setProvider to be async and remove setProviderAndWait
## Overview

When `setProvider` was first introduced, OpenFeature didn't have support for provider initialization. It was expected that a provider was initialized prior to registering it with OpenFeature. Since then, OpenFeature has added support for initialization so that the SDK could have an awareness of the provider lifecycle and is able to emit events that developers can subscribe to. Unfortunately, this meant that a provider may not be available immediately after registration. To overcome this, developers could wait for the provider ready event to fire before performing flag evaluations. A method called `setProviderAndWait` was also added that returned a promise. The promise resolves when the provider is ready or is rejected if there's an error during initialization.

We've received feedback that `setProviderAndWait` behaves the way people expect `setProvider` to work. Also, `setProviderAndWait` is not a good name in JavaScript because it does not block a thread as the name implies. For these reasons, `setProvider` should be updated to behave like `setProviderAndWait` and `setProviderAndWait` should be removed. This will be a breaking change that will require a `v2.0.0` of the server SDK.

## Requirements

- `setProvider` returns a promise that's resolved when the provider is ready or is rejected if the provider fails during initialization.
- Remove `setProviderAndWait`
- Update the documentation in the readme

ScottVenkataraman:
## This PR
- Addresses open issue #695 
- Update `setProvider` to return a promise that's resolved when the provider is ready or is rejected if the provider fails during initialization. 
- Remove `setProviderAndWait`
- Update the documentation in the readme to reflect these changes.

### Related Issues
Fixes #695 

### Notes
This is my first PR for Open Feature, please let me know if I missed something in the contributing process or need to make any changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3485,6 +3485,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "detect-libc": "^1.0.3",
         "is-glob": "^4.0.3",
@@ -3527,6 +3528,7 @@
       "os": [
         "android"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 10.0.0"
       },
@@ -3548,6 +3550,7 @@
       "os": [
         "darwin"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 10.0.0"
       },
@@ -3569,6 +3572,7 @@
       "os": [
         "darwin"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 10.0.0"
       },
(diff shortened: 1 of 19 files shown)


lukas-reining (commented):
Thank you for the change @ScottVenkataraman!
We can not merge this immediately due to the other changes discussed for the 2.0 release. #799 
Once we get to this point, we will happily include this!
 
@open-feature/sdk-javascript-maintainers @open-feature/sdk-javascript-maintainers

Issue #1158, opened 2025-03-04
[BUG] Type soundness issue with Provider.resolveObjectEvaluation


adams85:
### Observed behavior

Currently the [TypeScript method signature of resolveObjectEvaluation](https://github.com/open-feature/js-sdk/blob/server-sdk-v1.13.5/packages/server/src/provider/provider.ts#L54-L59) makes a promise that the SDK cannot keep at run-time, which allows the caller to make mistakes.

For example:

```ts
const result = await provider.resolveObjectEvaluation('jsonFlag', { "prop": "default" }, { targetingKey }));
// At this point, the inferred compile-time type of the `result` variable is `{ "prop": string }`, which may not be in sync with the actual run-time type.
console.log(result.prop.length); // throws "TypeError: result.prop is undefined"
```

Since the default value type and the return value type is connected via the `T` type parameter, the inferred type of `result` will be `{ "prop": string }` in the example above. However, when the value resolution is successful and an actual value is returned, nothing guarantees that the resolved value has a string property named "prop". Some of the providers not even guarantee that the returned value is an object.

### Expected Behavior

In my opinion, the inferred type of `result` should be `JsonValue`, that is, the correct method signature should look like this:

```ts
resolveObjectEvaluation(flagKey: string, defaultValue: JsonValue, context: EvaluationContext, logger: Logger): Promise<ResolutionDetails<JsonValue>>;
```

This way the SDK wouldn't make a "false promise" and would force the caller to think about the return value type.

For example:

```js
const result = await provider.resolveObjectEvaluation('jsonFlag', { "prop": "default" }, { targetingKey }));
// At this point, the inferred compile-time type of the `result` variable is `JsonValue`, that is, caller needs to make checks or assertions on the actual run-time type.
if (typeof result === "object" && typeof result.prop === "string") {
  console.log(result.prop.length);
}
```

PR #1361, opened 2026-03-06
fix: Type soundness issue with Provider.resolveObjectEvaluation #1158

ScottVenkataraman:
## This PR
Updates the TypeScript method signature of `resolveObjectEvaluation` so that the `defaultValue` has a type of `JsonValue` and the returned promise's `ResolutionDetails` has a type of `JsonValue`


toddbaert:
@ScottVenkataraman This is a breaking change, and reduces the flexibility of the API. I understand there's some potential issues with the generic, but can you provide some detailed reasoning why you think this is worthwhile (especially considering that it's a breaking API change)?

Issue #1375, opened 2026-04-08
[FEATURE] Expose JS SDK/framework identity on ClientMetadata


jonathannorris:
## Context

We have run into an issue as a vendor where we don't know what OpenFeature JS SDK type our clients are running.

Because that information is not exposed by the SDK, we were forced to create simple wrapper providers such as `@devcycle/openfeature-react-provider` so we could attach that metadata to our internal telemetry and systems.

It would be better if the OpenFeature JS SDKs exposed this information directly to providers and hooks, so vendors do not need framework-specific wrapper packages purely to identify client usage.

## The Problem

Today `ClientMetadata` does not provide enough information to answer:

- is this client using the `web` or `server` SDK?
- is this call coming through `react`, `angular`, or `nest`?

Expected behavior:
- the JS SDK should expose stable metadata for SDK family and framework
- that metadata should be available anywhere `clientMetadata` is already surfaced, including hook contexts
- downstream providers should not need framework-specific wrapper packages just to infer usage

## Suggested Fix

Extend `ClientMetadata` with JS-specific identity fields:

```ts
interface ClientMetadata {
  domain?: string;
  sdk?: 'web' | 'server';
  framework?: 'react' | 'angular' | 'nest';
  version?: string;
  providerMetadata: ProviderMetadata;
}
```

Examples:
- `@openfeature/web-sdk` -> `{ sdk: 'web', version, providerMetadata }`
- `@openfeature/react-sdk` -> `{ sdk: 'web', framework: 'react', version, providerMetadata }`
- `@openfeature/angular-sdk` -> `{ sdk: 'web', framework: 'angular', version, providerMetadata }`
- `@openfeature/nestjs-sdk` -> `{ sdk: 'server', framework: 'nest', version, providerMetadata }`

## Notes

`providerMetadata` answers a different question than `sdk`/`framework`. It identifies the provider bound to the client, while `sdk` and `framework` identify the OpenFeature JS layer the application is using.

This does not require a spec change. The current spec only requires `client metadata` to expose `domain`, and SDKs can already expose additional metadata.

PR #1376, opened 2026-04-08
feat: expose js sdk identity on client metadata

jonathannorris:
## Summary

- split `ClientMetadata` into separate JS package identity and runtime identity fields
- set `sdk` to `js-web` / `js-server`, and add `paradigm` as `client` / `server`
- keep framework metadata as-is for `react`, `angular`, and `nest`
- avoid redundant Angular directive client initialization when the `domain` input already initialized the client

## Motivation

Relates to #1375.

The original version of this PR overloaded `sdk` to represent runtime identity. I think it probably makes sense to keep those concerns separate:

- `sdk` identifies the JS package family: `js-web` or `js-server`
- `paradigm` identifies the runtime model: `client` or `server`
- `framework` continues to identify wrapper SDKs like `react`, `angular`, and `nest`

This keeps the metadata more explicit and avoids conflating package family with runtime paradigm.

## Usage

```ts
const client = OpenFeature.getClient('my-domain');

client.metadata;
// web-sdk:
// {
//   domain: 'my-domain',
//   sdk: 'js-web',
//   paradigm: 'client',
//   providerMetadata: { name: '...' }
// }
```

```ts
hook.before = (hookContext) => {
  hookContext.clientMetadata;
  // react-sdk:
  // {
  //   domain: 'my-domain',
  //   sdk: 'js-web',
  //   paradigm: 'client',
  //   framework: 'react',
  //   providerMetadata: { name: '...' }
  // }
};
```

## Implementation

- `ClientMetadata` now exposes both `sdk` and `paradigm`
- the base web/server clients populate those fields directly
- the framework SDKs still use the small `Proxy` wrapper so framework metadata is visible both via `client.metadata` and from evaluations that read `this.metadata`
- Angular keeps its existing service/client behavior; I did not cache wrapped clients there because `OpenFeature.getClient()` in this SDK returns a new client instance rather than a singleton

### Notes

- the Angular directive still initializes from the `domain` input setter, but now avoids the redundant second `initClient()` call from `ngOnInit()`
- this keeps the existing initialization behavior while removing duplicate setup/teardown work

### Related Issues

- Relates to #1375


jonathannorris:
Hum... should `sdk` actually reuse [`Paradigm`](https://github.com/open-feature/js-sdk/blob/main/packages/shared/src/types/paradigm.ts) instead of introducing a new union on [`ClientMetadata`](https://github.com/open-feature/js-sdk/blob/main/packages/shared/src/client/client.ts)?

Right now this PR uses `sdk?: 'web' | 'server'`, but `Paradigm` is already `'server' | 'client'`.

I think it probably makes sense to decide whether this field is meant to represent JS package family (`web | server`) or runtime paradigm (`client | server`) before we merge this.

packages/shared/src/client/framework-metadata.ts
```diff
@@ -1,18 +1,22 @@
...
- * Wraps a web client so Angular evaluations surface framework metadata.
- * @template {Client} T
+ * Wraps a client so framework metadata is visible through `metadata` and `this.metadata`.
+ * @template T
  * @param {T} client client to wrap
+ * @param {NonNullable<ClientMetadata['framework']>} framework framework metadata to expose
  * @returns {T} framework-aware client proxy
  */
-export function withAngularFrameworkMetadata<T extends Client>(client: T): T {
+export function withFrameworkMetadata<T extends object>(
```
MattIPv4: Can we define a minimal `Client` interface here, that has `metadata` on it? Ideally, both the web + server `Clients` would then extend that minimal interface like they do other bits from the core package.

MattIPv4:
> Hum... should `sdk` actually reuse [`Paradigm`](https://github.com/open-feature/js-sdk/blob/main/packages/shared/src/types/paradigm.ts) instead of introducing a new union on [`ClientMetadata`](https://github.com/open-feature/js-sdk/blob/main/packages/shared/src/client/client.ts)?
> 
> Right now this PR uses `sdk?: 'web' | 'server'`, but `Paradigm` is already `'server' | 'client'`.
> 
> I think it probably makes sense to decide whether this field is meant to represent JS package family (`web | server`) or runtime paradigm (`client | server`) before we merge this.

Reusing makes sense to me. I think they are essentially representing the same thing? The server package sets its `runsOn` to `'server'`, and the web package sets its `runsOn` to `'client'`. Seems silly to introduce a new union to also represent the same thing with different values?

packages/angular/projects/angular-sdk/src/lib/feature-flag.directive.ts
```diff
@@ -132,7 +133,7 @@ export abstract class FeatureFlagDirective<T extends FlagValue> implements OnIni
     if (this._client) {
       this.disposeClient(this._client);
     }
-    this._client = OpenFeature.getClient(this._featureFlagDomain);
+    this._client = withAngularFrameworkMetadata(OpenFeature.getClient(this._featureFlagDomain));
```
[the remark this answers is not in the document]
jonathannorris: makes sense. updated this so ngOnInit only initializes if the domain input didn't already create the client.

packages/angular/projects/angular-sdk/src/lib/feature-flag.service.ts
```diff
@@ -217,7 +218,7 @@ export class FeatureFlagService {
     domain: string | undefined,
     options?: AngularFlagEvaluationOptions,
   ): Observable<EvaluationDetails<T>> {
-    const client = domain ? OpenFeature.getClient(domain) : OpenFeature.getClient();
+    const client = withAngularFrameworkMetadata(domain ? OpenFeature.getClient(domain) : OpenFeature.getClient());
```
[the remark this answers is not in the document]
jonathannorris: I looked at this a bit more and I don't think we should cache clients in the Angular service.

In this SDK, OpenFeature.getClient() returns a new client instance rather than a singleton for a domain, so caching here would change Angular's existing client identity/lifecycle behavior rather than just avoiding proxy allocation. I kept the service aligned with the existing Angular structure and only kept the directive double-init cleanup.

jonathannorris:
`sdk` is now JS package identity (`js-web` / `js-server`), and `paradigm` is the runtime identity (`client` / `server`).

I think this setup probably makes more sense if we ever expand this metadata shape into something more standard across other OpenFeature SDKs, since package identity and runtime paradigm won't necessarily be the same thing in every implementation.

jonathannorris:
@MattIPv4 @beeme1mr @toddbaert @lukas-reining 

I put together an alternate implementation in [#1377](https://github.com/open-feature/js-sdk/pull/1377) if you want to compare the two approaches directly.

The main difference is that this PR (#1376) uses a Proxy wrapper to inject framework metadata, while #1377 avoids Proxy entirely and instead marks the concrete SDK-owned client instances directly. Both end up exposing the same sdk / paradigm / framework metadata to providers and hooks, but the implementation tradeoff is different:

#1376: cleaner encapsulation, but framework identity comes from a wrapper layer
#1377: no Proxy, no public getClient(...) API change, and metadata comes from the real client instance, but the framework SDKs do need a small internal cast-and-mark helper

Would be good to get feedback on which direction feels cleaner to merge.

MattIPv4:
👀 Took a look at the alternative, and I think I prefer this approach with a Proxy, as it isolates the metadata change to just where we know for sure the client is being used within that framework -- your other PR mutates the actual client itself, which in the case of at least the React PR, could result in the framework metadata leaking beyond the framework?

packages/react/src/provider/provider.tsx
```diff
@@ -32,7 +33,10 @@ type ProviderProps = {
  * @returns {OpenFeatureProvider} context provider
  */
 export function OpenFeatureProvider({ client, domain, children, ...options }: ProviderProps) {
-  const stableClient = React.useMemo(() => client || OpenFeature.getClient(domain), [client, domain]);
+  const stableClient = React.useMemo(
+    () => withFrameworkMetadata(client || OpenFeature.getClient(domain), 'react'),
+    [client, domain],
+  );
```
toddbaert, line 39: One thing I want to make sure we've considered: since `@openfeature/web-sdk` is a peer dependency of the React (and Angular) SDKs, it's possible for users to import `OpenFeature.getClient()` directly from `@openfeature/web-sdk` in a React app, bypassing OpenFeatureProvider (this isn't a problem, we re-export everything, so until now these have always been exactly the same objects and imports, even in terms of `instanceof` checks, etc - they are literally nominatively the same classes, etc. 

However, with this change, there's a difference - in the case above,, sdk and paradigm would be populated, but framework would be undefined.

I think that's arguably correct; if you're not going through the framework wrapper, the metadata should reflect that. But worth calling out since a hook or provider branching on framework === 'react' could get inconsistent results depending on how the client was obtained.

WDYT? Worth a note in the docs/JSDoc, or is this fine as-is?

(Note this impacts both implementations)

toddbaert (approved):
After looking at https://github.com/open-feature/js-sdk/pull/1377, I think I prefer this, and I don't think there's much risk in the proxy after digging into it. The ref test made me nervous but I don't think it will actually have an impact in anyone's prod code.

Consider this though - not sure if [this](https://github.com/open-feature/js-sdk/pull/1376/changes#r3089122204) is a deal-breaker for you @jonathannorris (it's not for me).

packages/react/src/provider/provider.tsx
lukas-reining: As you said, I think this is okay. 

> I think that's arguably correct; if you're not going through the framework wrapper, the metadata should reflect that. But worth calling out since a hook or provider branching on framework === 'react' could get inconsistent results depending on how the client was obtained.

Yeah I think you are right and this could be confusing. 
But I think is is acceptable.

lukas-reining (approved):
This makes sense to me and looks good.

packages/react/src/provider/provider.tsx
jonathannorris: Added a JSDoc note on `ClientMetadata.framework` in 62a68dcf explicitly calling out that `framework` is only populated when the client is obtained via the corresponding framework wrapper (React provider, Angular service/directive, Nest injection), and that clients created directly from `@openfeature/web-sdk` / `@openfeature/server-sdk` will have `framework` undefined.

PR #1377, opened 2026-04-10
feat: expose js sdk identity without proxy wrappers

jonathannorris:
## Summary

- extend `ClientMetadata` with JS-specific identity fields: `sdk`, `paradigm`, and `framework`
- set `sdk` / `paradigm` in the base `web` and `server` clients
- keep the public `getClient(...)` API unchanged
- make React, Angular, and Nest mark SDK-owned clients internally so `client.metadata` and hook `clientMetadata` expose framework identity without a `Proxy`

## Motivation

Relates to #1375.

Today providers and hooks can access `clientMetadata`, but the JS SDKs do not expose enough identity to tell whether a call came from `js-web` vs `js-server`, or from `react`, `angular`, or `nest`.

This version avoids `Proxy`, keeps the existing public `getClient(...)` signatures in place, and treats framework identity as internal framework-owned metadata rather than a new public client-construction API.

## Usage

```ts
const client = OpenFeature.getClient('my-domain', '1.2.3');

client.metadata;
// {
//   domain: 'my-domain',
//   version: '1.2.3',
//   sdk: 'js-web',
//   paradigm: 'client',
//   providerMetadata: { name: '...' }
// }
```

```ts
hook.before = (hookContext) => {
  hookContext.clientMetadata;
  // react-sdk:
  // {
  //   domain: 'my-domain',
  //   sdk: 'js-web',
  //   paradigm: 'client',
  //   framework: 'react',
  //   providerMetadata: { name: '...' }
  // }
};
```

## Notes

- framework metadata is carried on the concrete web/server client instances without a wrapper layer
- the public `getClient(...)` signatures remain unchanged
- framework SDKs set `framework` internally on the SDK-owned client instances they use
- the React `client` prop path preserves the provided client instance and marks it locally so metadata stays aligned with provider-created clients

## Related Issues

- Relates to #1375


packages/angular/projects/angular-sdk/src/lib/framework-metadata.ts
```diff
@@ -0,0 +1,15 @@
...
+  setFrameworkMetadata?: (framework: 'angular') => Client;
+};
+
+/**
+ * Marks an SDK-owned web client as Angular-backed while preserving instance identity.
+ * @param {Client} client client instance to update
+ * @returns {Client} the same client instance
+ */
+export function setAngularFrameworkMetadata(client: Client): Client {
+  (client as FrameworkMetadataClient).setFrameworkMetadata?.('angular');
```
MattIPv4, line 13: Not sure I'm particularly a fan of relying on type-casts to enable this

packages/react/src/provider/provider.tsx
```diff
@@ -32,7 +36,22 @@ type ProviderProps = {
  * @returns {OpenFeatureProvider} context provider
  */
 export function OpenFeatureProvider({ client, domain, children, ...options }: ProviderProps) {
-  const stableClient = React.useMemo(() => client || OpenFeature.getClient(domain), [client, domain]);
+  const stableClient = React.useMemo(() => {
+    if (client) {
+      return setReactFrameworkMetadata(client);
```
MattIPv4, line 41: This mutates a user-provided client that _might_ also be used elsewhere, this seems problematic?

jonathannorris:
Closing in favor of #1376. Reviewer preference was to keep framework metadata isolated to the framework boundary via the Proxy wrapper, rather than marking the SDK-owned client instances directly — that avoids the risk of framework identity leaking onto a client the user may also obtain/use outside the framework (e.g. the React `client` prop path).

PR #1378, opened 2026-04-17
docs: fix inaccuracies in package READMEs

jonathannorris:
## Summary

- Fix code examples across the JS SDK monorepo package READMEs that would not compile or work correctly at runtime
- Document missing options and correct stale peer dependency versions

### `@openfeature/server-sdk`

1. **MultiProvider constructor**: Fixed bare provider arrays (`[primaryProvider, backupProvider]`) to use the correct `ProviderEntryInput` shape (`[{ provider: primaryProvider }, { provider: backupProvider }]`). The original pattern fails at runtime with `Cannot read properties of undefined (reading 'metadata')` because the internal `registerProviders` accesses `.provider` on each entry. Fixed in both "Multi-provider" and "Migration" examples
2. **Available Strategies list**: Added missing `FirstSuccessfulStrategy` and clarified the description of `FirstMatchStrategy` to contrast with it
3. **Transaction Context Propagation example**: Fixed Express middleware that read the IP from the wrong object. `res.headers.get('X-Forwarded-For')` is wrong on two counts — it uses `res` instead of `req`, and Express `req.headers` is a plain object, not a `Headers` instance. Fixed to `req.headers['x-forwarded-for']`

### `@openfeature/react-sdk`

4. **`OpenFeatureProvider` props**: Added documentation for the `client` prop (previously undocumented); noted that `domain` and `client` are mutually exclusive

### `@openfeature/angular-sdk`

5. **Module section**: Added a note explaining that the feature flag directives are standalone and not re-exported by `OpenFeatureModule` — NgModule-based apps must import the directives/pipes individually

### `@openfeature/nestjs-sdk`

6. **Peer dependency versions**: Updated stale `@openfeature/server-sdk` version from `>=1.7.5` to `^1.17.1` (matching `package.json`); corrected `rxjs` range to `^6.0.0 || ^7.0.0 || 8.0.0` (previously incorrectly listed `^8.0.0`)

All code examples verified to work correctly against the current main branch.

### DIFF
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -75,7 +75,7 @@ The following list contains the peer dependencies of `@openfeature/nestjs-sdk` w
 - `@openfeature/server-sdk`: ^1.17.1
 - `@nestjs/common`: ^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0
 - `@nestjs/core`: ^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0
-- `rxjs`: ^6.0.0 || ^7.0.0 || ^8.0.0
+- `rxjs`: ^7.0.0 || ^8.0.0
 
 The minimum required version of `@openfeature/server-sdk` currently is `1.17.1`.
 
--- a/packages/nest/README.md
+++ b/packages/nest/README.md
@@ -72,12 +72,12 @@ yarn add @openfeature/nestjs-sdk @openfeature/server-sdk @openfeature/core
 
 The following list contains the peer dependencies of `@openfeature/nestjs-sdk` with its expected and compatible versions:
 
-- `@openfeature/server-sdk`: >=1.7.5
+- `@openfeature/server-sdk`: ^1.17.1
 - `@nestjs/common`: ^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0
 - `@nestjs/core`: ^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0
 - `rxjs`: ^6.0.0 || ^7.0.0 || ^8.0.0
 
-The minimum required version of `@openfeature/server-sdk` currently is `1.7.5`.
+The minimum required version of `@openfeature/server-sdk` currently is `1.17.1`.
 
 ### Usage
 
--- a/packages/server/README.md
+++ b/packages/server/README.md
@@ -162,7 +162,10 @@ const primaryProvider = new YourPrimaryProvider();
 const backupProvider = new YourBackupProvider();
 
 // Create multi-provider with a strategy
-const multiProvider = new MultiProvider([primaryProvider, backupProvider], new FirstMatchStrategy());
+const multiProvider = new MultiProvider(
+  [{ provider: primaryProvider }, { provider: backupProvider }],
+  new FirstMatchStrategy(),
+);
 
 // Register the multi-provider
 await OpenFeature.setProviderAndWait(multiProvider);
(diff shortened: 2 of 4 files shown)


lukas-reining (approved):

packages/nest/README.md
```diff
@@ -72,12 +72,12 @@ yarn add @openfeature/nestjs-sdk @openfeature/server-sdk @openfeature/core
 
 The following list contains the peer dependencies of `@openfeature/nestjs-sdk` with its expected and compatible versions:
 
-- `@openfeature/server-sdk`: >=1.7.5
+- `@openfeature/server-sdk`: ^1.17.1
```
lukas-reining, line 75: Why did you do this?

packages/nest/README.md
```diff
@@ -72,12 +72,12 @@ yarn add @openfeature/nestjs-sdk @openfeature/server-sdk @openfeature/core
...
 - `@nestjs/common`: ^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0
 - `@nestjs/core`: ^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0
-- `rxjs`: ^6.0.0 || ^7.0.0 || ^8.0.0
+- `rxjs`: ^6.0.0 || ^7.0.0 || 8.0.0
```
lukas-reining: `8.0.0` without caret seems wrong, any reason you did that?

packages/nest/README.md
beeme1mr: I wonder if we could try to automate this by looking at the package.json. It seems like something we could easily forget about. Obviously, out of scope of this PR and doesn't answer your quesiton though. I'm also not sure why the version was bumped. Perhap it was because of the new typings that were recently added.

packages/nest/README.md
beeme1mr: It looks like Gemini recommended the change but I disagree with its recommendation.

https://github.com/open-feature/js-sdk/pull/1378/changes#r3100902912

packages/server/README.md
```diff
@@ -357,7 +361,7 @@ OpenFeature.setTransactionContextPropagator(new AsyncLocalStorageTransactionCont
  */
 const app = express();
 app.use((req: Request, res: Response, next: NextFunction) => {
-  const ip = res.headers.get('X-Forwarded-For');
+  const ip = req.headers['x-forwarded-for'];
```
beeme1mr, line 364: Wow, good catch.

beeme1mr (changes requested):
Most of the changes look great but please look at @lukas-reining before merging. Thanks!

packages/nest/README.md
jonathannorris: The agent changed these because the README was out of sync with what `packages/nest/package.json` actually declares (`@openfeature/server-sdk: ^1.17.1`). The `>=1.7.5` in the README was stale.

@beeme1mr +1 on automating this from `package.json` in a follow-up — would avoid exactly this drift.

packages/nest/README.md
jonathannorris: Reverted the README back to `^8.0.0`. The agent aligned it with `packages/nest/package.json`, which currently declares `rxjs: ^6.0.0 || ^7.0.0 || 8.0.0` (no caret on `8.0.0`). That missing caret in `package.json` is probably a mistake and should be fixed too.

beeme1mr (approved):

juanparadox (approved):

toddbaert (approved):

PR #1379, opened 2026-04-21
fix(nest): correct rxjs 8.0.0 peer, remove rxjs 6 support (EOL)

jonathannorris:
## Summary

- Add missing caret to the `rxjs` peer dependency range in `packages/nest/package.json`

## Context

While reviewing the README in #1378, the `rxjs` peer dependency range was noted as inconsistent: `"^6.0.0 || ^7.0.0 || 8.0.0"` — the first two versions use caret ranges, but `8.0.0` is an exact version. That restricts compatibility to exactly rxjs `8.0.0` and excludes any future rxjs `8.x` patch/minor releases, which is almost certainly not the intent.

This PR aligns the range with the caret convention used for the other versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16264,7 +16264,11 @@
         "@angular/build": "^21.0.3",
         "@angular/cli": "^21.0.3",
         "@angular/common": "^21.0.4",
+<<<<<<< HEAD
         "@angular/compiler": "^21.1.0",
+=======
+        "@angular/compiler": "^21.0.4",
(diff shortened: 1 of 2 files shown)


beeme1mr (approved):
Approving but please review Gemini's feedback before merging.

toddbaert (approved):

toddbaert:
I removed the rxjs 6 compatibility. I've confirmed it's EOL, and our policy is removing support for EOL frameworks/libs doesn't constitute a breaking change.

PR #1402, opened 2026-04-28
chore: resolve open dependabot security alerts

jonathannorris:
## Summary

- Resolved open Dependabot security alerts by bumping direct dev dependencies and pinning patched transitive versions in the lockfile
- No `overrides` in `package.json` — patched versions are held via the lockfile
- Two alerts remain open: [alert 130](https://github.com/open-feature/js-sdk/security/dependabot/130) (`@tootallnate/once`, low) and [alert 165](https://github.com/open-feature/js-sdk/security/dependabot/165) (`uuid`, medium) — both require breaking changes to resolve


package.json
```diff
@@ -70,13 +70,63 @@
...
     "shx": "^0.4.0",
-    "ts-jest": "^29.1.2",
+    "ts-jest": "^29.4.9",
     "ts-node": "^10.9.2",
     "tslib": "^2.3.0",
     "typedoc": "^0.26.0",
     "typescript": "^4.7.4",
     "uuid": "^11.0.0"
   },
+  "overrides": {
```
toddbaert: Please remove the overrides - I think most of these can be avoided with updates to the tools that transitively require them, but that would require some manual testing, which I'm happy to do later this week.

MattIPv4: If the packages that rely on them don't have updates, you can likely set the overrides locally, do an install, and then remove the overrides and the bumped versions will remain in the lockfiles.

toddbaert (approved):
I'll investigate on Friday if any dev tooling can be updated to further improve the transitive deps that were previously overridden. 

Thanks @jonathannorris ❤️

PR #1403, opened 2026-05-05
fix: Extract request object from correct execution context

### LINKED ISSUE #1369: [FEATURE] NestJS Graphql RequireFlagsEnabled Decorator Support
### Requirements

The nestjs-sdk RequireFlagsEnabled decorator does not support graphql contexts. The `context.switchToHttp().getRequest()` won't get you the http request with NestJS + Fastify + Graphql.  You should check the `host.getType()` for `http` or `graphql` and switch accordingly.

Or, since the context is really only used for creating the default error, you could remove it entirely and make a more generic error. Something like `cannot access this endpoint` would be less informative, but not rely on the correct request object.

benkenawell:
## This PR

Adds graphql context support to the RequireFlagsEnabled decorator in the nestjs package.

Unfortunately, if using the decorator on a graphql resolver or operation, the request object lives in a slightly different place.  This PR adds @nestjs/graphql as an optional peer dependency and adds support in the RequireFlagsEnabled decorator for the graphql context.  If you aren't using nestjs/graphql, this should still work fine. If you are, it should also transparently work.

### Related Issues

Fixes #1369 

### Follow-up Tasks

There may be other places where the request object is accessed, those should also use the `getRequestFromContext` helper function I've added to the `utils.ts` file.

### How to test

npm ci and npm test should cover it.

### DIFF
--- a/packages/nest/src/utils.ts
+++ b/packages/nest/src/utils.ts
@@ -1,6 +1,16 @@
 import { withFrameworkMetadata } from '@openfeature/core';
 import type { Client, EvaluationContext } from '@openfeature/server-sdk';
 import { OpenFeature } from '@openfeature/server-sdk';
+import { ExecutionContext } from '@nestjs/common';
+import type { GqlExecutionContext as GqlExecutionContextClass, GqlContextType } from '@nestjs/graphql';
+
+let GqlExecutionContext: typeof GqlExecutionContextClass | undefined;
+try {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  ({ GqlExecutionContext } = require('@nestjs/graphql') as typeof import('@nestjs/graphql'));
+} catch {
+  // @nestjs/graphql is an optional peer dependency
+}
 
 /**
  * Returns a domain scoped or the default OpenFeature client with the given context.
@@ -14,3 +24,15 @@ export function getClientForEvaluation(domain?: string, context?: EvaluationCont
     'nest',
   );
 }
+
+/**
+ * @param {ExecutionContext} context NestJS's execution context
+ * @returns request object
+ */
+export const getRequestFromContext = <T = any>(context: ExecutionContext): T | undefined => {
+  if (context.getType() === 'http') {
+    return context.switchToHttp().getRequest<T>();
+  } else if (context.getType<GqlContextType>() === 'graphql') {
+    return GqlExecutionContext?.create(context)?.getContext()?.req as T;
+  }
+};
--- a/packages/nest/src/utils.ts
+++ b/packages/nest/src/utils.ts
@@ -1,12 +1,12 @@
 import { withFrameworkMetadata } from '@openfeature/core';
 import type { Client, EvaluationContext } from '@openfeature/server-sdk';
 import { OpenFeature } from '@openfeature/server-sdk';
-import { ExecutionContext } from '@nestjs/common';
+import type { ExecutionContext } from '@nestjs/common';
 import type { GqlExecutionContext as GqlExecutionContextClass, GqlContextType } from '@nestjs/graphql';
 
 let GqlExecutionContext: typeof GqlExecutionContextClass | undefined;
 try {
-  // eslint-disable-next-line @typescript-eslint/no-require-imports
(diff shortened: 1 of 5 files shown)


packages/nest/src/utils.ts
```diff
@@ -14,3 +24,15 @@ export function getClientForEvaluation(domain?: string, context?: EvaluationCont
...
+
+/**
+ * @param {ExecutionContext} context NestJS's execution context
+ * @returns request object
+ */
+export const getRequestFromContext = (context: ExecutionContext): any => {
+  if (context.getType() === 'http') {
+    return context.switchToHttp().getRequest();
+  } else if (context.getType<GqlContextType>() === 'graphql') {
+    return GqlExecutionContext?.create(context).getContext().req;
```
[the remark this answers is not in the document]
benkenawell: I've added these, thanks for the suggestion

packages/nest/src/utils.ts
```diff
@@ -1,6 +1,16 @@
 import { withFrameworkMetadata } from '@openfeature/core';
 import type { Client, EvaluationContext } from '@openfeature/server-sdk';
 import { OpenFeature } from '@openfeature/server-sdk';
+import { ExecutionContext } from '@nestjs/common';
+import type { GqlExecutionContext as GqlExecutionContextClass, GqlContextType } from '@nestjs/graphql';
```
beeme1mr, line 5: I know this is just importing the types but could this be an issue if `@nestjs/graphql` isn't available?

beeme1mr (commented):
Thanks for the PR @benkenawell. It looks good from what I can tell. Would it be possible to add a test?

benkenawell:
yes! It might be a few days before I get the chance, but I would be happy to write a few tests.  I believe the types won't be an issue, but I will double check that by importing into a new nest project as well!

benkenawell:
@beeme1mr I added some unit tests for the getRequestFromContext function.  I think I may need to add a graphql driver like [apollo](https://docs.nestjs.com/graphql/quick-start) to add proper integration tests for graphql. If you're okay with that, I'm happy to write those tests as well.

packages/nest/src/utils.ts
benkenawell: I created [this branch](https://github.com/benkenawell/nestjs-openfeature-testing/tree/gql-bugfix-test) as a little sample repository for showing that this shouldn't be an issue if `@nestjs/graphql` isn't available.  I have my changes vendored in, so you could clone and install and try it if you want. (I realize that sounds pretty insecure. Changing the install path for those packages to this PR would also work. I only vendored them in to make it easier to share)

beeme1mr (approved):
Looks good to me. Thanks @benkenawell 

@lukas-reining would you mind taking a look when you have a moment?

lukas-reining:
Sorry for the delay, I am on a conference right now and will try to make sure to look latest next week @beeme1mr @benkenawell!

packages/nest/src/utils.ts
```diff
@@ -1,6 +1,16 @@
 import { withFrameworkMetadata } from '@openfeature/core';
 import type { Client, EvaluationContext } from '@openfeature/server-sdk';
 import { OpenFeature } from '@openfeature/server-sdk';
+import type { ExecutionContext } from '@nestjs/common';
+import type { GqlExecutionContext as GqlExecutionContextClass, GqlContextType } from '@nestjs/graphql';
+
+let GqlExecutionContext: typeof GqlExecutionContextClass | undefined;
+try {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/consistent-type-imports
+  ({ GqlExecutionContext } = require('@nestjs/graphql') as typeof import('@nestjs/graphql'));
```
toddbaert, line 10: I think this won't ESM build; we'll get "Dynamic require not supported" in true ESM and the try/catch silently swallows it. Could you switch to `await import('@nestjs/graphql')`? intercept() is already async so it should be pretty straightforward unless I'm missing something.

toddbaert (changes requested):
Thanks for this, and sorry for the delayed review. I think you're right.

Unfortunately [this](https://github.com/open-feature/js-sdk/pull/1403/changes#r3443436056) is a blocker I think. If you disagree, let me know.

PR #1409, opened 2026-06-09
feat(angular): add provideOpenFeature() standalone provider function

Krishan27:
## This PR
Adds `provideOpenFeature()` as the modern, recommended way to configure OpenFeature in standalone Angular apps, and deprecates `OpenFeatureModule` without removing it.

- adds this new feature

### Related Issues

Fixes #1365


lukas-reining (approved):
This looks great and I found noting to change about it! 
Thank you so much again @Krishan27!

PR #1430, opened 2026-06-19
fix: reset API state on close

### LINKED ISSUE #1374: shutdown/close doesn't fully reset API state
The spec (requirement 1.6.2) says shutdown MUST reset all API state, removing hooks, event handlers, and providers. There's also a pending clarification to explicitly include evaluation context and transaction context propagators (open-feature/spec#375).

Right now, `close()` only shuts down providers but doesn't clear or reset them. `clearProviders()` resets providers to no-op but doesn't touch hooks, event handlers, evaluation context, or transaction context propagators. Neither function fully satisfies 1.6.2.

Found this while doing a cross-SDK audit for Python 1.0 readiness.

Krishan27:
Fixes #1374.

Updates the `close()` shutdown path so it fully resets API-managed state after provider shutdown, including providers, hooks, event handlers, global evaluation context, and transaction-context propagator state.

`clearProviders()` remains provider-only for backward compatibility, matching maintainer guidance.

### DIFF
--- a/packages/shared/src/open-feature.ts
+++ b/packages/shared/src/open-feature.ts
@@ -395,13 +395,18 @@ export abstract class OpenFeatureCommonAPI<
   }
 
   async close(): Promise<void> {
-    await this._shutdownAllProviders();
-    this._hooks = [];
-    this._context = {};
-    this._domainScopedContext.clear();
-    this._clientEventHandlers.clear();
-    this._clientEvents.clear();
-    this._apiEmitter.removeAllHandlers();
+    try {
+      await this._shutdownAllProviders();
+    } catch (err) {
+      this._logger.error('Unable to cleanly close providers. Resetting state.');
+    } finally {
+      this._hooks = [];
+      this._context = {};
+      this._domainScopedContext.clear();
+      this._clientEventHandlers.clear();
+      this._clientEvents.clear();
+      this._apiEmitter.removeAllHandlers();
+    }
   }
 
   protected async clearProvidersAndSetDefault(defaultProvider: P): Promise<void> {
--- a/packages/shared/src/open-feature.ts
+++ b/packages/shared/src/open-feature.ts
@@ -395,6 +395,31 @@ export abstract class OpenFeatureCommonAPI<
   }
 
   async close(): Promise<void> {
+    await this._shutdownAllProviders();
+    this._hooks = [];
+    this._context = {};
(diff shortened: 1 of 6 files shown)


MattIPv4 (approved):

Krishan27:
@MattIPv4  @beeme1mr  Thank you for review,  @lukas-reining  please review and If there is nothing else to edit in this PR we might can merge it and close it. Thank you again.

packages/shared/src/open-feature.ts
```diff
@@ -395,28 +395,18 @@ export abstract class OpenFeatureCommonAPI<
...
-    await Promise.all(
-      wrappers.map(async ([, wrapper]) => {
-        try {
-          await wrapper?.provider.onClose?.();
-        } catch (err) {
-          this.handleShutdownError(wrapper?.provider, err);
-        }
-      }),
-    );
+    await this._shutdownAllProviders();
```
lukas-reining: If `handleShutdownError` in `_shutdownAllProviders` throws, the state will not be reset. 
In `clearProvidersAndSetDefault` we have catch for this case.

lukas-reining (commented):
This looks great beside one thought. Thank you @Krishan27!

packages/shared/src/open-feature.ts
Krishan27: @lukas-reining Good catch, thanks!  Now I just Wrapped _shutdownAllProviders() in close() with the same try/catch/finally pattern used in clearProvidersAndSetDefault(), so state is always reset even if provider shutdown throws.

toddbaert: Seems fixed, but there's no explicit test for this; probably worth adding.

toddbaert (approved):
Thanks @Krishan27 ... looks good, though I left a small recommendation for an added test.

I want to point out that for this to work as advertised, we need to release the `core` module first and bump the corresponding peers in `server/web` so that they actually are forced to use the improvements in this version. Failing do to so won't break anything, but the improvement won't really work.

I think @lukas-reining 's issues are resolved, so I will merge in the next few days if I don't hear objections.

Krishan27:
Thanks @toddbaert will try to add the recommended test. 

Good to know on the core release + server/web peer bump sequencing too, appreciate the heads up so I don't assume this is active until that's done.

PR #1433, opened 2026-06-26
feat(core): pass bound domain to provider initialize and enforce domain-scoped binding

### LINKED ISSUE #1434: Supply the bound domain to provider initialization
The spec now supplies the bound `domain` to a provider's `initialize` function, and adds an opt-in `domain-scoped` provider declaration that the API enforces by binding such a provider to at most one `domain`. This lets stateful providers (e.g. an OFREP static-context provider with a persistent cache) scope their state to a single, unambiguous `domain`.

### What needs to happen

- Pass the bound `domain` (if any) to the provider's `initialize` when it's registered via the provider mutator
- Extend the provider `initialize` signature to accept the optional `domain` (non-breaking, additional optional parameter)
- Add a way for a provider to declare itself `domain-scoped`
- Enforce that a `domain-scoped` provider is bound to at most one `domain`, rejecting any attempt to bind an already-bound instance to another `domain`
- Associated testing and docs

### Status

In progress: #1433

### Links

- Parent issue: https://github.com/open-feature/spec/issues/403
- Spec sections: [1.1 Flag evaluation](https://openfeature.dev/specification/sections/flag-evaluation), [2.4 Initialization](https://openfeature.dev/specification/sections/providers)

jonathannorris:
## Summary

- Add optional `domain` parameter to provider `initialize` and pass the bound domain from the provider mutator (spec 1.1.2.2, 2.4.1)
- Add `domainScoped` provider declaration and reject binding a domain-scoped instance to more than one domain (spec 2.4.3, 1.1.8.1)
- Forward `domain` through web and server MultiProvider `initialize`
- Add conformance tests in web and server SDK test suites, including legacy single-argument `initialize` backward compatibility

## Motivation

Unblocks OFREP static-context providers that need the bound `domain` at init time to scope persisted cache keys per [open-feature/spec#393](https://github.com/open-feature/spec/pull/393) and [protocol ADR 0009](https://github.com/open-feature/protocol/blob/main/service/adrs/0009-local-storage-for-static-context-providers.md).

### Notes

- Non-breaking: `domain` is an optional second parameter to `initialize`
- Non-domain-scoped providers retain existing behavior (single instance can back multiple domains; `initialize` runs once)
- Legacy providers that only implement `initialize(context)` remain compatible; the extra `domain` argument is ignored if unused

### Related Issues

Closes #1434

Relates to: [open-feature/spec#393](https://github.com/open-feature/spec/pull/393)

## Test plan

- [x] `npx jest --selectProjects=shared --selectProjects=web --selectProjects=server`

### DIFF
--- a/packages/server/test/multi-provider.spec.ts
+++ b/packages/server/test/multi-provider.spec.ts
@@ -22,6 +22,7 @@ import {
   FirstSuccessfulStrategy,
   ComparisonStrategy,
 } from '../src';
+import { legacyInitTestProvider } from '../../shared/test/legacy-initialize-provider';
 
 class TestProvider implements Provider {
   public metadata: ProviderMetadata = {
@@ -42,28 +43,6 @@ class TestProvider implements Provider {
   emitEvent(type: ServerProviderEvents) {
     this.events.emit(type, { providerName: this.metadata.name });
   }
-}
-
-/** Child provider with a single-argument initialize that ignores any extra arguments. */
-class LegacyInitTestProvider implements Provider {
-  public metadata: ProviderMetadata = {
-    name: 'LegacyInitTestProvider',
-  };
-  public events = new OpenFeatureEventEmitter();
-  public hooks: Hook[] = [];
-  public track = jest.fn();
-  public initializeCalls = 0;
-  public lastContext?: EvaluationContext;
-
-  async initialize(context?: EvaluationContext): Promise<void> {
-    this.lastContext = context;
-    this.initializeCalls++;
-  }
-
-  resolveBooleanEvaluation = jest.fn().mockResolvedValue({ value: false });
-  resolveStringEvaluation = jest.fn().mockResolvedValue({ value: 'default' });
-  resolveObjectEvaluation = jest.fn().mockResolvedValue({ value: {} });
-  resolveNumberEvaluation = jest.fn().mockResolvedValue({ value: 0 });
 }
 
 const callEvaluation = async (multi: MultiProvider, context: EvaluationContext, logger: Logger) => {
@@ -185,16 +164,25 @@ describe('MultiProvider', () => {
     });
 
     it('forwards domain to legacy single-argument child providers without error', async () => {
-      const provider1 = new LegacyInitTestProvider();
-      const provider2 = new LegacyInitTestProvider();
-      const multiProvider = new MultiProvider([{ provider: provider1 }, { provider: provider2 }]);
+      const legacyProvider1 = legacyInitTestProvider(
+        { runsOn: 'server', asyncResolvers: true, name: 'LegacyInitTestProvider' },
+        { events: new OpenFeatureEventEmitter() },
+      );
+      const legacyProvider2 = legacyInitTestProvider(
+        { runsOn: 'server', asyncResolvers: true, name: 'LegacyInitTestProvider' },
+        { events: new OpenFeatureEventEmitter() },
+      );
+      const multiProvider = new MultiProvider([
+        { provider: legacyProvider1 as unknown as Provider },
+        { provider: legacyProvider2 as unknown as Provider },
+      ]);
 
       await multiProvider.initialize({ targetingKey: 'user' }, 'my-domain');
 
-      expect(provider1.initializeCalls).toBe(1);
-      expect(provider2.initializeCalls).toBe(1);
-      expect(provider1.lastContext).toEqual({ targetingKey: 'user' });
-      expect(provider2.lastContext).toEqual({ targetingKey: 'user' });
+      expect(legacyProvider1.initializeCalls).toBe(1);
+      expect(legacyProvider2.initializeCalls).toBe(1);
+      expect(legacyProvider1.lastContext).toEqual({ targetingKey: 'user' });
+      expect(legacyProvider2.lastContext).toEqual({ targetingKey: 'user' });
     });
 
     it('throws error if a provider errors on initialization', async () => {
--- a/packages/server/test/multi-provider.spec.ts
+++ b/packages/server/test/multi-provider.spec.ts
@@ -44,6 +44,28 @@ class TestProvider implements Provider {
   }
 }
 
+/** Child provider with a single-argument initialize that ignores any extra arguments. */
+class LegacyInitTestProvider implements Provider {
+  public metadata: ProviderMetadata = {
+    name: 'LegacyInitTestProvider',
+  };
+  public events = new OpenFeatureEventEmitter();
+  public hooks: Hook[] = [];
+  public track = jest.fn();
+  public initializeCalls = 0;
+  public lastContext?: EvaluationContext;
+
+  async initialize(context?: EvaluationContext): Promise<void> {
+    this.lastContext = context;
+    this.initializeCalls++;
+  }
+
+  resolveBooleanEvaluation = jest.fn().mockResolvedValue({ value: false });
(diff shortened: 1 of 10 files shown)


packages/shared/src/open-feature.ts
```diff
@@ -252,8 +256,9 @@ export abstract class OpenFeatureCommonAPI<
 
     // initialize the provider if it implements "initialize" and it's not already registered
     if (typeof provider.initialize === 'function' && !this.allProviders.includes(provider)) {
+      const initContext = domain ? (this._domainScopedContext.get(domain) ?? this._context) : this._context;
       initializationPromise = provider
-        .initialize?.(domain ? (this._domainScopedContext.get(domain) ?? this._context) : this._context)
+        .initialize?.(initContext, domain)
         ?.then(() => {
```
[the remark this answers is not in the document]
jonathannorris: bot: The truthy `domain` checks are pre-existing; this PR only forwards the same value as the new second `initialize` argument.

packages/shared/src/provider/provider.ts
```diff
@@ -123,9 +129,10 @@ export interface CommonProvider<S extends ClientProviderStatus | ServerProviderS
    * When the returned promise resolves, the SDK fires the ProviderEvents.Ready event.
    * If the returned promise rejects, the SDK fires the ProviderEvents.Error event.
    * Use this function to perform any context-dependent setup within the provider.
-   * @param context
+   * @param context the global evaluation context
+   * @param domain the bound domain, if any
    */
-  initialize?(context?: EvaluationContext): Promise<void>;
+  initialize?(context?: EvaluationContext, domain?: string): Promise<void>;
```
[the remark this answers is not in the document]
jonathannorris: bot: The context argument behavior is unchanged in this PR; we only added the optional `domain` second parameter. Domain-bound registration already passed domain-scoped context (falling back to global) before this change.

packages/server/test/multi-provider.spec.ts
```diff
@@ -162,6 +163,28 @@ describe('MultiProvider', () => {
       expect(initializations).toBe(2);
     });
 
+    it('forwards domain to legacy single-argument child providers without error', async () => {
```
MattIPv4, line 166: Should there also be coverage for forwarding the domain?

packages/web/test/multi-provider-web.spec.ts
```diff
@@ -162,6 +163,28 @@ describe('MultiProvider', () => {
       expect(initializations).toBe(2);
     });
 
+    it('forwards domain to legacy single-argument child providers without error', async () => {
```
MattIPv4, line 166: Ditto, coverage for forwarding the domain?

MattIPv4 (commented):
How does the MultiProvider handle domain-scoped providers? Or is wrapping a provider in MultiProvider essentially a way to bypass that restriction?

packages/server/test/multi-provider.spec.ts
jonathannorris: Added `forwards domain to child provider initialize` in both specs (`199f03b7`). Asserts both children receive `(context, 'my-domain')`.

packages/web/test/multi-provider-web.spec.ts
jonathannorris: Added `forwards domain to child provider initialize` in both specs (`199f03b7`). Asserts both children receive `(context, 'my-domain')`.

jonathannorris:
@MattIPv4 MultiProvider forwards `(context, domain)` to children; it doesn't enforce `domainScoped`. The SDK guard only runs on direct `setProvider(domain, provider)` binding, so wrapping in MultiProvider doesn't bypass that.

Reusing the same domain-scoped child across multiple MultiProviders isn't caught at registration (children aren't in `allProviders`); the child should reject a second domain in `initialize`. Happy to add a conformance test if useful.

packages/shared/src/provider/provider.ts
beeme1mr, line 135: I wonder if we should make this an initialization config object so we can extend it in the future without breaking anything.

toddbaert: I don't think it matters as much here due to overloading, but it's possible and maybe a good idea. We could have an `InitializationContext` (we already have a [HookContext](https://github.com/open-feature/js-sdk/blob/bc470dca88afa26462b2513e2d106df324e7ba46/packages/shared/src/hooks/hooks.ts#L9) that is comparable). "Context" is already a bit overloaded, but this is mostly a provider author interface so I'm not that worried about that.

packages/shared/test/legacy-initialize-provider.ts
```diff
@@ -0,0 +1,73 @@
...
+    resolveStringEvaluation: mockValue({ value: '' }),
+    resolveNumberEvaluation: mockValue({ value: 0 }),
+    resolveObjectEvaluation: mockValue({ value: {} }),
+  };
+}
+
+/**
+ * Provider with a single-argument initialize that ignores any extra arguments passed by the SDK.
+ * Pass optional extras when MultiProvider needs events, hooks, or track stubs on the child.
+ */
```
toddbaert, line 50:
```suggestion
/**
 * Provider with a single-argument initialize that ignores any extra arguments passed by the SDK.
 * Pass optional extras when MultiProvider needs events, hooks, or track stubs on the child.
 * @param options provider configuration (paradigm, name, resolver mode)
 * @param extras optional MultiProvider child stubs (events, hooks, track)
 * @returns a single-argument-initialize provider
 */
```

Fixes a lint.

packages/shared/src/open-feature.ts
```diff
@@ -241,6 +247,10 @@ export abstract class OpenFeatureCommonAPI<
       throw new GeneralError(`Provider '${provider.metadata.name}' is intended for use on the ${provider.runsOn}.`);
     }
 
+    if (provider.domainScoped && this.allProviders.includes(provider)) {
+      throw new GeneralError(`Cannot bind domain-scoped provider '${provider.metadata.name}' to more than one domain.`);
+    }
+
```
toddbaert: Nit: I think the "domainScoped" check could be consolidated into a single upfront check that inspects both `_defaultProvider` and `_domainScopedProviders` directly for this instance and rejects if any binding differs from the current one, letting the no-op branch stay focused on being a no-op. Just for consideration; happy either way... I think it would be more lines of code, but it would lock it to a single place...  maybe extracted to a method.

Totally optional, only do it if you agree.

toddbaert (approved):
I'm approving because I see no fundamental problems, but consider mine and other's nits, and since this is a fundamentally new thing, let's leave it open for a few days at least.

I'm somewhat interested in doing what @beeme1mr said [here](https://github.com/open-feature/js-sdk/pull/1433#discussion_r3516049353), but I don't feel 100% sure. Interested in the opinions of others.

packages/shared/src/provider/provider.ts
jonathannorris: I don't think we need an object here; overloading gives us room to extend without breaking existing `initialize(context)` implementations. I'd keep positional `(context?, domain?)` for this PR per spec#393.

If we do move to an options bag for the second arg later, I'd avoid another `*Context` name. Some options: `ProviderBinding`, `ProviderRegistration`, `ProviderScope`, `InitializeMetadata`.

toddbaert: Ya I think my recommendation is we don't do this, but I'm open to disagreements.

lukas-reining (approved):
Looks good to me!

PR #1455, opened 2026-09-03
feat(svelte): add @openfeature/svelte-sdk

pharv:
## Summary

- Add `@openfeature/svelte-sdk` (`packages/svelte`), an official Svelte binding for the Web SDK that mirrors the surface of `@openfeature/react-sdk`
- Reactive flag evaluation via `useFlag` and the typed `use{Boolean,String,Number,Object}Flag{Value,Details}` functions, re-evaluating on `Ready`, `ContextChanged` and `ConfigurationChanged` (honouring `flagsChanged`) and only notifying effects when the evaluation details actually changed
- `setOpenFeatureScope({ domain | client, ...options })` binds a client and default evaluation options to a component subtree via Svelte context (the `<OpenFeatureProvider>` equivalent); without a scope, the default client is used
- Reactive provider status and readiness (`useOpenFeatureClientStatus`, `useWhenProviderReady`), plus `useContextMutator` and `useTrack`
- `setOpenFeatureTestScope({ flagValueMap, delayMs, provider, domain })` testing helper (the `<OpenFeatureTestProvider>` equivalent)
- Vitest test suite (rune-level reactivity via `$effect.root` and rendered components), README, and workspace/release/CI wiring; `'svelte'` added to the `framework` union in `@openfeature/core`

Usage:

```svelte
<script lang="ts">
  import { OpenFeature, TypedInMemoryProvider, setOpenFeatureScope, useFlag } from '@openfeature/svelte-sdk';

  OpenFeature.setProvider(new TypedInMemoryProvider(flagConfig));
  setOpenFeatureScope({ domain: 'my-domain' });

  const newMessage = useFlag('new-message', false);
</script>

{#if newMessage.value}
  <p>Welcome to this OpenFeature-enabled Svelte app!</p>
{/if}
```

## Motivation

Svelte is on the OpenFeature roadmap but has no framework SDK, so every Svelte app currently hand-rolls the same reactive wrapper around `@openfeature/web-sdk` (re-evaluate on provider/context/config events, gate on readiness). This closes that gap the same way the React and Angular SDKs do for their frameworks. It is distinct from Vercel's Flags SDK SvelteKit adapter, which is server-side only and a different abstraction.

### Notes

- **No runes in library code.** Reactivity is built on [`createSubscriber`](https://svelte.dev/docs/svelte/svelte-reactivity#createSubscriber) from `svelte/reactivity`, the pattern Svelte recommends for external event sources. Getters register effect dependencies; web-sdk handlers are attached while something depends on the object and removed (via `AbortController`) when the last effect is destroyed. The package is therefore plain JS: it builds with the existing esbuild + rollup-dts pipeline, needs no compilation by consumers, ships no `.svelte` files, and flag objects can live in components, `.svelte.ts` modules, or module scope without manual cleanup. Peer dependency is `svelte: ^5.7.0` (when `createSubscriber` landed); Svelte 4 is out of scope, `toStore(() => flag.value)` is documented as the bridge.
- **Primitives are returned as `{ current }` boxes** (`useBooleanFlagValue(...).current`, `useWhenProviderReady().current`), following the convention of Svelte's own reactive classes, since a bare primitive can't be reactive.
- **Scope falls back to the default client** instead of throwing like React's `MissingContextError`, because Svelte 5 code legitimately creates flag objects outside components (module-level state), where no context can exist.
- **Naming:** `setOpenFeatureScope` was chosen over "provider"/"context" to avoid colliding with OpenFeature's own *provider* and *evaluation context* terms. Happy to rename.
- **SvelteKit SSR** is documented as out of scope (as with the web SDK); the README points to the Server SDK for `load` functions.
- Suspense options and the declarative `<FeatureFlag>` component have no counterpart (`{#if flag.value}` is native); a component could follow later if there's demand.
- **ESM-only.** Svelte 5 is ESM-only (no `require` condition on `svelte/reactivity`), so a CommonJS build would fail with `ERR_REQUIRE_ESM` on Node versions without `require(esm)`. Like the Angular SDK, this package publishes ESM only.
- `prettier-plugin-svelte` is added as a root devDependency so the `.svelte` test fixtures are formatted like everything else.
- Tests use Vitest (`@sveltejs/vite-plugin-svelte`, `@testing-library/svelte`, jsdom) like the Angular package, wired in as `npm run test:svelte`. Package version starts at `0.0.0` so the first release-please cut is `0.1.0`.

### Follow-up Tasks

- Docs site page under `client/web/svelte`, and an example in `js-sdk-examples`

## Test plan

- [x] `npm run build`
- [x] `npm run test:svelte` (58 tests, 98% coverage)
- [x] `npm run test:jest` (612 tests) and `npm run test:package-exports`
- [x] `npm run lint` and `npm run format`

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -43,6 +43,9 @@ jobs:
       - name: Test Angular SDK
         run: npm run test:angular
 
+      - name: Test Svelte SDK
+        run: npm run test:svelte
+
   format-lint:
     runs-on: ubuntu-latest
 
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,5 @@
 {
   "singleQuote": true,
-  "printWidth": 120
+  "printWidth": 120,
+  "plugins": ["prettier-plugin-svelte"]
 }
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -4,5 +4,6 @@
   "packages/web": "1.10.0",
   "packages/server": "1.23.0",
   "packages/shared": "1.12.0",
-  "packages/angular/projects/angular-sdk": "1.3.2"
+  "packages/angular/projects/angular-sdk": "1.3.2",
+  "packages/svelte": "0.0.0"
 }
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ For details, including API documentation, see the respective README files.
   - [NestJS SDK](./packages/nest/README.md), a distribution of the Server SDK with built-in NestJS-specific features.
 - [Web SDK](./packages/web/README.md), for use in the web browser.
   - [React SDK](./packages/react//README.md), a distribution of the Web SDK with built-in React-specific features.
+  - [Svelte SDK](./packages/svelte/README.md), a distribution of the Web SDK with built-in Svelte-specific features.
 
 Each have slightly different APIs, but share many underlying types and components.
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,8 @@
         "packages/react",
         "packages/angular",
         "packages/angular/projects/angular-sdk",
-        "packages/nest"
+        "packages/nest",
+        "packages/svelte"
       ],
       "devDependencies": {
         "@angular/compiler": "^21.2.19",
@@ -47,6 +48,7 @@
         "jest-junit": "^17.0.0",
         "lint-staged": "^16.2.7",
         "prettier": "^3.7.4",
+        "prettier-plugin-svelte": "^4.1.1",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
         "rollup": "^4.0.0",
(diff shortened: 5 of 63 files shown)


lukas-reining:
Hey @pharv, thank you for this huge contribution!
Please excuse the response time for this one. I will try to reach out to the other maintainers the next days. 
For this one, before accepting we need to be sure that we have someone from the maintainers to take responsibility for maintaining the SDK. Maybe I can do it. Also it would be great to know if you are willing to continue maintaining it.

pharv:
> Hey @pharv, thank you for this huge contribution! Please excuse the response time for this one. I will try to reach out to the other maintainers the next days. For this one, before accepting we need to be sure that we have someone from the maintainers to take responsibility for maintaining the SDK. Maybe I can do it. Also it would be great to know if you are willing to continue maintaining it.

Hey Lukas, I appreciate it.  I would gladly help maintain it.  We plan on using OpenFeature extensively for the foreseeable future, so would love to help contribute to this project in any way I can.

